PR #263, opened 2015-02-26
fixes audio page phone margin

upsworld:
so text wont disappear behind additional infobox


tf:
Maybe we can also clean up [these mentions](https://github.com/codevise/pageflow/blob/master/app/assets/stylesheets/pageflow/page.css.scss#L180) of `.videoPage` and `.audioPage` in `page.css.scss` while we are at it?

tf:
I guess no answer means no.

PR #278, opened 2015-04-02
Don't provide alternative images for open graph if share image is picked

schoetty:
When the user picks a share image, that very image should be used to share the pageflow. facebook does not offer to choose between multiple images when sharing, so we shouldn't provide alternative images.

### DIFF
--- a/app/helpers/pageflow/social_share_helper.rb
+++ b/app/helpers/pageflow/social_share_helper.rb
@@ -32,14 +32,16 @@ module Pageflow
     def social_share_entry_image_tags(entry)
       image_urls = []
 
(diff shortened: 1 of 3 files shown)


app/helpers/pageflow/social_share_helper.rb
```diff
@@ -32,14 +32,16 @@ def social_share_page_description(page)
...
+      else
+        entry.pages.each do |page|
+          if image_urls.size >= 4
+            break
+          else
+            image_urls << page.thumbnail_url(:medium)
+            image_urls.uniq!
+          end
         end
       end
```
tf: Can we add a test here? I'm surprised we didn't so far, given the complexity of the logic.

schoetty:
Merged in #280.

PR #279, opened 2015-04-02
Add fallbacks for page and entry share description texts

schoetty:


app/helpers/pageflow/social_share_helper.rb
```diff
@@ -50,5 +59,9 @@ def social_share_normalize_protocol(url)
       url.gsub(/^(\/\/|https:\/\/)/, 'http://')
     end
 
+    private
+    def strip(text)
+      strip_tags(text.gsub(/<br ?\/?>/, ' ').squish)
+    end
```
tf: Private helper methods with short names like this are prone to namespace collisions. If another helper also defines a `strip` method, only one will win and the other will be overwritten.

app/helpers/pageflow/social_share_helper.rb
```diff
@@ -24,8 +24,17 @@ def social_share_page_title(page)
...
+      social_share_entry_description(page.chapter.revision)
+    end
+
+    def social_share_entry_description(entry)
+      return strip(entry.summary) if entry.summary.present?
+
+      entry.pages.each do |page|
+        return strip(page.configuration['text']) if page.configuration['text'].present?
+      end
       ''
```
tf: Given the increasingly complex logic of these methods, I think, we should add some tests.

app/helpers/pageflow/social_share_helper.rb
```diff
@@ -24,8 +24,17 @@ def social_share_page_title(page)
     end
 
     def social_share_page_description(page)
-      return page.configuration['text'] if page.configuration['text'].present?
-      return page.configuration['description'] if page.configuration['description'].present?
+      return strip(page.configuration['text']) if page.configuration['text'].present?
+      return strip(page.configuration['description']) if page.configuration['description'].present?
+      social_share_entry_description(page.chapter.revision)
```
tf: I'd prefer adding an `entry` parameter to the method. Looking up the revision like this might trigger separate database accesses. Moreover, this only works since the passed revision has the same duck-type as the `entry` expected by `social_share_entry_description` method. Still `Revision#summary` does not need to have the name behavior as `PublishedEntry#summary`.

schoetty:
Merged in #280.

PR #280, opened 2015-04-08
Improve sharing features

schoetty:
* Add fallbacks for page and entry share description texts 
* Don't provide alternative images for open graph if share image is picked


tf:
see #288.

PR #282, opened 2015-04-22
Fix installation procedure

tilsammans:
Following the steps with a blank Rails app didn't actually work.

Fixed the steps so that the procedure runs without errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,19 @@ Run bundler to install dependencies:
 
     $ bundle install
 
+### Create the database
+
+As the generator needs the presence of a database, you must create it:
+
+    $ rake db:create
+
+### Generate a User model
+
+Pageflow assumes the presence of a `User` model and corresponding `users`
+table in the database. If your application doesn't have that, generate it:
+
+    $ rails generate resource user name email encrypted_password
+
 ### Running the Generator
 
 Now you can run the generator to setup Pageflow and its dependencies:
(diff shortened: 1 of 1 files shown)


tf:
Hi,

thanks very much for the PR. Looking at the [rails template](https://github.com/codevise/pageflow/blob/master/spec/support/pageflow/dummy/rails_template.rb) that is used to generate the dummy pageflow application in CI, the database indeed has to be created upfront. I don't think creating the user manually is necessary, though. Anyway, you are right that the installation procedure has to be fixed. I'll also try to run through it so we can merge our changes.

Cheers
Tim

tilsammans:
Hi @tf 

If you run through it yourself you will see that the `User` model is indeed a requirement.

Cheers.

tf:
I've tried the install procedure and creating the user was not necessary. Devise generated an extensionless migration file though. This has been mentioned in the [trouble shooting wiki page](https://github.com/codevise/pageflow/wiki/Troubleshooting#devise-migration-without-file-extension), but that might not be prominent enough for such an inconvenient bug.

If made some changes to the README in #285 following your suggestions and also filling some other gaps I encountered. Would you mind taking a look if I missed anything important?

tilsammans:
hi @tf looks good. :+1: 

I'll close this pull request.

tf:
Ok, thanks. I've merged the other PR.

PR #283, opened 2015-04-22
Cannot use ruby 2.2.x

tilsammans:
Because of ActiveAdmin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Pageflow assumes the following choice of libraries:
 
 Pageflow runs in environments with:
 
-* Ruby 1.9.3 or higher
+* Ruby >= 1.9.3 and < 2.2.0 because of [ActiveAdmin](https://github.com/activeadmin/activeadmin/issues/3715)
 * Rails 4.0 or 4.1
 * Redis server (for Resque)
 * A database server supported by Active Record (tested with MySQL)


tf:
Hi,

thank for the PR. The linked issue seems to refer to Rails 3.2 compatibility with Ruby 2.2. Pageflow uses Rails 4 though. Is ActiveAdmin known to be incompatible with Ruby 2.2?

tilsammans:
Yes. I had Ruby 2.2.2 and got that exact same error.

After switching to 2.1.6 the error in ActiveAdmin went away.

tf:
Ah ok. Just out of curiosity: What was the error message?

g-w:
Maybe we should also change the lower bound to 2.0.0? This reflects the end of life of ruby 1.9.3. See https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

tilsammans:
The error message is:

```
undefined method `reorder' for nil:NilClass
```

I would recommend against changing the lower bounds to Ruby 2.0.0. I know 1.9.3 is EOL but my patch is concerned about Pageflow just not working at all, and not about policies regarding version numbers.

tf:
I agree. Long as we are running Pageflow instances with Ruby 1.9.3 ourselves, we can keep it in the support matrix.

PR #284, opened 2015-04-24
use ENV for secret keys

tilsammans:
Hard-coding secret keys in configuration files isn't good practice.

This pull request assumes the secrets are present in ENV instead. People who are not using ENV can still hard-code them in the file if they want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ For a high level introduction and example Pageflow stories see
 [pageflow.io](http://pageflow.io).
 
 In addition to this README, there is also a [Getting Started Wiki page](https://github.com/codevise/pageflow/wiki/Getting-Started)
-to guide you through the steps of setting up a Rails application with Pageflow 
+to guide you through the steps of setting up a Rails application with Pageflow
 on your development machine.
 
 ## Updating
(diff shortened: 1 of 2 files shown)


tf:
I'd be open to taking this route. Originally, the template used hard coded strings to leave the dotenv/secrets.yml decision to the user. But probably using env is a more sane default.

* There one line using `env` instead of `ENV` causing the build to fail.
* I wonder if we should `ENV.fetch` in order to fail loudly if something is missing.
* The comment stating that hard coding is bad practice should be rephrased.
* Could you remove the whitespace fixes from the commit? No idea, why whitespace is so messed up in the README, but fixing it should not be part of this change i guess.

tilsammans:
No problem. I will build on this PR and we can continue to discuss! Thanks for your consideration.

tf:
I'll close this for now due to inactivity. Feel free to re-open any time. I still think this is a good idea.

tilsammans:
I opened up #306 since commits don't show up for a closed PR.

I agree with using fetch, this makes the requirement very clear. I am not sure what you'd like to see differently in the README, it's clear to me, but feel free to comment on it and I will use your copy.

PR #306, opened 2015-06-19
Env for api keys

tilsammans:
Continuation of #284

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,8 @@ Pageflow stores files in S3 buckets also in development
 mode. Otherwise there's no way to have Zencoder encode them. See the
 wiki page [Setting up external services](https://github.com/codevise/pageflow/wiki/Setting-up-External-Services).
 
+The host application can utilize environment variables to configure the API keys for S3 and Zencoder. The variables can be found in the generated Pageflow initializer.
+
 For available configuration options and examples see the inline docs
 in `config/initializers/pageflow.rb` in your generated rails app.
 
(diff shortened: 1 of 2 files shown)


tilsammans:
@tf the ENV keys are not found in testing. Where do you prefer I put those keys? We have options ... use one of the gems for it, or directly in spec_helper, or even travis.yml. What do you think?

tf:
Hmm. We have the same problem when running specs locally. So `.travis.yml` is not an option. Also the app needs to be able to initialize on its own to run things like `rails g` inside the gem repository. So only setting environment variables inside `spec_helper.rb` won't do either.

It would probably be most expressive to have the `pageflow:install` generator create a `.env.example` file and use that inside the dummy app. But that would mean adding a dependency to `dotenv` - a choice that we leave to the host application at the moment.

On the other hand, with the changes of this PR so far, we loose some example data (i.e. "what should `s3_host_name` look like?"). That's also something a `.env.example` file would give us.

Maybe there should be a `--dotenv` flag for the `pageflow:install` generator? Not sure really.

tilsammans:
If I'm honest I don't like any of those options. :neutral_face: 

A dotenv runtime dependency is a bad choice. We must not care what the host app does to deal with env vars. A 12factor app will have them set via the command line anyway.

Dotenv as a development dependency for the gem might be acceptable to you. I'll leave that decision to you, I feel slightly bad about it, but not terrible.

The most simple option might just be to not use `ENV.fetch` but use a fallback value for all ENV vars. Something like `ENV['S3_HOST_NAME'] || 's3_host_name'`. It's maybe aesthetically the least pleasing but does what we need.

tf:
Your are right. Adding defaults is the pragmatic thing to do. Then we also have example config values again. `fetch` takes a second parameter which is used as default value if the passed key is not present.

tilsammans:
Okay, I will push that up.

tf:
Great. Could you squash to a single commit, so we can skip the intermediate steps in the git history.

tilsammans:
@tf done and the build is green.

tf:
Thanks :+1:

PR #382, opened 2015-08-14
Fix a bug in collection_view.js

romanvolkov-cs:
togglePlaceHolder function didn't work properly such that no loading spinner view could be displayed.


romanvolkov-cs:
togglePlaceHolder function didn't work properly such that no loading spinner view could be displayed.

PR #385, opened 2015-08-18
fix theme name replacement in generator

### LINKED ISSUE #384: Theme generation doesn't apply $theme-name
I've just generated a new theme but the `$theme-name` variable is still listed as `default`.

Isn't this supposed to become the theme name?

The contents of my `app/assets/stylesheets/pageflow/themes/scrollytelling/variables.css.scss`:

```
$theme-name: "default";

$main-color: #e84d2c;
$main-color-dark: #b14a08;

$focus-color: #c1c1c1;

$anchor-color: #e84d2c;

$basic-background-color: #353535;
$basic-background-color-transparent: rgba(53, 53, 53, 0.9);
$header-background-color: gray;

$addInfobox-text-color: #fff; /* for the description text inside the info box */
$addInfobox-controls-text-color: #ccc; /* for control text like durations */
$addInfobox-background-color: #191919;
$addInfobox-border-color: #505050;

$standard-font: SourceSansPro;
$standard-font-size: 16px;
```

tilsammans:
fixes #384 

In commit a0ad811 the generator was refactored, the contents of the main
stylesheet were moved into variables.css.scss. This change fixes the
generator so that the generated theme name is replaced again.

### DIFF
--- a/lib/generators/pageflow/theme/theme_generator.rb
+++ b/lib/generators/pageflow/theme/theme_generator.rb
@@ -13,9 +13,10 @@ module Pageflow
         directory('images/pageflow/themes/default', themes_path('images', name))
 
         template('stylesheets/pageflow/themes/default.css.scss', themes_path('stylesheets', "#{name}.css.scss")) do |content|
-          content.gsub!('$theme-name: "default";', %Q'$theme-name: "#{name}";')
           content.gsub!('@import "./default/', %Q'@import "./#{name}/')
         end
+
+        gsub_file themes_path('stylesheets', "#{name}/variables.css.scss"), '$theme-name: "default";', %Q'$theme-name: "#{name}";'
       end
 
       private


lib/generators/pageflow/theme/theme_generator.rb
```diff
@@ -13,9 +13,10 @@ def copy_template
         directory('images/pageflow/themes/default', themes_path('images', name))
 
         template('stylesheets/pageflow/themes/default.css.scss', themes_path('stylesheets', "#{name}.css.scss")) do |content|
-          content.gsub!('$theme-name: "default";', %Q'$theme-name: "#{name}";')
           content.gsub!('@import "./default/', %Q'@import "./#{name}/')
         end
+
+        gsub_file themes_path('stylesheets', "#{name}/variables.css.scss"), /default/, name
```
tf: Replacing all occurrences of the string `default` might be dangerous if there ever is a variable like `pageflow-default-color`, right? Maybe we should stick with matching `$theme-name: ...`?

tilsammans: I had a feeling you would say that. :grin: 

Agreed. It makes it more brittle in case anyone changes the double quotes to singles. But it becomes less dangerous too. I'll make the change.

tf: You are right about the quotes, though. Can we use a regex like `/$theme-name: ['"]default["']/`?

tf:
Would you mind squashing the commits?

PR #386, opened 2015-08-18
replaced sample video with smaller file

### LINKED ISSUE #374: Use Smaller Fixture Video
#369 includes the `spec/fixtures` directory in the gem since those files are [used by the FactoryGirl factories](https://github.com/codevise/pageflow/blob/master/spec/factories/video_files.rb#L7) which in turn are included in the gem to be available to host applications.

Especially the [fixture video file](https://github.com/codevise/pageflow/blob/b77c926bb315c62095bb87a440fad706f35ab46f/spec/fixtures/video.mp4) is quite large and bloats the gem size. It should be replaced with a much smaller fixture file.

tilsammans:
Video was downloaded from https://support.apple.com/en-us/HT201549
fixes #374


tf:
Thanks! :+1:

PR #396, opened 2015-08-25
Replace Audio Loop Page with Atmo

tf:
An audio loop page has almost the same functionality as a background image page with an atmo audio file.

* Remove atmo feature flag
* Remove audio loop page type
* Add migration to convert all audio loop pages to background image pages with an atmo audio file


tf:
We need a better way to do that migration. Using AR models does not work if there are other pending migrations.

tf:
We need a more gentle transition process. Simply removing the page type might leave users not finding the new way to achieve background audio.

PR #401, opened 2015-09-04
Implement options for disabling rich text and links in TextAreaInputView

romanvolkov-cs:
TextAreaInputView introduces two new options, disableRichtext & disableLinks. Both are false by default.

### DIFF
--- a/app/assets/javascripts/pageflow/ui/views/inputs/text_area_input_view.js
+++ b/app/assets/javascripts/pageflow/ui/views/inputs/text_area_input_view.js
@@ -22,11 +22,11 @@ pageflow.TextAreaInputView = Backbone.Marionette.ItemView.extend({
       toolbar: this.ui.toolbar[0],
       parserRules: {
         tags: {
-          u: {},
-          b: {},
-          i: {},
+          u: this.options.disableRichtext ? 0 : {},
+          b: this.options.disableRichtext ? 0 : {},
+          i: this.options.disableRichtext ? 0 : {},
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/pageflow/ui/views/inputs/text_area_input_view.js
```diff
@@ -39,6 +39,15 @@ pageflow.TextAreaInputView = Backbone.Marionette.ItemView.extend({
       }
     });
 
+    if (this.options.disableRichtext) {
+      this.ui.toolbar.find('a[data-wysihtml5-command="bold"]').remove();
```
tf: Let's toggle a css class on the outer element and hide controls with `display: none` instead of removing them from the DOM.

app/assets/javascripts/pageflow/ui/views/inputs/text_area_input_view.js
```diff
@@ -22,11 +22,11 @@ pageflow.TextAreaInputView = Backbone.Marionette.ItemView.extend({
       toolbar: this.ui.toolbar[0],
       parserRules: {
         tags: {
-          u: {},
-          b: {},
-          i: {},
+          u: this.options.disableRichtext ? 0 : {},
```
tf: Does using `0` here do something special or could it also be `false`? How does the text field behave when inserting text containing formatted text? Is the format removed correctly if `disableRichtext` is `true`?

Issue #388, opened 2015-08-19
Allow SSL for published entries


tilsammans:
A private controller method named `prevent_ssl` is used in two places. I was wondering why? We have our entire site behind SSL and this is causing problems.

https://github.com/codevise/pageflow/blob/1babee2a31d78d2ce509af849733a5a104484093/app/controllers/pageflow/application_controller.rb#L57

tf:
We have always been serving assets over CDN endpoints which do not support HTTPS. To prevent mixed content warnings we chose to redirect back to HTTP for published pageflows. I agree, though, that this should be configurable via some kind of `Pageflow.config.prevent_ssl_for_public_sites` setting, which should default to `true` for backward compatibility. If you are interested in looking into that, I could give you a few more ideas how I'd implement and test this feature.

tilsammans:
We went the other way, force https for everyone, and use the secure endpoints on CloudFront.

Sure, I can work on that feature. It won't be today or tomorrow (and I monkey-patched the gem, don't tell anyone) but I'd like to solve it cleanly. Any pointers are welcome.


Op 20 aug. 2015, om 10:45 heeft Tim Fischbach <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


We have always been serving assets over CDN endpoints which do not support HTTPS. To prevent mixed content warnings we chose to redirect back to HTTP for published pageflows. I agree, though, that this should be configurable via some kind of Pageflow.config.prevent_ssl_for_public_sites setting, which should default to true for backward compatibility. If you are interested in looking into that, I could give you a few more ideas how I'd implement and test this feature.

tf:
Great. And it shouldn't be too complicated either. Maybe there should then even be three possible values for some `public_entry_ssl_mode` like `:allow`, `:prevent` and `:enforce`?

Anyway, the additional config option would go into [`Pageflow::Configuration`](https://github.com/codevise/pageflow/blob/master/lib/pageflow/configuration.rb). The `preven_ssl` could be renamed and moved from the application controller junk drawer to some controller concern module which could be mixed into `Pageflow::EntriesController` and `Pageflow::FilesController`. 

We should test the different redirects at controller level, possibly with a shared example group. `Pageflow.config` can simply be manipulated inside the test setup since it is reset before ever example run.

PR #403, opened 2015-09-08
make https mode configurable

tilsammans:
Pageflow has always been serving assets over a CDN without https
endpoint. The prevent_ssl filter made sure to redirect to a non-https
method when assets are involved. This change makes that setting
configurable. You can now ignore, enforce or prevent https.

The default is to prevent https, like how Pageflow < 1.0 behaved.
However in new initializers the generated value is :ignore. This should
make the setting compatible with most applications out of the box.

fixes codevise/pageflow#388

### DIFF
--- a/app/controllers/concerns/pageflow/public_https_mode.rb
+++ b/app/controllers/concerns/pageflow/public_https_mode.rb
@@ -1,15 +1,13 @@
 module Pageflow
   module PublicHttpsMode
-
     protected
 
     def check_public_https_mode
       if request.ssl? && Pageflow.config.public_https_mode == :prevent
-        redirect_to("http://#{request.host}#{request.fullpath}", :status => :moved_permanently)
+        redirect_to("http://#{request.host}#{request.fullpath}", status: :moved_permanently)
       elsif !request.ssl? && Pageflow.config.public_https_mode == :enforce
-        redirect_to("https://#{request.host}#{request.fullpath}", :status => :moved_permanently)
+        redirect_to("https://#{request.host}#{request.fullpath}", status: :moved_permanently)
       end
     end
-
   end
 end
--- a/app/controllers/concerns/pageflow/public_https_mode.rb
+++ b/app/controllers/concerns/pageflow/public_https_mode.rb
@@ -1,0 +1,15 @@
+module Pageflow
+  module PublicHttpsMode
+
+    protected
+
+    def check_public_https_mode
+      if request.ssl? && Pageflow.config.public_https_mode == :prevent
+        redirect_to("http://#{request.host}#{request.fullpath}", :status => :moved_permanently)
+      elsif !request.ssl? && Pageflow.config.public_https_mode == :enforce
+        redirect_to("https://#{request.host}#{request.fullpath}", :status => :moved_permanently)
+      end
+    end
+
+  end
+end
--- a/app/controllers/pageflow/application_controller.rb
+++ b/app/controllers/pageflow/application_controller.rb
@@ -54,12 +54,6 @@ module Pageflow
       root_url(:protocol => 'http')
     end
 
-    def prevent_ssl
-      if request.ssl?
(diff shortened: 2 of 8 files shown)


tilsammans:
wow, hound is strict! :fearful: let me know if you want me to resolve these issues in this PR.

tf:
First of all thanks for the well rounded PR!

Yeah, the Hound config is still a bit rough. And I'm also aware the code base as it stands absolutely does not follow all conventions (especially Ruby 1.9 hash syntax). But I'd really like it to converge in that direction.

We should probably change the Hound config to allow long lines. Especially in tests I'm often not sure readability is improved by spanning multiple lines. But for the rest, it would be great if you could fix those.

Two more small nitpicks: 

* I tend to prefer verbs for filter names. So I think something like `check_public_https_mode` or `ensure_public_https_mode` would be a bit clearer. What do you think?
* Most part of the test suite follows the convention of grouping the lines of specs according to the three A's (Arrange Act Assert). So could you insert two blank lines in each of your specs to separate test setup, requests triggering and expectations? 

Thanks again
Tim

tilsammans:
No problem. I was not aware of the three As. I will change the spec in that way. Totally agree on long lines too. 100 chars even isn't that long.

I can submit a revised PR tonight at the earliest.



Op 8 sep. 2015 om 15:23 heeft Tim Fischbach <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


First of all thanks for the well rounded PR!

Yeah, the Hound config is still a bit rough. And I'm also aware the code base as it stands absolutely does not follow all conventions (especially Ruby 1.9 hash syntax). But I'd really like it to converge in that direction.

We should probably change the Hound config to allow long lines. Especially in tests I'm often not sure readability is improved by spanning multiple lines. But for the rest, it would be great if you could fix those.

Two more small nitpicks:

  *   I tend to prefer verbs for filter names. So I think something like check_public_https_mode or ensure_public_https_mode would be a bit clearer. What do you think?
  *   Most part of the test suite follows the convention of grouping the lines of specs according to the three A's (Arrange Act Assert). So could you insert two blank lines in each of your specs to separate test setup, requests triggering and expectations?

Thanks again
Tim

-

tf:
Great. You can simply rebase your commit and force push to the branch of this PR.

tilsammans:
Revised commit pushed up.

If you don't mind I would like to keep the 1.9 hash syntax out of this and submit a separate pull request to change all hashes in the entire project to the new syntax. Using `perl -pi -e 's/:([\w\d_]+)(\s*)=>/\1:/g' **/*.rb` on the entire project. It seems cleaner to me that way.

tf:
That will cause too many merge conflicts for other pending PRs and some branches I am working on waiting to become PRs. I think "Don't rewrite existing code to follow the style guide" is a good rule (see also the [last item in the Hound FAQ](https://houndci.com/faq) which makes this point quite well).

tilsammans:
Okay, fair enough. My OCD prevents me from having two different hash notations in one file though. ;) What do you think? You don't mind having two hash styles in one file?

tf:
I totally see your point. But I'm also afraid of messing things up once I have to rebase a PR and resolve merge conflicts with style guide fixes. I guess my my rule of thumb would be: If I am already rewriting a good part of a file, I might as well fix the remaining violations as long as I do it in a separate commit so "essential" merge conflicts stand out from "non-essential" one later on. Makes sense? 

Otherwise I would probably tell myself that `=>` hash notation is still valid Ruby and there are even still situations where you have to use it anyway (i.e. for non-symbol keys).

tilsammans:
It's an ugly PR page but a clean branch; with the Hound improvements in a second commit. @tf ready for final review :)

tf:
Yeah, I have to say, Hound is quite the noisy fellow. If you only fixed the violations in your own changes you could have squashed the commits. But it's fine. I'll merge now. I also raised the max line length to 100 chars for the future.

PR #404, opened 2015-09-09
Implement options for disabling rich text and links in TextAreaInputView

romanvolkov2:
TextAreaInputView introduces two new options, disableRichtext & disableLinks. Both are false by default.


tf:
When pasting text that contains formatted words, bold parts etc are stripped out completely. While it's right to remove the formatting, the contents itself should remain intact.

tf:
when pasting text with `strong` or `em` tags, they are turned into `b` and `i` tags even if richtext is disabled. Formatting is not actually saved, but it would be good to prevent the initial display. I guess, we also need `unwrap` instructions for those tags instead of the `rename` used now if richtext is disabled.

Issue #406, opened 2015-09-10
Extract translations of public sites


tf:
Pageflow's translations are managed via [Localeapp](https://www.localeapp.com/) projects. There's a project [for each pageflow gem](https://github.com/codevise/pageflow/wiki/Editing-Translations). Still translations for editor elements are stored side by side with translations used in the UI of published entries.

This makes adding a new language quite hard, even if it is only supposed to be used in published entries:

* It is not clear which translations are relevant for public pages.
* Each Pageflow gem has to be edited separately.

To remedy this situation, we'd like to create a new `pageflow-public-i18n` gem which contains only those translations used in published entries.

Normally translation keys are closely coupled to the containing gem and the place of usage:

    pageflow.progress_navigation_bar.widget.home_button.home_title: "Go to homepage"

To reduce coupling, the new gem should provide a separate `pageflow.public` namespace in which key names are chosen according to the meaning of the translated text instead of it's position of usage. So the above example should be:

    pageflow.public.go_to_homepage: "Go to homepage"

In many cases the existing english translation might be a good starting point when choosing the key. If necessary, the key can contain additional context. It is important that no key is used in two different places where it's meaning differs, only because the existing translations use the same word.

That way new Pageflow plugins requiring only, for example, the caption for a close button can simply use an existing key. If they need new translations, they need to make a PR to the `pageflow-public-i18n` gem and depend on its next minor version.

Somebody willing to translate Pageflow to a new language then only needs to come up with translations for the existing keys.

---

The first step now is to go through all of the existing Pageflow repositories, find translations visible in published entries and replace the keys with `pageflow.public...` keys. These changes can be submitted as a PR together with locale files of the following form:

    # config/locales/new/public.de.yml
    de:
      pageflow:
        public:
          go_to_homepage: "Zur Startseite"
      
    # config/locales/new/public.de.yml
    en:
      pageflow:
        public:
          go_to_homepage: "Go to homepage"

    # config/locales/new/public.removed.yml
    de:
        progress_navigation_bar:
          widget:
            home_button:
              home_title: "NOT_USED"

That way, the first two files can be imported into a new Localeapp project for `pageflow-public` and the last one can be imported into the existing Localeapp project of the corresponding gem. Then we can search for translations with "NOT_USED" and delete those.

---

The following repositories need PRs:

- [x] [codevise/pageflow](https://github.com/codevise/pageflow) (#410)
[...]

PR #410, opened 2015-10-09
Extract publicly visible i18n strings from locale

romanvolkov2:
Solves issue #406 “Extract translations of public sites”


tf:
Thanks! I'll merge after I have tested it locally. I've updated #406 with a list of repositories that need to be looked at next.

PR #412, opened 2015-10-15
use the current Locale

tilsammans:
and not the hard-coded "de"


tf:
That makes sense. Can you rebase onto master? I've enabled the GitHub feature "protected branch" for master, which prevents non fast forward merges, I think.

tilsammans:
@tf done

PR #453, opened 2015-12-09
Add commented editor_route_constraint option

tilsammans:
This is useful to have generated in the initialiser. When people need the default option presented in the Wiki, they can just uncomment this line.


tf:
I can see that. Not sure though if it might be confusing that one has to define `HostConstraint` to make it work.

tilsammans:
@tf that's true. :smile: 

Your call, I don't mind if you close it.

tf:
Well, let's go for an improved wiki regarding CNAME and redirects instead.

PR #473, opened 2016-01-16
features must be registered with a String

tilsammans:
This is explicitly checked:
https://github.com/codevise/pageflow/blob/127e61bb2b371ed0c799bd3abb26de77602a31b5/lib/pageflow/features.rb#L30

passing a symbol here will not work.


tf:
True. The `register` method might be a little bit to picky, though. I guess one could check for symbols as well and call `to_s` somewhere to ensure comparison with the enabled feature names coming from the db works out alright later.

PR #481, opened 2016-01-29
Fix wrong text alignment in right position mode

romanvolkov2:
If text position is set to right and the user's screen is too small, subheadings align incorrectly.

bildschirmfoto 2016-01-25 um 14 51 01


tf:
This actually changes the line-width. The 60% was chosen deliberately. This should either only be changed for "text position right" case or the width of the title should also be set to 60%. To make this part of the CSS more clear we should merge [text_variants.css.scss](https://github.com/codevise/pageflow/blob/master/app/assets/stylesheets/pageflow/text_variants.css.scss) with page.css.scss.

PR #491, opened 2016-02-05
Add columns and name filter to admin accounts panel

romanvolkov-io:
Counter caches are being used for users and entries which belong to an account. Translations must be added manually.

### DIFF
--- a/admins/pageflow/accounts.rb
+++ b/admins/pageflow/accounts.rb
@@ -3,13 +3,23 @@ module Pageflow
     menu :priority => 3
 
     config.batch_actions = false
-    config.clear_sidebar_sections!
 
     index do
       column :name do |account|
         link_to account.name, admin_account_path(account)
       end
+      column :entries_count do |account|
+        account.entries_count
+      end
+      column :users_count do |account|
+        account.users_count
+      end
+      column :default_theming do |account|
+        account.default_theming.theme_name
+      end
     end
+
+    filter :name
 
     form :partial => 'form'
 
@@ -65,6 +75,10 @@ module Pageflow
         end
 
         result
+      end
+
+      def scoped_collection
+        super.includes(:users, :entries, :themings)
       end
 
       private
--- a/lib/pageflow/user_mixin.rb
+++ b/lib/pageflow/user_mixin.rb
@@ -9,20 +9,20 @@ module Pageflow
     include Suspendable
 
     included do
-      belongs_to :account, :class_name => 'Pageflow::Account'
+      belongs_to :account, counter_cache: true, class_name: 'Pageflow::Account'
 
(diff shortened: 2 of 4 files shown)


lib/pageflow/user_mixin.rb
```diff
@@ -9,7 +9,7 @@ module UserMixin
     include Suspendable
 
     included do
-      belongs_to :account, :class_name => 'Pageflow::Account'
+      belongs_to :account, :counter_cache => true, :class_name => 'Pageflow::Account'
```
tf: Yes. New or changed code should use new hash syntax. That way we can migrate the code base step by step.

admins/pageflow/accounts.rb
```diff
@@ -3,14 +3,24 @@ module Pageflow
     menu :priority => 3
 
     config.batch_actions = false
-    config.clear_sidebar_sections!
 
     index do
       column :name do |account|
         link_to account.name, admin_account_path(account)
       end
+      column I18n.t('pageflow.admin.accounts.entries_count') do |account|
```
tf: Add translation key `activerecord.attributes."pageflow/account".entries_count` instead. That way it is picked up by ActiveAdmin automatically if you use `column :entries_count`.

admins/pageflow/accounts.rb
```diff
@@ -3,14 +3,24 @@ module Pageflow
...
+        account.entries_count
+      end
+      column I18n.t('pageflow.admin.accounts.users_count') do |account|
```
tf: see above

admins/pageflow/accounts.rb
```diff
@@ -3,14 +3,24 @@ module Pageflow
...
+        account.users_count
+      end
+      column I18n.t('pageflow.admin.accounts.default_theming') do |account|
```
tf: see above

app/jobs/pageflow/reset_counter_caches.rb
```diff
@@ -0,0 +1,13 @@
+module Pageflow
+  class ResetCounterCachesJob
```
tf: This should be a migration instead. Using ActiveRecord models from migrations is not allowed though (the (future) schema assumed by the models might not match the schema at that point during the migration process), so we'll have to go with raw SQL queries here ([see example](https://github.com/codevise/pageflow/blob/master/db/migrate/20151022080518_add_first_published_at_to_entries.rb#L5))

admins/pageflow/accounts.rb
```diff
@@ -3,14 +3,24 @@ module Pageflow
...
+        account.default_theming.theme_name
```
tf: Please override `scoped_scollection` in account admin to prevent N+1 query. See http://activeadmin.info/docs/2-resource-customization.html section "Customizing resource retrieval".

tf:
see #568

PR #492, opened 2016-02-05
Use embedded index table for entries, members and users

romanvolkov-io:
Implement column sorting as well.

### DIFF
--- a/app/assets/stylesheets/pageflow/admin.css.scss
+++ b/app/assets/stylesheets/pageflow/admin.css.scss
@@ -1,5 +1,6 @@
 @import "pageflow/mixins";
 
+@import "pageflow/admin/accounts";
 @import "pageflow/admin/columns";
 @import "pageflow/admin/entries";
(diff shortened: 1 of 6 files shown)


tf:
Please clean up the Hound warnings first, it'll be easier to discuss the code then.

The sorting/order logic should be moved inside the `embedded_index_table`. I suggest using ActiveAdmin's [`OrderClause`](https://github.com/activeadmin/activeadmin/blob/b44b4f4896ef78b50c35c82a343356d68b937cde/lib/active_admin/order_clause.rb) instead of custom string manipulations.

We should also start adding tests for this component. I've began with the first two [here](https://github.com/codevise/pageflow/compare/master...tf:embedded-index-table-specs). Please add tests for the blank slate case and the new sorting functionality.

Regarding sorting memberships: If you use `entry.memberships.includes(:user)` then you should be able to do use `pageflow_users.last_name ASC` to order the table.

tf:
see #563

PR #494, opened 2016-02-10
Add columns and filters to admin entries panel

romanvolkov-io:
Optimizations for the admin entries panel such as new filters and index columns.


tf:
see #567

PR #499, opened 2016-02-16
Make author and publisher fields dynamic

tilsammans:
Set the :author and :publisher values in entry show, which allows them
to be picked up in the page <head>. Fallback to static value 'Pageflow'
for now.

### DIFF
--- a/app/helpers/pageflow/entries_helper.rb
+++ b/app/helpers/pageflow/entries_helper.rb
@@ -1,5 +1,12 @@
 module Pageflow
   module EntriesHelper
+    def pretty_entry_title(entry)
+      [].tap do |title|
+        title << entry.title
+        title << entry.theming.cname_domain if entry.theming.cname_domain.present?
+      end.join(' - ')
+    end
+
     def pretty_entry_url(entry)
       pageflow.short_entry_url(entry.to_model, Pageflow.config.theming_url_options(entry.theming))
     end
--- a/app/helpers/pageflow/entries_helper.rb
+++ b/app/helpers/pageflow/entries_helper.rb
@@ -1,10 +1,7 @@
 module Pageflow
   module EntriesHelper
     def pretty_entry_title(entry)
-      [].tap do |title|
-        title << entry.title
-        title << entry.theming.cname_domain if entry.theming.cname_domain.present?
-      end.join(' - ')
+      [entry.title, entry.theming.cname_domain.presence].compact.join(' - ')
     end
 
     def pretty_entry_url(entry)
--- a/app/models/pageflow/draft_entry.rb
+++ b/app/models/pageflow/draft_entry.rb
@@ -21,6 +21,7 @@ module Pageflow
              :files,
              :image_files, :video_files, :audio_files,
              :locale,
+             :author, :publisher, :keywords,
              :to => :draft)
 
     def initialize(entry, draft = nil)
--- a/app/models/pageflow/draft_entry.rb
+++ b/app/models/pageflow/draft_entry.rb
@@ -29,6 +29,11 @@ module Pageflow
       @draft = draft || entry.draft
     end
 
+    # So we can always get to the original Entry title.
+    def entry_title
+      entry.title
+    end
+
     def create_file(model, attributes)
       file = model.create(attributes) do |f|
         f.entry = entry
--- a/app/models/pageflow/published_entry.rb
+++ b/app/models/pageflow/published_entry.rb
@@ -11,6 +11,7 @@ module Pageflow
              :enabled_feature_names,
              :to_model, :to_key, :persisted?,
              :authenticate,
+             :users,
              :to => :entry)
 
     delegate(:widgets,
--- a/app/models/pageflow/published_entry.rb
+++ b/app/models/pageflow/published_entry.rb
@@ -23,6 +23,7 @@ module Pageflow
              :emphasize_new_pages,
              :share_image_id, :share_image_x, :share_image_y,
              :locale,
+             :author, :publisher, :keywords,
              :password_protected?,
              :to => :revision)
 
--- a/app/models/pageflow/published_entry.rb
+++ b/app/models/pageflow/published_entry.rb
@@ -33,11 +33,7 @@ module Pageflow
     end
 
     def title
-      if revision
-        revision.title.presence || entry.title
-      else
-        entry.title
-      end
+      revision.title.presence || entry.title
     end
 
     def stylesheet_model
--- a/app/models/pageflow/published_entry.rb
+++ b/app/models/pageflow/published_entry.rb
@@ -11,7 +11,6 @@ module Pageflow
              :enabled_feature_names,
              :to_model, :to_key, :persisted?,
              :authenticate,
-             :users,
              :to => :entry)
 
     delegate(:widgets,
--- a/app/models/pageflow/published_entry.rb
+++ b/app/models/pageflow/published_entry.rb
@@ -18,7 +18,7 @@ module Pageflow
              :storylines, :main_storyline_chapters, :chapters, :pages,
              :files,
              :image_files, :video_files, :audio_files,
-             :title, :summary, :credits, :manual_start,
+             :summary, :credits, :manual_start,
              :emphasize_chapter_beginning,
              :emphasize_new_pages,
              :share_image_id, :share_image_x, :share_image_y,
@@ -30,6 +30,14 @@ module Pageflow
       @entry = entry
       @revision = revision || entry.published_revision
       @custom_revision = !!revision
+    end
+
+    def title
+      if revision
+        revision.title.presence || entry.title
+      else
+        entry.title
+      end
     end
 
     def stylesheet_model
--- a/app/views/pageflow/entries/_meta_tags.html.erb
+++ b/app/views/pageflow/entries/_meta_tags.html.erb
@@ -1,4 +1,0 @@
-  <meta name="dc.language" content="<%= I18n.locale %>" />
-  <% if entry.keywords.present? %><meta name="keywords" content="<%= entry.keywords %>"><% end %>
-  <% if entry.author.present? %><meta name="author" content="<%= entry.author %>"><% end %>
-  <% if entry.publisher.present? %><meta name="publisher" content="<%= entry.publisher %>"><% end %>
--- a/app/views/pageflow/entries/_meta_tags.html.erb
+++ b/app/views/pageflow/entries/_meta_tags.html.erb
@@ -1,0 +1,4 @@
+  <meta name="dc.language" content="<%= I18n.locale %>" />
+  <% if entry.keywords.present? %><meta name="keywords" content="<%= entry.keywords %>"><% end %>
+  <% if entry.author.present? %><meta name="author" content="<%= entry.author %>"><% end %>
+  <% if entry.publisher.present? %><meta name="publisher" content="<%= entry.publisher %>"><% end %>
--- a/app/views/pageflow/entries/show.html.erb
+++ b/app/views/pageflow/entries/show.html.erb
@@ -1,4 +1,4 @@
-<% @page_title = pretty_entry_title(@entry.title) %>
+<% @page_title = pretty_entry_title(@entry) %>
 
 <% content_for(:head) do %>
   <%= cache [@entry, :head, @entry.share_target] do %>
--- a/app/views/pageflow/entries/show.html.erb
+++ b/app/views/pageflow/entries/show.html.erb
@@ -1,4 +1,4 @@
-<% @page_title = @entry.title %>
+<% @page_title = pretty_entry_title(@entry.title) %>
 
 <% content_for(:head) do %>
   <%= cache [@entry, :head, @entry.share_target] do %>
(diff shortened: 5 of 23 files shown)


tilsammans:
I have thoughts about `_entry_meta_tags.html.erb` as well, but wanted to get your input on this change right away. Do you like this approach?

tilsammans:
@tf Would I have the go-ahead to introduce a `keywords` field per Entry? I would like to have the static ones `pageflow, multimedia, reportage` dynamic. If you agree, best to add it in the *General* section, or the *Social* section?

tilsammans:
@tf this pull request is ready for your review.

The keywords feels like a bigger issue so I would like to handle that separately.

tf:
I totally see the use case and really like the idea, but we have to do this in way that does not cause unwanted changes for some of our customers.

Some first thoughts:

Most of all, I'm not convinced that it's a good idea to reuse the users associated with an entry for the author field:
* Entry memberships are mainly an authorization mechanism. Just because I give one of my interns access to view a story does not mean he shall be mentioned as an author.
* On the other hand, if one of the authors happens to be an account admin, why do I have to keep adding her to entries she is contributing to even though her role gives her access anyway.
* Finally there may be authors that contribute content without even having a Pageflow account

I do not think these concepts really match up. 

Same goes for using the account name as publisher. Looking at Hosted Pageflow at least, account names already correspond to subdomains. But also more generally, account names need not reflect the legal name of the publishing organization.

What would you think about adding meta data text fields to the "General" tab, where this things could be edited? That would also be a good for the keywords. Maybe it might even be a good idea to inherit default values (for example for the publisher) from the `Theming`. That's the way it already works for the home button URL.

tilsammans:
I don't mind adding these meta data fields at all actually, although it could become a little crowded in there. Btw, _de Volkskrant_ is using the user-entry relationship to signify authors, but I agree the concept cannot be applied for all of Pageflow. I will make the required changes.

tf:
Maybe in a second step one could add another way to associate users with entries as authors. I think there are some very interesting features down that path, that could really help to increase the value of the user concept inside pageflow.

Things I could think of include:
* An auto generated list of authors with gravatars in the navigation credits box
* Links to other entries by the same author
* Ways to reach out to specific authors via their Twitter handle

Filling the meta data field is then really just one aspect.

To keep the form from being to crowded, we could add a checkbox "Custom Meta Data" or something which reveals the additional inputs.

I'll write up some hints with links to code examples which might be helpful. I realize much of the editor stuff is lacking documentation.

tilsammans:
I have to stop working on this feature for the rest of the day, but I think I'm on the right track with adding these fields to `Revision`. When I reload the editor they don't appear, but that's probably because they haven't been added to some JSON output. I'll investigate on Thursday.

tf:
Correct, the attributes have to be added to the [editor seed data](https://github.com/codevise/pageflow/blob/master/app/views/pageflow/editor/entries/seed.json.erb) of [the entry](https://github.com/codevise/pageflow/blob/master/app/views/pageflow/editor/entries/_entry.json.jbuilder).

The `input` method supports a `visibleBinding` option, which can be used to toggle the visibility of fields. For an example, look at the way [the home button check box toggles the home url text field](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js#L48). A corresponding `has_custom_meta_data` attribute would not need to be persisted but could initially be computed from the presence of any of the meta data attributes inside the [entry model](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/editor/models/entry.js). That way the section would always be expanded if custom values have been entered.

Default values from `Theming` attributes can be passed to the editor via the [theming seed data](https://github.com/codevise/pageflow/blob/master/app/views/pageflow/editor/themings/_theming.json.jbuilder). You can again look at the home url field for [an example of supplying a default value from the theming](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js#L47).

I'll add some more inline comments for minor issues.

Thanks for working on this!

app/models/pageflow/published_entry.rb
```diff
@@ -31,6 +33,10 @@ def initialize(entry, revision = nil)
       @custom_revision = !!revision
     end
 
+    def title
+      revision.title.presence || entry.title
+    end
+
```
tf: This is a good idea. What do you think about also displaying `entry.title` as an grey placeholder inside the title input field in the editor? That would make this logic clear for the editor.

app/helpers/pageflow/entries_helper.rb
```diff
@@ -1,5 +1,12 @@
 module Pageflow
   module EntriesHelper
+    def pretty_entry_title(entry)
+      [].tap do |title|
+        title << entry.title
+        title << entry.theming.cname_domain if entry.theming.cname_domain.present?
+      end.join(' - ')
+    end
+
```
tf: This can be simplified to `[entry.title, entry.themging.cname_domain.presence].compact.join(' - ')`, right?

app/views/pageflow/social_share/_entry_meta_tags.html.erb
```diff
@@ -2,10 +2,9 @@
 
 <%= social_share_entry_image_tags(entry) %>
 
-<%= tag :meta, :property => "og:title", :content => "#{entry.title} - #{entry.theming.cname_domain}" %>
+<%= tag :meta, :property => "og:title", :content => pretty_entry_title(entry) %>
 <%= tag :meta, :property => "og:description", :name => "description", :content => social_share_entry_description(entry) %>
 <%= tag :meta, :property => "og:url", :content => pretty_entry_url(entry) %>
-<%= tag :meta, :property => "og:site_name", :content => entry.theming.cname_domain %>
```
tf: Should this really be deleted?

config/locales/de.yml
```diff
@@ -169,15 +169,18 @@ de:
         title: Titel
```
tf: Please follow [this procedure for new translations](https://github.com/codevise/pageflow/wiki/Adding-new-Translation-Keys).

app/views/pageflow/entries/show.html.erb
```diff
@@ -33,3 +33,7 @@
 
   <%= render 'pageflow/entries/analytics', :entry => @entry %>
 <% end %>
+
+<% content_for :author, @entry.author %>
+<% content_for :publisher, @entry.publisher %>
+<% content_for :keywords, @entry.keywords %>
```
tf: Let's go for the breaking change and no longer  support [overriding `_meta_tags.html.erb](https://github.com/codevise/pageflow/wiki/Customizing-Views) now that we get a better mechanism.

Then we are free to move it to `pageflow/entries/_meta_tags.html.erb` render it from the `content_for(:head)` here and pass the entry explicitly as a local.

app/views/pageflow/social_share/_entry_meta_tags.html.erb
tilsammans: In my opinion, it does not follow the [OpenGraph site_name spec](http://ogp.me):

> If your object is part of a larger web site, the name which should be displayed for the overall site. e.g., "IMDb".

But it's not a deal breaker for me to keep it in. :)

app/models/pageflow/published_entry.rb
tilsammans: Great idea

app/views/pageflow/social_share/_entry_meta_tags.html.erb
tf: Always scared to change a winning team ;) For what it's worth, also not really related to this change.

tilsammans:
@tf I made the changes, review please?

I decided against the visibility checkbox after all. The UI does not look cluttered to me at all, in fact it's rather nice. If anything we should create a new tab section labeled `Meta` to contain these fields.

I also decided against filling in default publisher name from the `Theming`. If you don't use it like this, it will only annoy people. So these new meta fields will default to no value. If you prefer otherwise, let me know.

app/models/pageflow/draft_entry.rb
```diff
@@ -28,6 +29,11 @@ def initialize(entry, draft = nil)
       @draft = draft || entry.draft
     end
 
+    # So we can always get to the original Entry title.
+    def entry_title
+      entry.title
+    end
+
```
tf: This collides with the delegation to revision. This should probably behave like PublishedEntry. Maybe the test for DraftEntry can be extracted to a shared example group and be reused?

app/models/pageflow/published_entry.rb
```diff
@@ -11,17 +11,19 @@ class PublishedEntry
              :enabled_feature_names,
              :to_model, :to_key, :persisted?,
              :authenticate,
+             :users,
```
tf: This a left-over of the first implementation, right?

app/views/pageflow/entries/_meta_tags.html.erb
```diff
@@ -0,0 +1,4 @@
+  <meta name="dc.language" content="<%= I18n.locale %>" />
+  <% if entry.keywords.present? %><meta name="keywords" content="<%= entry.keywords %>"><% end %>
+  <% if entry.author.present? %><meta name="author" content="<%= entry.author %>"><% end %>
+  <% if entry.publisher.present? %><meta name="publisher" content="<%= entry.publisher %>"><% end %>
```
tf: Before, one could override the partial and set new defaults. Now everything is very configurable, but there is no longer a way to define default values. If the Theming integration is out of scope for you now, we could alternatively allow setting defaults in `Pageflow.conig` and fall back to these here. For the editor, we can pass values from the config in the [config seed data](https://github.com/codevise/pageflow/blob/master/app/views/pageflow/config/_editor_seeds.json.jbuilder) which are accessible in the editor [via `pageflow.config`](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/editor/initializers/setup_config.js). Then we can display placeholders in the text boxes.

tf: Let's also add a test this logic with fallback to config settings works. It can be a [controller test for `EntriesController#show`](https://github.com/codevise/pageflow/blob/master/spec/controllers/pageflow/entries_controller_spec.rb#L164). That spec is already a bit longer than I'm comfortable with, but I guess the cases covered there are still simple enough.

tf: Alternatively the render call for this template could be wrapped in a helper. Then we could test the view logic based on this helper. That's the way it works for the [SocialShareHelpers](https://github.com/codevise/pageflow/blob/master/spec/helpers/pageflow/social_share_helper_spec.rb).

config/locales/de.yml
```diff
@@ -430,7 +430,7 @@ de:
       long: ! '%e. %B %Y'
       short: ! '%e. %b'
     month_names:
-    - 
+    -
     - Januar
```
tf: Can you remove these changes? The files are generated by localeapp. It will always insert the trailing space again on next update.

app/models/pageflow/draft_entry.rb
tilsammans: It's not delegating `title`, this is a new method `entry_title` which always points at the entry title. I did this so that you can overwrite in the draft, because it's only used for the placeholder.

Do you have ideas to do this differently? The `title` is still delegated, that hasn't changed.

tilsammans:
I'm stuck at the breaking change for the meta partial; I don't fully understand the direction that should be taken here. Can we take a step back and reduce scope of the change somehow?

The other small remarks are done.

app/models/pageflow/draft_entry.rb
tf: Ah my bad. Somehow missed the point here.

tf:
I'd suggest the following:

* Add a new helper module `Pageflow::MetaTagsHelper` with a `meta_tags_for_entry` method.
* Instead of calling `render 'pageflow/entries/meta_tags', :entry => @entry` in `show.html.erb`, call that method passing the entry.
* Move the meta tags partial to `pageflow/meta_tags/_entry.html.erb` since it now belongs to the helper.
* Pass `keywords`, `author`, `publisher` locals to the partial instead of the whole `entry`. That way the logic can stay inside the helper.
* Add `default_keywords_meta_tag`, `default_author_meta_tag`, `default_publisher_meta_tag` attributes to `Pageflow::Configuration` and let them default to the values that were originally hardcoded inside `_meta_tags.html.erb`.
* In the `meta_tags_for_entry` method fall back to those global defaults if entry specific values are missing.
* Add a test for the new helper method checking that this fallback works. You can do it exactly like [the test for `social_share_entry_image_tags`](https://github.com/codevise/pageflow/blob/master/spec/helpers/pageflow/social_share_helper_spec.rb#L100).

That way the only breaking change is that the `_meta_tags.html.erb` partial cannot be overriden anymore, but instead things need to happen inside `Pageflow.config`.

What do you think?

tf:
Passing the defaults to the editor inside the config seed, to display them as defaults would be a nice extra.

tilsammans:
@tf made the changes. final review? (I can squish after)

tilsammans:
The [Travis error](https://travis-ci.org/codevise/pageflow/jobs/119786007) is unrelated to this PR.

tf:
Looks really good. The `meta_tags_for_entry` spec could test that values configured in the entry actually win against defaults. But other than that I'd say it's ready for merge. If you rebase onto master, the spec failure should be gone.

tilsammans:
Closing this in favour of #508

PR #500, opened 2016-03-02
ActiveAdmin/Devise Update

tf:
Try using upstream `master` instead of old `rails4` branch from our fork.

### DIFF
--- a/app/assets/stylesheets/pageflow/admin/tabs_view.css.scss
+++ b/app/assets/stylesheets/pageflow/admin/tabs_view.css.scss
@@ -18,9 +18,9 @@
       display: inline-block;
 
       &.active {
-        @include gradient(#f4f4f4, #fff);
+        background-color: #fff;
         color: #444;
-        box-shadow: 1px 1px 1px 0 rgba(0,0,0,0.1) inset;
+        box-shadow: 1px 1px 1px 0 rgba(0, 0, 0, 0.2) inset;
 
         a {
           cursor: default;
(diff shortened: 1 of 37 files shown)


tilsammans:
Good stuff, this is going to allow us to jump to Ruby 2.3.0 as well.

Do you plan on following ActiveAdmin `master` or will the release be tied to a specific version?

tf:
If 1.0 should be out soon, I'd suggest depending on `~> 1.0`. Otherwise, not sure. Every application would be free to pin down a specific version in its `Gemfile`. In the Readme it would probably be good to mention a specific version/commit ref that is known to work. But for CI it might be also be good to know if further breaking changes are introduced on the way to 1.0.

tilsammans:
Yes, agreed. The tilde operator should be the best compromise, I think.

Will there be another pre-1.0 release? Do you have a rough idea of planning?

tf:
I don't know. The [1.0 milestone](https://github.com/activeadmin/activeadmin/milestones/1.0.0) does not have many issue left, but then again it's [two years overdue](https://github.com/activeadmin/activeadmin/milestones). :wink:

app/assets/stylesheets/pageflow/admin/tabs_view.css.scss
```diff
@@ -18,9 +18,9 @@
       display: inline-block;
 
       &.active {
-        @include gradient(#f4f4f4, #fff);
+        background-color: #fff;
```
tilsammans: wow @houndci-bot that's fiercely strict!

PR #504, opened 2016-03-08
Extend permissions model

aviav:
Make memberships' entity polymorphic, require role

Memberships can now refer to accounts as well as entrys via the
polymorphic entity attribute, which replaces the entry attribute.

Memberships now also have a non-null role attribute which defaults to
'editor'. This attribute will also hold the values 'admin', 'publisher'
and 'reader' as soon as the permissions model is replaced.

Memberships as of this commit cannot be created or deleted in the admin
interface via their user. Ways to facilitate a desired interaction
between Formtastic, ActiveAdmin and the polymorphic entity attribute
might be found at
https://github.com/activeadmin/activeadmin/wiki/Nested-forms-with-polymorphic-association-in-Active-Admin-Formtastic
.


spec/support/dominos/admin/page.rb
```diff
@@ -21,11 +21,27 @@ def self.sign_in(options)
...
+        if ['previewer', 'editor', 'publisher', 'manager'].include?(role)
+          user = FactoryGirl.create(:user, options.reverse_merge(email: email,
+                                                                 password: password))
+        else
+          user = FactoryGirl.create(:user, role, options.reverse_merge(email: email,
+                                                                       password: password))
+        end
 
-        user = FactoryGirl.create(:user, role, options.reverse_merge(:email => email, :password => password))
+        unless role == :editor || on.blank?
```
tf: `unless` with further logical operators are hard to understand. Prefer `if`. Why is editor not allowed here?

spec/features/account_manager/duplicating_an_entry_spec.rb
```diff
@@ -1,15 +1,14 @@
 require 'spec_helper'
 
-feature 'account manager duplicating an entry' do
-  scenario 'from own account' do
-    account = create(:account)
-    entry = create(:entry, :title => 'Test Entry', :account => account)
+feature 'duplicating an entry' do
+  scenario 'as publisher of entry' do
+    user = Dom::Admin::Page.sign_in_as(:editor)
```
tf: For cases like this a simple `Dom::Admin::Page.sign_in` method would come in handy. Editor's used to be the unprivileged ones, but right now we really have not good way to create a unprivileged user.

spec/controllers/admin/entries_controller_spec.rb
```diff
@@ -455,20 +458,21 @@ def self.name
...
         post(:duplicate, :id => entry.id)
       }.not_to change { Pageflow::Entry.count }
     end
 
-    it 'allows account manager to duplicate entries of own account' do
-      user = create(:user, :account_manager)
-      entry = create(:entry, :account => user.account)
+    it 'allows entry publisher to duplicate entry' do
+      user = create(:user)
+      entry = create(:entry, with_publisher: user)
```
tf: Let's rename these factory attributes to `with_publisher_member`, `with_editor_member` etc. Especially `with_publisher` sounds a lot like the user actually already published the entry. What do you think?

lib/pageflow/policies/entry_policy.rb
```diff
@@ -1,11 +1,43 @@
...
+        attr_reader :user, :scope
+
+        def initialize(user, scope)
+          @user = user
+          @scope = scope
+        end
+
+        def resolve
+          scope.joins(ActiveRecord::Base.send(
+                        :sanitize_sql_array, [
```
tf: We should extract a private method here to make it more readable.

spec/controllers/admin/entries_controller_spec.rb
```diff
@@ -155,8 +161,10 @@ def self.name
       expect(Pageflow::Entry.last.custom_field).to eq('some value')
     end
 
-    it 'does not allows account manager to define custom field not registered as form input' do
-      user = create(:user, :account_manager)
+    it 'does not allows account publisher to define custom field not registered as form input' do
+      user = create(:user)
+      account = create(:account)
+      create(:membership, user: user, entity: account, role: 'publisher')
```
tf: Using the `with_publisher_member` factory attributes would make this more concise.

spec/support/dominos/admin/page.rb
```diff
@@ -27,7 +27,10 @@ def self.sign_in_as(role, options = {})
                options.delete(:on)
              end
 
-        if ['previewer', 'editor', 'publisher', 'manager'].include?(role)
+        if ['previewer', :previewer,
+            'editor', :editor,
+            'publisher', :publisher,
+            'manager', :manager].include?(role)
```
tf: `role.to_sym`?

spec/features/account_manager/duplicating_an_entry_spec.rb
aviav: The way to create an unprivileged user is now to say sign_in_as(:editor), without passing a pair with key 'on' in the options hash. I just adjusted it so that this should now work exactly as it did before. sign_in_as(:account_manager) also works as before, however sign_in_as(:account_manager, account: account) will soon not work anymore because the account Membership replaces the account attribute. It has to be replaced by sign_in_as(:account_manager, on: account) or even finally by sign_in_as(:manager, on: account) if the role attribute on user can be let alone.

admins/pageflow/entry.rb
```diff
@@ -33,7 +33,10 @@ module Pageflow
     sidebar :folders, :only => :index do
       text_node(link_to(I18n.t('pageflow.admin.entries.add_folder'), new_admin_folder_path, :class => 'new'))
       grouped_folder_list(Folder.accessible_by(Ability.new(current_user), :read),
-                          :class => authorized?(:manage, Folder) ? 'editable' : nil,
+                          class: if params[:id] &&
+                                    authorized?(:configure_folder_on, resource.account)
+                                   'editable'
+                                 end,
```
tf: The grouped folder list is only every displayed on the index page. `params[:id]` will never be present. `editable` will have to be decided inside the grouped folder list for each account separately.

admins/pageflow/entry.rb
```diff
@@ -48,7 +51,7 @@ module Pageflow
             f.input :theming, :include_blank => false
           end
         end
-        if authorized?(:manage, Folder)
+        if authorized?(:configure_folder_for, resource || resource_class)
```
tf: If `resource` is `nil`, the check `authorized?(:configure_folder_for,, Entry)` will be performed, which will always be true. Can we make it more explicit, that the folder fields should always be visible if we are creating a new entry (which is the only case where `resource` can be `nil` here).

admins/pageflow/entry.rb
```diff
@@ -153,7 +156,10 @@ def permitted_attributes
         result += Pageflow.config_for(target).admin_form_inputs.permitted_attributes_for(:entry)
 
         result += [:account_id, :theming_id] if authorized?(:read, Account)
-        result << :folder_id if authorized?(:manage, Folder)
+        if params[:id] && authorized?(:configure_folder_for, resource || resource.class)
+          result << :folder_id
+        end
```
tf: If `params[:id]` is present, `resource` will either return a record or cause a not found exception. The `resource.class` case should never occur.

app/views/components/pageflow/admin/members_tab.rb
```diff
@@ -24,7 +24,7 @@ def build(entry)
             end
           end
         end
-        if authorized? :manage, Pageflow::Entry
+        if authorized? :add_member_to, Pageflow::Entry
```
tf: This will always be true since CanCan does not check the block if we pass a class instead of a record.

lib/pageflow/ability_mixin.rb
```diff
@@ -5,38 +5,70 @@ module AbilityMixin
...
-        can_edit_entry?(user, revision.entry)
-      end
+        can [:edit,
+             :update,
+             :edit_outline,
+             :publish,
+             :restore,
+             :snapshot,
+             :confirm_encoding], Entry do |entry|
+          can_edit_entry?(user, entry)
```
tf: This still has to be migrated, right?

lib/pageflow/ability_mixin.rb
```diff
@@ -5,38 +5,70 @@ module AbilityMixin
...
+      can :view, [Admin::MembersTab, Admin::RevisionsTab]
+      unless user.admin?
+        can [:create, :update, :destroy], Folder do |folder|
+          Policies::FolderPolicy.new(user, folder).manage?
+        end
 
-      can [:read, :use_files], Entry, :id => user.entry_ids
+        can :configure_folder_on, Account do |account|
+          Policies::AccountPolicy.new(user, account).publish?
+        end
```
tf: I'd suggest adding alias methods to the policies which reflect the permission we are trying to check. Otherwise we leak implementation details that the permission to publish coincides with the permission to configure folders. If this changes in the future, we only want to have to make changes to the policy and controller tests - not glue code like this one found here.

lib/pageflow/ability_mixin.rb
```diff
@@ -5,38 +5,70 @@ module AbilityMixin
...
-      end
+        can :read, Entry, Policies::EntryPolicy::Scope.new(user, Entry).resolve do |entry|
+          Policies::EntryPolicy.new(user, entry).read?
+        end
 
-      can :manage, Page do |page|
-        can_edit_entry?(user, page.chapter.entry)
-      end
+        can [:create, :duplicate], Entry do |entry|
+          Policies::EntryPolicy.new(user, entry).publish?
```
tf: see above

lib/pageflow/policies/entry_policy.rb
```diff
@@ -0,0 +1,71 @@
...
+            'pageflow_memberships.entity_type = "Pageflow::Entry"',
+            user_id: user.id])).joins(sanitize_sql_array([
+              'LEFT OUTER JOIN pageflow_memberships as pageflow_memberships_2 ON ' \
+              'pageflow_memberships_2.user_id = :user_id AND ' \
+              'pageflow_memberships_2.entity_id = pageflow_entries.account_id AND ' \
+              'pageflow_memberships_2.entity_type = "Pageflow::Account"',
+              user_id: user.id])).where(
+                'pageflow_memberships.entity_id IS NOT NULL OR ' \
+                'pageflow_memberships_2.entity_id IS NOT NULL')
+        end
```
tf: This is still a bit hard to read. What do you think about refactoring it to something like

```
  scope
    .joins(memberships_for_entries(user))
    .joins(memberships_for_account_of_entries(user))
    .where(either_membership_is_present)
```

lib/pageflow/policies/folder_policy.rb
```diff
@@ -0,0 +1,51 @@
...
+        end
+
+        def accounts_where_user_is_at_least_publisher(user)
+          user.membership_accounts.joins(sanitize_sql_array([
+            'LEFT OUTER JOIN pageflow_memberships as pageflow_memberships_2 ON ' \
+            'pageflow_memberships_2.user_id = :user_id AND ' \
+            'pageflow_memberships_2.entity_type = "Pageflow::Account" AND ' \
+            'pageflow_memberships_2.entity_id = pageflow_accounts.id AND ' \
+            'pageflow_memberships_2.role IN ("publisher", "manager")',
+            user_id: user.id])).where('pageflow_memberships_2.entity_id IS NOT NULL')
```
tf: why the `_2` suffix?

spec/pageflow/policies/account_policy_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+module Pageflow
+  module Policies
+    describe AccountPolicy do
+      terms = {class: AccountPolicy,
+               subject: FactoryGirl.create(:account),
```
tf: Creating models outside of test runs does not work correctly with database resets in before hooks.

spec/pageflow/policies/entry_policy_spec.rb
```diff
@@ -0,0 +1,123 @@
+require 'spec_helper'
+
+module Pageflow
+  module Policies
+    describe EntryPolicy do
+      terms = {class: EntryPolicy,
+               subject: FactoryGirl.create(:entry),
```
tf: see above

spec/pageflow/policies/folder_policy_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+
+module Pageflow
+  module Policies
+    describe FolderPolicy do
+      terms = {class: FolderPolicy,
+               subject: FactoryGirl.create(:folder),
```
tf: see above

spec/pageflow/policies/folder_policy_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      terms = terms.merge!(entity: terms[:subject].account)
+      describe :manage do
+        it_behaves_like 'a membership-based permission that',
+                        terms: terms,
+                        entity_type: :account,
+                        permission_type: :manage,
+                        minimum_required_role: 'publisher',
+                        maximum_forbidden_role: 'editor'
+      end
+    end
```
tf: I still find this a bit hard to read. Especially the `terms` hash makes it more opaque. According to [RSpec docs](https://www.relishapp.com/rspec/rspec-core/v/2-0/docs/example-groups/shared-example-group#shared-example-group-applied-to-two-groups) the `described_class` is visible inside shared examples. So no  need to pass the policy. What do you think of making the parameters read more like a sentence. That would also work nicely with the trailing "that" in the description:

```
describe FolderPolicy do
   it_behaves_like 'a membership-based permission that',
                            allows: 'publisher',
                            but_forbids: 'editor',
                            of: -> { |subject| subject.account }
                            to: :manage,
                            subject: -> { create(:folder) }
end
```

That way we make the policy subject explicit. Moreover, we make explicit how the membership entity shall be obtained from the subject. `name` parameter can be obtained from `params[:subject].class.name`. The entity name is `params[:of].class.name`

spec/pageflow/policies/entry_policy_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      describe :edit do
+        include_examples 'for entry and account',
+                         terms: terms,
+                         permission_type: :edit,
+                         minimum_required_role: 'editor',
+                         maximum_forbidden_role: 'previewer'
+      end
+
+      describe :preview do
+        include_examples 'for entry and account',
```
tf: this shared example name is too global. If I understand `include_examples` correctly the description is not visible in the resulting spec description, right? So we should b free to choose a more descriptive name.

admins/pageflow/entry.rb
aviav: This probably means that there are no working spec tests for anything that relies on 'editable'. I would have expected this one here from managing_folders_spec.rb to fail:
```ruby
  scenario 'renaming a folder' do
    folder = create(:folder, name: 'A folder')
    Dom::Admin::Page.sign_in_as(:publisher, on: folder.account)

    visit(admin_entries_path)
    Dom::Admin::FolderPanelItem.find_by_name('A folder').edit_link.click
    Dom::Admin::FolderForm.first.submit_with(:name => 'Renamed folder')

    expect(Dom::Admin::FolderPanelItem.find_by_name('Renamed folder')).to be_present
  end
```

lib/pageflow/ability_mixin.rb
aviav: Among others, yes.

lib/pageflow/policies/folder_policy.rb
aviav: Because the user gets their accounts via pageflow_memberships also, requiring a way to distinguish the columns

spec/pageflow/policies/folder_policy_spec.rb
aviav: Can we make the anonymous function take the environment of the shared example? Otherwise the lambda construct acts as a closure, and will for that reason not work as intended

tf:
See #634

PR #508, opened 2016-04-01
Make the html meta tags more configurable

tilsammans:
* allow meta author, keywords and publisher to be set per entry
* allow meta author, keywords and publisher defaults to be configured

Also moves rendering of the meta tags into a helper, where we
encapsulate the logic of rendering the values or falling back to the
configured defaults.

This is a breaking change for folks who have overriden the
app/views/layouts/pageflow/_meta_tags.html.erb partial!

Fixes codevise/pageflow#499


tf:
I much prefer the new `pageflow_configure` based specs since they are less coupled to the concrete default configuration values. But so far there is still no spec ensuring that _entry specific_ overrides take effect. I could erase `entry.keywords.presence ||` etc from the helper and still get a passing test suite.

tilsammans:
@tf all done & a little spec file cleanup too

PR #509, opened 2016-04-03
use <tt>render :partial</tt> when rendering a partial

tilsammans:
Even though <tt>render :template</tt> works, the templates that are
being referenced are partials. By using the recommended API we reduce
non-standard things in the code.

I also suspect, but cannot proof, that render partial behaves slightly
different with the locals hash.


tf:
For built in page types this holds true, but for plugin based page types deriving from [`PageType`](https://github.com/codevise/pageflow/blob/master/lib/pageflow/page_type.rb#L5) the template is actually located at `pageflow/somthing/page.erb.html` (i.e. in the [`pageflow-panorama` gem](https://github.com/codevise/pageflow-panorama/tree/master/app/views/pageflow/panorama)). 

I can't really say what the objective of not using a partial here originally was, but far as I can tell this change would break compatibility. The fact that the tests still pass indicate a lack of coverage regarding this plugin interface.

tilsammans:
Okay, clear. I will raise an issue to improve test coverage.

PR #511, opened 2016-04-05
introduce .logo_header

tilsammans:
In the current theme, the logo header on the first page is styled using
a CSS pseudo element. This can be troublesome to work with, since the
element does not have a class.

It is also troublesome to discover that this element exists and change
it according to one's own needs. The CSS-rule is quite exact and does
not allow for much flexibility.

This patch aims to change that.

By introducing a new div with class name "logo_header" it becomes much
easier to style this element using CSS, and to override it in a custom
theme. The element becomes easier to inspect in the browser and change
some style attributes there. Lastly it's evident what this element does,
and ties in directly with the image name that's already in use.

It does mean that there is now a DIV in each page. This is an acceptable
tradeoff to me; in fact I prefer this since it will allow me to style
the header on the last page too. It gives me more power and flexibility.

@tf thoughts?

### DIFF
--- a/app/assets/stylesheets/pageflow/themes/default/page.css.scss
+++ b/app/assets/stylesheets/pageflow/themes/default/page.css.scss
@@ -74,8 +74,7 @@ $page-color: $main-color;
 
 .js .page:first-child {
 
-  .content_and_background .scroller > div:after {
-    content: " ";
+  .logo_header {
     position: absolute;
     background-image: image-url("pageflow/themes/#{$theme-name}/logo_header.png");
     background-repeat: no-repeat;
@@ -95,7 +94,7 @@ $page-color: $main-color;
     }
   }
 
-  &.invert .content_and_background .scroller > div:after {
+  &.invert .logo_header {
     background-image: image-url("pageflow/themes/#{$theme-name}/logo_header_invert.png");
   }
 
--- a/app/views/pageflow/pages/templates/_audio.html.erb
+++ b/app/views/pageflow/pages/templates/_audio.html.erb
@@ -50,6 +50,7 @@
     </div>
     <div class="scroller">
       <div>
+        <div class="logo_header"></div>
         <div class="contentWrapper">
           <div class="contentText audioPage">
             <p class="non_js_audio">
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/pageflow/themes/default/page.css.scss
```diff
@@ -74,8 +74,7 @@ $page-color: $main-color;
 
 .js .page:first-child {
 
-  .content_and_background .scroller > div:after {
-    content: " ";
+  .logo_header {
```
tilsammans: @tf what does that mean?

tf:
I'm rather skeptical of this change for several reasons:

* wide spread change: to make this work, we would have to amend all page templates in all page type plugins
* semantic dom structure: even if it is just a div, semantically the header logo does not really feel like a part of the page to me
* limited gains in flexibility: in the past we have found the need to turn the logo into a link (see below). That would still not be possible with this solution, but the concrete DOM to represent the logo would be duplicated across a lot of gems.
* inconsistency with existing solutions (see below)

So far, the approach to have a logo with DOM representation has been to override the `_header.html.erb` partial and place a `div.header_logo` inside there. The [header widget](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/widgets/header.js#L30) then already takes care of dynamically applying helpful css classes like `invert` or `first_page`.

The new [configurable default theme](https://github.com/codevise/pageflow/pull/470), which we are already using in a couple of production instances, will provide a lot more flexibility when it comes to choosing or adding new [ways to display the logo](https://github.com/codevise/pageflow/pull/470/files#diff-d33ef1a0ea695494fbef39c4f0ae6c31).

So instead, I'd consider adding a hidden `div.header_logo` to the `_header.html.erb` partial, such that the theme configuration does not depend on a correctly overriden partial.

app/assets/stylesheets/pageflow/themes/default/page.css.scss
tf: I guess there are 12 properties assigned in this rule, while scsslint thinks there should be at most 10. The limit seems a bit arbitrary. Then again, it is quite a long rule.

tilsammans:
All fair points. For me working with widgets is still very opaque and often surprising in how it behaves. It's _much_ simpler to just hack a view in place and work with actual things I can control. I've often had to go into the production DB to nuke widgets that kept coming back from revision to revision, so my experience with them has not been great so far.

I agree that the implications of this change are quite far-reaching. I am not sure how I will be able to work with the procedure you've outlined but I'll take it one step at a time.

PR #535, opened 2016-06-13
Add config option to forbid deleting user account

aviav:
In some cases it might make sense to prevent the last user of an account from loosing his ability to sign in. This configuration option leaves the details of this policy to the host application.

### DIFF
--- a/admins/pageflow/user.rb
+++ b/admins/pageflow/user.rb
@@ -120,7 +120,8 @@ module Pageflow
 
     collection_action 'delete_me', :title => I18n.t('pageflow.admin.users.account'), :method => [:get, :delete] do
       if request.delete?
-        if current_user.destroy_with_password(params.require(:user)[:current_password])
+        if authorized?(:delete_own_user, current_user) &&
+           current_user.destroy_with_password(params.require(:user)[:current_password])
           redirect_to admin_root_path, :notice => I18n.t('pageflow.admin.users.me.updated')
         end
       end
--- a/app/helpers/pageflow/admin/users_helper.rb
+++ b/app/helpers/pageflow/admin/users_helper.rb
@@ -4,6 +4,15 @@ module Pageflow
       def collection_for_user_roles
         User.roles_accessible_by(current_ability).index_by { |role| t(role, :scope => 'pageflow.admin.users.roles') }
       end
+
+      def delete_own_user_section
+        user_deletion_permission = Pageflow.config.authorize_user_deletion.call(current_user)
+        if user_deletion_permission == true
+          render('pageflow/admin/users/may_delete')
+        else
+          render('pageflow/admin/users/cannot_delete', reason: user_deletion_permission)
+        end
+      end
     end
   end
 end
--- a/app/views/admin/users/me.html.erb
+++ b/app/views/admin/users/me.html.erb
@@ -1,7 +1,7 @@
 <%= semantic_form_for current_user, :url => me_admin_users_path, :html => {:class => 'profile'} do |form| %>
   <%= form.inputs :first_name, :last_name %>
 
-  <%= form.inputs  do %>
+  <%= form.inputs do %>
     <%= form.input :locale,
                    as: :select,
                    include_blank: false,
@@ -13,9 +13,8 @@
     <%= form.input :password %>
     <%= form.input :password_confirmation %>
   <% end %>
+
   <%= form.actions %>
-  <p>
-    <%= t('pageflow.admin.users.me.delete_account_hint') %>
-    <%= link_to(t('pageflow.admin.users.me.delete_account'), delete_me_admin_users_path, :class => 'delete_account') %>
-  </p>
+
+  <%= delete_own_user_section %>
 <% end %>
--- a/app/views/pageflow/admin/users/_may_delete.html.erb
+++ b/app/views/pageflow/admin/users/_may_delete.html.erb
@@ -1,0 +1,6 @@
+<p class="may_delete">
+  <%= t('pageflow.admin.users.me.delete_account_hint') %>
+    <%= link_to(t('pageflow.admin.users.me.delete_account'),
+                main_app.delete_me_admin_users_path,
+                class: 'delete_account') %>
+</p>
--- a/lib/pageflow/ability_mixin.rb
+++ b/lib/pageflow/ability_mixin.rb
@@ -35,6 +35,10 @@ module Pageflow
 
       can :view, [Admin::MembersTab, Admin::RevisionsTab]
 
+      can :delete_own_user, ::User do |user_to_delete|
+        Pageflow.config.authorize_user_deletion.call(user_to_delete) == true
+      end
+
       if user.admin?
         can [:read, :create, :update], Account
         can :destroy, Account do |account|
--- a/lib/pageflow/configuration.rb
+++ b/lib/pageflow/configuration.rb
@@ -228,6 +228,21 @@ module Pageflow
     attr_accessor :default_author_meta_tag
     attr_accessor :default_publisher_meta_tag
 
+    # Whether a user can be deleted.
+    #
+    # @example
+    #
+    #     config.authorize_user_deletion =
+    #       lambda do |user_to_delete|
+    #         if user_to_delete.account.users.length > 1
+    #           true
+    #         else
+    #           'Last user on account. Permission denied'
+    #         end
+    #       end
+    # @since edge
+    attr_accessor :authorize_user_deletion
+
     def initialize
       @paperclip_filesystem_default_options = {validate_media_type: false}
       @paperclip_s3_default_options = {validate_media_type: false}
@@ -265,6 +280,8 @@ module Pageflow
       @default_keywords_meta_tag = 'pageflow, multimedia, reportage'
       @default_author_meta_tag = 'Pageflow'
       @default_publisher_meta_tag = 'Pageflow'
+
+      @authorize_user_deletion = lambda { |_user| true }
     end
 
     # Activate a plugin.
--- a/spec/controllers/admin/users_controller_spec.rb
+++ b/spec/controllers/admin/users_controller_spec.rb
@@ -146,5 +146,32 @@ module Pageflow
         expect(user.reload.account).to eq(other_account)
       end
     end
+
+    describe '#delete_me' do
+      it 'allows to destroy the user by default' do
+        sign_in(create(:user, password: '@qwert123'))
+
+        expect do
+          delete(:delete_me, user: {current_password: '@qwert123'})
+        end.to change { User.count }
+      end
+
+      it 'does not allow to destroy the user when authorize_user_deletion non-true' do
+        user = create(:user, password: '@qwert123')
+        sign_in(user)
+        Pageflow.config.authorize_user_deletion =
+          lambda do |user_to_delete|
+            if user_to_delete.account.users.length > 1
+              true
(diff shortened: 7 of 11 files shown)


admins/pageflow/user.rb
```diff
@@ -120,7 +120,8 @@ module Pageflow
 
     collection_action 'delete_me', :title => I18n.t('pageflow.admin.users.account'), :method => [:get, :delete] do
       if request.delete?
-        if current_user.destroy_with_password(params.require(:user)[:current_password])
+        if Pageflow.config.authorize_user_deletion.call == true &&
```
tf: Let's use `authorized?` here and move the call of the configured lambda to the `AbilityMixin` (and later on to a policy).

app/views/admin/users/me.html.erb
```diff
@@ -15,7 +15,11 @@
...
   <%= form.actions %>
   <p>
-    <%= t('pageflow.admin.users.me.delete_account_hint') %>
-    <%= link_to(t('pageflow.admin.users.me.delete_account'), delete_me_admin_users_path, :class => 'delete_account') %>
+    <% if Pageflow.config.authorize_user_deletion.call == true %>
+      <%= t('pageflow.admin.users.me.delete_account_hint') %>
+        <%= link_to(t('pageflow.admin.users.me.delete_account'), delete_me_admin_users_path, :class => 'delete_account') %>
+    <% else %>
+        <%= Pageflow.config.authorize_user_deletion.call %>
+    <% end %>
```
tf: Let's move this to a helper rendering a partial. That way we can also avoid calling the lambda multiple times as it can potentially perform expensive queries. Also let's apply different class names to the surrounding paragraph tag in each case. That way we can style the message to include a warning icon etc when deleting is forbidden.

lib/pageflow/configuration.rb
```diff
@@ -228,6 +228,8 @@ class Configuration
     attr_accessor :default_author_meta_tag
     attr_accessor :default_publisher_meta_tag
 
+    attr_accessor :authorize_user_deletion
+
```
tf: Please add inline help describing the option and the lambda signature. Also add a `@since edge` doc tag. That will automatically be rewritten to the correct version when the gem is released.

admins/pageflow/user.rb
tf: Also let's always pass a `user` parameter. Otherwise there is no way for implementors to know which user we are talking about.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -146,5 +146,27 @@ module Pageflow
         expect(user.reload.account).to eq(other_account)
       end
     end
+
+    describe '#delete_me' do
+      it 'allows to destroy the user when authorize_user_deletion true' do
```
tf: I'd rephrase that to "by default". That way we do not leak to much implementation detail into spec description. The actual default setup might change in the future.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -146,5 +146,27 @@ module Pageflow
...
+          delete(:delete_me, user: {current_password: '@qwert123'})
+        end.to change { User.count }
+      end
+
+      it 'does not allow to destroy the user when authorize_user_deletion non-true' do
+        user = create(:user, password: '@qwert123')
+        account_users_count = user.account.users.length
+        sign_in(user)
+        Pageflow.config.authorize_user_deletion =
+          lambda { account_users_count > 1 ? true : 'Last user on account. Permission denied' }
```
tf: Normally I'd say, simply return `false` to make the test as simple as possible. On the other hand, we need to test that the block gets passed a user record. So performing an actual check inside the lambda is one way of expressing that while also demonstrating real world usage. The alternative would be to pass a mock and explicitly expect that `call` is invoked with given arguments.

admins/pageflow/user.rb
aviav: Fixed

app/views/admin/users/me.html.erb
aviav: Done

lib/pageflow/configuration.rb
aviav: Added.

spec/controllers/admin/users_controller_spec.rb
aviav: Rephrased.

app/helpers/pageflow/admin/users_helper.rb
```diff
@@ -4,6 +4,14 @@ module UsersHelper
       def collection_for_user_roles
         User.roles_accessible_by(current_ability).index_by { |role| t(role, :scope => 'pageflow.admin.users.roles') }
       end
+
+      def delete_me_para
+        if Pageflow.config.authorize_user_deletion.call(current_user) == true
+          render('may_delete')
+        else
+          render('cannot_delete')
+        end
```
tf: * Let's save the result of `call` here and pass it to the partial as local variable. That way we do not call the lambda multiple times and do not risk calling it without parameters in some cases (see below)
* To catch such errors, let's add a test which covers the two cases and makes sure the returned string is rendered.
* Since all helpers share a global scope, helper method names should include the name of their module, i.e. `delete_own_user_section` etc.

lib/pageflow/ability_mixin.rb
```diff
@@ -35,6 +35,10 @@ def pageflow_default_abilities(user)
 
       can :view, [Admin::MembersTab, Admin::RevisionsTab]
 
+      can :delete_own, ::User do |current_user|
+        Pageflow.config.authorize_user_deletion.call(current_user) == true
+      end
+
```
tf: No need to pass in `current_user` here. Let's call the ability `:delete_own_user`.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -146,5 +146,32 @@ module Pageflow
...
+            else
+              'Last user on account. Permission denied'
+            end
+          end
+
+        expect do
+          delete(:delete_me, user: {current_password: '@qwert123'})
+        end.not_to change { User.count }
+      end
+    end
```
tf: I like the real life example here.

app/helpers/pageflow/admin/users_helper.rb
aviav: Fixed

lib/pageflow/ability_mixin.rb
aviav: Yes need. With calls of the form `authorized?(:delete_own_user, ::User)` CanCan will ignore the block. See `can?(:read, Account)`

app/views/admin/users/me.html.erb
```diff
@@ -14,8 +14,5 @@
     <%= form.input :password_confirmation %>
   <% end %>
   <%= form.actions %>
-  <p>
-    <%= t('pageflow.admin.users.me.delete_account_hint') %>
-    <%= link_to(t('pageflow.admin.users.me.delete_account'), delete_me_admin_users_path, :class => 'delete_account') %>
-  </p>
+    <%= delete_own_user_section %>
```
tf: Indentation looks off here.

spec/features/editor/deleting_account_spec.rb
```diff
@@ -11,4 +11,23 @@
...
+          true
+        else
+          'Deletion is only possible for members of a dedicated account'
+        end
+      end
+
+    visit('/admin/users/me')
+
+    expect(page).to have_content(Dom::Admin::ProfileForm.first.cannot_delete_comment.text)
+  end
```
tf: I think this test would be more helpful (and faster) as a helper spec. We are also mostly covering happy paths in a lot of the other feature specs.

tf:
Starting to look really good. Please squash the commits into a single one when you are done.

app/views/admin/users/me.html.erb
aviav: Fixed

spec/features/editor/deleting_account_spec.rb
aviav: Agreed, moved non-happy path to helper spec

aviav:
Squashed all commits into one

spec/helpers/pageflow/admin/users_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+          expect(helper).to receive(:delete_me_admin_users_path).and_return('#')
+
+          section = helper.delete_own_user_section
+
+          expect(section).to have_selector('.may_delete')
+          expect(section).not_to have_selector('.cannot_delete')
+        end
+
+        it 'renders cannot_delete if authorize_user_deletion gives any string' do
+          Pageflow.config.authorize_user_deletion = lambda { |_user| 'true' }
```
tf: This is confusing. The quotes are hard to spot. Return a string like "Not allowed" and also query for it inside the resulting HTML.

tf: Maybe again use the user in the lambda to ensure it is passed.

spec/helpers/pageflow/admin/users_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Pageflow
+  module Admin
+    describe UsersHelper do
+      describe '#delete_own_user_section' do
+        it 'renders may_delete if authorize_user_deletion gives true' do
+          Pageflow.config.authorize_user_deletion = lambda { |_user| true }
+
+          expect(helper).to receive(:current_user).and_return(create(:user))
+          expect(helper).to receive(:delete_me_admin_users_path).and_return('#')
```
tf: Let's use `allow` here. Try prefixing the route helpers in the template with `main_app`. Then stubbing is not required. Compare app/views/components/pageflow/admin/grouped_folder_list.rb

app/views/pageflow/admin/users/_may_delete.html.erb
```diff
@@ -0,0 +1,6 @@
+<p class="may_delete">
+  <%= t('pageflow.admin.users.me.delete_account_hint') %>
+    <%= link_to(t('pageflow.admin.users.me.delete_account'),
+                delete_me_admin_users_path,
+                :class => 'delete_account') %>
```
tf: Where is Hound? ;)

aviav: Is this stylistically incorrect? I do not see anything wrong here

tf: Hash syntax?

aviav: Hm, where is Hound indeed? https://twitter.com/houndci does not report any malfunction

tf: I guess they do not lint templates?

aviav: In all fairness, they do not explicitly advertise checking embedded Ruby on their home page https://houndci.com/

spec/helpers/pageflow/admin/users_helper_spec.rb
aviav: Done

spec/helpers/pageflow/admin/users_helper_spec.rb
aviav: Done

PR #538, opened 2016-06-17
Improve meta fields

tilsammans:
This pull request adds default fields for author, publisher and keywords to Theming (Account).

### DIFF
--- a/app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
+++ b/app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
@@ -34,9 +34,16 @@ pageflow.EditMetaDataView = Backbone.Marionette.Layout.extend({
       });
 
       this.input('credits', pageflow.TextAreaInputView);
-      this.input('author', pageflow.TextInputView);
-      this.input('publisher', pageflow.TextInputView);
-      this.input('keywords', pageflow.TextInputView);
+
+      this.input('author', pageflow.TextInputView, {
+        placeholder: pageflow.config.defaultAuthorMetaTag
+      });
+      this.input('publisher', pageflow.TextInputView, {
+        placeholder: pageflow.config.defaultPublisherMetaTag
+      });
+      this.input('keywords', pageflow.TextInputView, {
+        placeholder: pageflow.config.defaultKeywordsMetaTag
+      });
     });
 
     configurationEditor.tab('widgets', function() {
--- a/app/models/pageflow/entry.rb
+++ b/app/models/pageflow/entry.rb
@@ -42,7 +42,7 @@ module Pageflow
     after_create unless: :skip_draft_creation do
       create_draft!(home_button_enabled: theming.home_button_enabled_by_default)
       draft.storylines.create!(configuration: {main: true})
-      theming.widgets.copy_all_to(draft)
+      theming.copy_defaults_to(draft)
     end
 
     def edit_lock
--- a/app/models/pageflow/theming.rb
+++ b/app/models/pageflow/theming.rb
@@ -26,5 +26,18 @@ module Pageflow
     def name
       I18n.t('pageflow.admin.themings.name', :account_name => account.name, :theme_name => theme_name)
     end
+
+    def copy_defaults_to(revision)
+      widgets.copy_all_to(revision)
+      copy_default_meta_tags(revision)
+    end
+
+    def copy_default_meta_tags(revision)
+      revision.update(
+        author: default_author.presence || Pageflow.config.default_author_meta_tag,
+        publisher: default_publisher.presence || Pageflow.config.default_publisher_meta_tag,
+        keywords: default_keywords.presence || Pageflow.config.default_keywords_meta_tag
+      )
+    end
   end
 end
--- a/db/migrate/20160617123557_add_meta_fields_to_theming.rb
+++ b/db/migrate/20160617123557_add_meta_fields_to_theming.rb
@@ -1,0 +1,7 @@
+class AddMetaFieldsToTheming < ActiveRecord::Migration
+  def change
+    add_column :pageflow_themings, :default_author, :string
+    add_column :pageflow_themings, :default_publisher, :string
+    add_column :pageflow_themings, :default_keywords, :string
+  end
+end
--- a/admins/pageflow/accounts.rb
+++ b/admins/pageflow/accounts.rb
@@ -89,7 +89,10 @@ module Pageflow
           :copyright_link_url,
           :copyright_link_label,
           :home_url,
-          :home_button_enabled_by_default
+          :home_button_enabled_by_default,
+          :default_author,
+          :default_publisher,
+          :default_keywords
         ] +
           permitted_attributes_for(:theming)
       end
--- a/app/views/admin/accounts/_form.html.erb
+++ b/app/views/admin/accounts/_form.html.erb
@@ -19,6 +19,18 @@
     <% end %>
 
     <%= f.inputs do %>
+      <%= theming.input :default_author,
+        hint: t('pageflow.admin.themings.default_author_hint'),
+        placeholder: Pageflow.config.default_author_meta_tag %>
+      <%= theming.input :default_publisher,
+        hint: t('pageflow.admin.themings.default_publisher_hint'),
+        placeholder: Pageflow.config.default_publisher_meta_tag %>
+      <%= theming.input :default_keywords,
+        hint: t('pageflow.admin.themings.default_keywords_hint'),
+        placeholder: Pageflow.config.default_keywords_meta_tag %>
+    <% end %>
+
+    <%= f.inputs do %>
       <%= theming.input :home_button_enabled_by_default %>
       <%= theming.input :imprint_link_label %>
       <%= theming.input :imprint_link_url %>
--- a/app/views/admin/accounts/_theming_details.html.arb
+++ b/app/views/admin/accounts/_theming_details.html.arb
@@ -3,4 +3,7 @@ attributes_table_for account.default_theming do
   row :theme, :class => 'theme' do
     account.default_theming.theme.name
   end
+  row :default_author, class: 'default_author'
+  row :default_publisher, class: 'default_publisher'
+  row :default_keywords, class: 'default_keywords'
 end
--- a/app/views/pageflow/config/_editor_seeds.json.jbuilder
+++ b/app/views/pageflow/config/_editor_seeds.json.jbuilder
@@ -1,3 +1,6 @@
 json.key_format!(camelize: :lower)
 json.(Pageflow.config, :confirm_encoding_jobs, :available_locales, :available_public_locales)
 json.file_types(Pageflow.config.file_types, :collection_name, :type_name, :param_key, :i18n_key)
+json.default_author_meta_tag Pageflow.config.default_author_meta_tag
+json.default_publisher_meta_tag Pageflow.config.default_publisher_meta_tag
+json.default_keywords_meta_tag Pageflow.config.default_keywords_meta_tag
--- a/config/locales/new/default-meta-on-theming.de.yml
+++ b/config/locales/new/default-meta-on-theming.de.yml
@@ -1,0 +1,13 @@
+de:
+  pageflow:
+    admin:
+      themings:
+        default_author_hint: Voreinstellung 'author' Meta-Feld für Eintrage
+        default_publisher_hint: Voreinstellung 'publisher' Meta-Feld für Eintrage
+        default_keywords_hint: Voreinstellung 'keywords' Meta-Feld für Eintrage
+    ui:
+      inline_help:
+        pageflow/entry:
+          author: Den Namen des Autors.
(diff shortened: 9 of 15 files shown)


app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
```diff
@@ -34,9 +34,16 @@ pageflow.EditMetaDataView = Backbone.Marionette.Layout.extend({
...
+
+      this.input('author', pageflow.TextInputView, {
+        placeholder: pageflow.theming.attributes.default_author
+      });
+      this.input('publisher', pageflow.TextInputView, {
+        placeholder: pageflow.theming.attributes.default_publisher
+      });
+      this.input('keywords', pageflow.TextInputView, {
+        placeholder: pageflow.theming.attributes.default_keywords
+      });
```
tf: Now that values are copied, displaying placeholders here might be misleading. Clearing the text box does not reset the values to the theming default.

db/migrate/20160617123557_add_meta_fields_to_theming.rb
```diff
@@ -0,0 +1,7 @@
+class AddMetaFieldsToTheming < ActiveRecord::Migration
+  def change
+    add_column :pageflow_themings, :default_author, :string
+    add_column :pageflow_themings, :default_publisher, :string
+    add_column :pageflow_themings, :default_keywords, :string
```
tf: I'm wondering if we should set explicit defaults from the configuration for existing themings. Then we could remove the fallback logic completely. Otherwise we would have to display placeholders in the theming form.

On the other hand when we remove fallback logic we have to also migrate existing revisions

app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
tilsammans: I agree. Drop them completely? Or replace with some generic placeholder?

```
Author
Publisher
comma, separated, keywords
```

db/migrate/20160617123557_add_meta_fields_to_theming.rb
tilsammans: I guess with the introduction of the fields in `Theming` we can think more strictly about responsibilities.

Are the defaults in the config file _only_ used to give a draft revision its initial values? Something can be said for it, but I don't look forward to migrating published entries. If you make the call though, I will. It means dropping the fallbacks from the [helper](https://github.com/scrollytelling/pageflow/blob/2fc46a31a20139071fa1fc8a25cd98f874516fac/app/helpers/pageflow/meta_tags_helper.rb#L5).

As for the [fallback in `theming.rb`](https://github.com/scrollytelling/pageflow/blob/2fc46a31a20139071fa1fc8a25cd98f874516fac/app/models/pageflow/theming.rb#L32) I am quite fond of it and would like to keep it there.

tf: As an alternative, we can keep the fallback in the helper and display the default values from the global config as placeholders both in the theming form and the editor. That way defaults from the configuration can be used without having to migrate existing records. Soon as a theming has default values, those will be copied to the draft revision of new entries. Maybe this is a good compromise: Changing defaults in the initializer takes effect for all entries which do not have custom values. Changes there require a deploy and will probably invalidate caches anyway. Default values from the theming are copied to keep things simple for account managers. I'd remove the fallback from the `copy_defaults_to` in `theming.rb`, though.

app/models/pageflow/entry.rb
```diff
@@ -43,6 +43,7 @@ class PasswordMissingError < StandardError
       create_draft!(home_button_enabled: theming.home_button_enabled_by_default)
       draft.storylines.create!(configuration: {main: true})
       theming.widgets.copy_all_to(draft)
+      theming.copy_defaults_to(draft)
```
tf: What do you think about also moving the `widgets.copy_all_to` call to the new `copy_defaults_to` method.

db/migrate/20160617123557_add_meta_fields_to_theming.rb
tilsammans: I don't think we're talking about the same thing? A placeholder is just a html5 attribute and does not get posted to the form. Are you talking about that? If you're not, I'm afraid I don't understand what you mean.

app/models/pageflow/entry.rb
tilsammans: My preference would be to separate them for now, since one is about defaults and one is about duplicating. But if this grows any bigger it will be ripe for optimization, sure. It doesn't bother me yet.

tf: Actually `widgets.copy_all_to` is also about defaults. It sets up the default widgets configured in the theming for a new entry.

db/migrate/20160617123557_add_meta_fields_to_theming.rb
tf: I'm also talking about HTML5 placeholders. My suggestion is to display the default values from the config as placeholders in the input fields, since - given the current fallback logic - these actually are the values which are used if the entry specific meta attributes are blank, right? Same thing in the theming form.

app/models/pageflow/entry.rb
tilsammans: @tf pushed up a couple commits to do this.

tilsammans:
@tf pushed up the required changes. Review?

app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
tf: Looking at [the meta tag helper](https://github.com/scrollytelling/pageflow/blob/8bec49b35e67fdb47a0326834e49fa46cee4d798/app/helpers/pageflow/meta_tags_helper.rb) it looks like we are actually falling back to global config, not the theming default when the user clears an input here. So I guess, we should be should be displaying that as a placeholder.

app/models/pageflow/theming.rb
```diff
@@ -26,5 +26,17 @@ def theme
...
+      widgets.all.each do |widget|
+        widget.copy_to(revision)
+      end
+
+      revision.update(
+        author: default_author.presence || Pageflow.config.default_author_meta_tag,
+        publisher: default_publisher.presence || Pageflow.config.default_publisher_meta_tag,
+        keywords: default_keywords.presence || Pageflow.config.default_keywords_meta_tag
+      )
+    end
```
tf: I think Hound is right here. Let's keep the `Widget.copy_all_to` and extract the revision update into `copy_default_meta_tags` method. That way `copy_defaults_to` has a nicer abstraction level.

tilsammans:
I can continue tomorrow at the earliest.

app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
tilsammans: I believe that's what my [latest change accomplished](https://github.com/scrollytelling/pageflow/blob/8bec49b35e67fdb47a0326834e49fa46cee4d798/app/views/pageflow/editor/themings/_theming.json.jbuilder#L8).

tilsammans:
`Widget.copy_all_to` is now retained, and rebased on `master`.

spec/models/pageflow/widget_spec.rb
```diff
@@ -6,18 +6,6 @@ module Pageflow
...
-        theming = create(:theming)
-        revision = create(:revision)
-        create(:widget, subject: theming, role: 'header', type_name: 'custom_header')
-
-        theming.widgets.copy_all_to(revision)
-
-        expect(revision.widgets).to include_record_with(type_name: 'custom_header', role: 'header')
-      end
-    end
-
```
tf: We lost a spec here.

app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
tf: But it would still display `theming.default_author` if that is present or am I missing something?

app/models/pageflow/entry.rb
```diff
@@ -43,6 +43,7 @@ class PasswordMissingError < StandardError
       create_draft!(home_button_enabled: theming.home_button_enabled_by_default)
       draft.storylines.create!(configuration: {main: true})
       theming.widgets.copy_all_to(draft)
+      theming.copy_default_meta_tags(draft)
```
tf: Ah ok, I actually liked the `copy_defaults_to` method you had before, just would have made these two calls the method body instead of inlining them. But this is really getting nitpicky.

tf:
Looks good! Can you address the last remaining points and squash the commits? Then we're good to merge.

tilsammans:
I will address them.

As far as squishing, this is now something built-in to GitHub. It works quite well! You can select the "squash commits" option on the merge button. I will update the PR probably later today.

spec/models/pageflow/widget_spec.rb
tilsammans: Oops, forgot to paste it back. 😟

tf:
Interesting. To be honest, I actually like the merge commits though since they state the PR number and branch name. Those would not show up if I hit the squash button, right? I just don't need the fine grain commit history that lead to the final state of the PR on master.

tilsammans:
> Those would not show up if I hit the squash button, right?

PR number does, in parentheses. And each commit message is added to a bullet list in the merge description. It's pretty nice. I'll squash-merge your PR for our navigation plugin so you can see if you like it. Otherwise happy to squash myself.

tilsammans:
[this](https://github.com/scrollytelling/scrollytelling-pageflow-navigation/commit/7ff8dcd62db1ced8f0e3316186e23a2e1bc638f6) is what a GitHub-squashed commit looks like. If there had been more than one commit, it would all have become just one.

tf:
I might try this on a new project, but for consistency I think I will stick with merging here. So it would be great if you could squash once your done. Thanks.

app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
tilsammans: I see.

Maybe I need to add the `theming` value to the meta tag helper, actually. So the logic in the helper would be entry, theming, configuration. Then we have consistency and a natural hierarchy in the helper. What do you think?

db/migrate/20160617123557_add_meta_fields_to_theming.rb
tilsammans: These [placeholders are now in place](https://github.com/scrollytelling/pageflow/commit/37d51c4d29b78ee6c65b0f66d5a96ab7c4c005d4).

app/models/pageflow/theming.rb
tilsammans: [Done](https://github.com/scrollytelling/pageflow/commit/33dc47fa011e330076f1ecaecf10123f0ab80b1b)

spec/models/pageflow/widget_spec.rb
tilsammans: This has been [added back in](https://github.com/scrollytelling/pageflow/commit/2bc90b326d24058e8ad74f9d6c708f3f25f3bdf3).

app/models/pageflow/entry.rb
tilsammans: Nitpicky is great! Fixed in 05feb74.

app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js
tf: This goes back to our discussion about copying vs "inheriting" defaults. If we extend the fallback logic of the helper, I'm not sure whether things might get more confusing since there then can be

- entries with copied default meta data which do not change when the defaults in the theming change
- entries with emptied meta data attributes which then fallback to theming defaults but also change when the theming changes

Being able to retroactively change the meta data of a lot of entries by editing the theming seems like to much power to me. I'd say let's just display the defaults form the global config as placeholders here. Then if somebody chooses to use those by emptying his meta data fields, at least there is less risk of unexpected change because changing these defaults at least means editing the initializer and deploying. Makes sense?

tilsammans: Right. I knew there was something.

Fixed in e9f6f66.

I will squash & rebase now.

tf: One last thing: Let's pass the global default values via `app/views/pageflow/config/_editor_seeds.json.jbuilder`

tf: Using the theming as a transport might trip us up later, since those or not really the theming attributes we are passing.

tf: Attributes rendered in the config seed are available as `pageflow.config...` inside the editor js.

tilsammans:
moved meta defaults to global JS config in 598e32a.

PR #554, opened 2016-06-22
Disable atmo on mobile platform

tf:
As long as there are still auto play bugs, WDR does not want to have
it play at all.

### DIFF
--- a/app/assets/javascripts/pageflow/slideshow/atmo.js
+++ b/app/assets/javascripts/pageflow/slideshow/atmo.js
@@ -61,7 +61,7 @@
     update: function() {
       var configuration = this.slideshow.currentPageConfiguration();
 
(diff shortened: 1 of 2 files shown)


tilsammans:
@tf could we make this configurable? We're sad to see this disabled across the board ... would like to keep it for us.

tf:
This was a request from one of our customers whose editors were confused by the fact that atmo sometimes did not work on mobile. There are still kinks with auto play restrictions to work out. We hope to improve in this area. So disabling the feature is sort of an interim solution. I'd be open to making it configurable, but I'm not sure how to reflect that in the help topic and inline help. As part of the change, we had to mention the restriction there. If there was a config option, than the help entry would either also have to be configurable in some way or be incorrect for one state of the option.

PR #557, opened 2016-06-22
Introduce widget margins

### LINKED ISSUE #413: Ability to completely hide navigation
For some stories or even accounts we would like the ability to totally hide the navigation menu.

Is that possible? In the theme we can make it invisible, but it would be even better if the HTML were not generated in the first place. Also, hiding the navigation still leaves room for it on video pages.

tf:
fixes #413 

Allow widgets to request a safe area along the page margin.

### DIFF
--- a/app/assets/stylesheets/pageflow/mixins/pageflow.css.scss
+++ b/app/assets/stylesheets/pageflow/mixins/pageflow.css.scss
@@ -37,3 +37,11 @@
     }
   }
 }
+
+@mixin pageflow-widget-margin($widget-name, $position) {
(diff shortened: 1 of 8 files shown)


tilsammans:
Nice. I have some reading to do with regards to new-style themes.

tf:
@tilsammans I've updated the [theme generator and corresponding guide](https://github.com/codevise/pageflow/blob/master/doc/creating_themes.md). Also, the [default theme's inline documentation](http://codevise.github.io/pageflow/theme/master/) is now generated in the Travis run. Everything is still a bit rough, but the basic setup is in place.

PR #564, opened 2016-06-27
Improve admin tab tables

tf:
Allow sorting and pagination for entries, member and user tables.


tf:
Duplicate of #563 since Hound appeared to be stuck.

PR #572, opened 2016-06-29
Improve display of 404 pages

### LINKED ISSUE #40: Move 404 Error Pages into View Path
The rendering of `public/pageflow/404.html` in the `format.any` block of the [`entries#show`](https://github.com/codevise/pageflow/blob/master/app/controllers/pageflow/entries_controller.rb#L35) action fails. Requesting `/foo.xml` results in a template missing.

tilsammans:
With this change, the EntriesController will now raise
`ActiveRecord::RecordNotFound` for all unknown formats.

This bubbles up to the ApplicationController where it is currently
dealth with in a `rescue_from` block. This has been changed to simply
render a 404 `head` response.

This frees us from trying to come up with appropriate response formats
while still giving the correct HTTP header.

I would strongly prefer to entirely ditch the rescuing from
ActiveRecord::RecordNotFound and let the host application deal with this
problem. It seems something that is very specific to the host, and not
up to Pageflow.

fixes #40


tilsammans:
@tf this is just an idea I had that will deal with this error gracefully, in my opinion. By raising this error we can let other layers deal with this problem. In this case that will be the application controller, but I would actually suggest to not rescue that and let the host application deal with it.

For example, I know that the regular 404 had hard-coded German in it, which is not something easily useful for English apps. I would say that raising is the most appropriate thing to do, even if we're not strictly trying to find a non-existing ActiveRecord.

Your ideas?

When pursued the `public` folder and associated stylesheets should be cleaned up too.

I never could find a reference to the `500.html`, it seems to be an orphan.

tf:
You make some compelling points and, as #40 states, things are currently a bit broken in this area.

First, I'd suggest removing the `format.any` from `EntriesController` completely since [we already are rescuing `UnknownFormat`](https://github.com/codevise/pageflow/blob/master/app/controllers/pageflow/application_controller.rb#L18) in the `ApplicationController`. We should probably improve our response there instead of rendering the text "Not found".

Regarding removing all rescuing logic I'm skeptical. Even if `Pageflow::ApplicationController` inherited from `ApplicationController`, forcing the host application to handle all cases there seems like leaking to much implementation detail to me. But what do you think about having the install generator create `public/pageflow/404.html` and rendering that? That way the host application has full control of what is rendered, but does not have to care about when this happens.

Keeping files in `public` also makes it easy to reuse files in the Nginx config. That's where the stray `500.html` comes from.

For more complex integration, I'd otherwise suggest doing this via new config options, i.e. allowing to register a lambda that handles the case where an entry is not found. That could then deal with specific needs like displaying a localized error page etc.

Thoughts?

tilsammans:
Tweaking the generator is an excellent idea! This is the least surprising alternative and quite flexible too.

I might be able to start work on this next week.

tilsammans:
@tf I pushed some new commits.

Once people upgrade to 0.11/0.12 we would have to urge them to re-run the generator.

lib/generators/pageflow/install/install_generator.rb
```diff
@@ -18,6 +18,7 @@ def generate_all
         invoke 'pageflow:user'
         invoke 'pageflow:seeds'
         invoke 'pageflow:active_admin_initializer'
+        invoke 'pageflow:not_found'
```
tf: I think `pageflow:error_pages` would be a bit more descriptive and would also allow to include files for other status codes should we ever need them.

lib/generators/pageflow/not_found/templates/404.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8"/>
   <title>Beitrag nicht gefunden (404)</title>
-  <link rel="stylesheet" type="text/css" href="error_assets/stylesheets/main.css" />
+  <link rel="stylesheet" type="text/css" href="/pageflow/stylesheets/main.css" />
```
tf: +1 for also using the `pageflow` namespace here. But `stylesheets/main.css` does not make is sound like these are only used for error pages. Maybe also the error pages themselves could live in `pageflow/error_pages/404.html` together with `pageflow/errors_pages/{stylesheets/main.css,images,fonts}`. That would make the scope clear and we'd still have the `public/pageflow` directory nice and clean if we ever need to add something else there.

tf:
Looks really good already! And for the upgrade step running the `pageflow:error_pages` generator should be enough, right?

tilsammans:
> And for the upgrade step running the pageflow:error_pages generator should be enough, right?

Yes.

Just pushed up the new commit to move error pages into its own directory. Final review?

app/controllers/pageflow/application_controller.rb
```diff
@@ -21,8 +21,14 @@ class ApplicationController < ActionController::Base
 
     rescue_from ActiveRecord::RecordNotFound do
       respond_to do |format|
-        format.html { render :file => Rails.root.join('public/404.html'), :status => :not_found }
-        format.any(:json, :css) { head :not_found }
+        format.html do
+          begin
+            render file: Rails.public_path.join('pageflow', 'error_pages', '404.html'), status: :not_found
```
tf: I'm still unsure if we should have a test for this? The line itself is covered. We could add an expection to the `EntriesController#show` spec that some text from the generated file shows up. For the rescue clause below, I see no way to get test coverage. A test could temporarily rename the `404.html` file in the dummy app. But I'm not sure it's worth it.

lib/generators/pageflow/error_pages/templates/404.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8"/>
   <title>Beitrag nicht gefunden (404)</title>
```
tf: The generated default file should probably be in english as that's also the default locale.

tf:
Added two more notes. Not sure about the test issue. Please again squash to one descriptive commit when you're done.

app/controllers/pageflow/application_controller.rb
tilsammans: I think it's worth it because the `head :not_found` in ApplicationController _is_ part of our code. I added 527f703 to test this. It might not look pretty but it is effective. I added the safeguard since the test could fail for other reasons and leave the dummy app in an unclean state.

An alternative could be to not bother with file-level stuff (which technically doesn't belong in a controller test anyway) but stub out the call to `render` and raise `ActionView::MissingTemplate` from the test. But then you're testing Rails framework detail which might change at some point...

Not sure what we should prefer here. But I do think a test has value.

tilsammans: As far as testing response content, that has little additional value to me over the status code. Also it makes the test brittle in case someone wants to put in new text in this file.

tf:
The test does not seem to work. I got suspicious when I saw that the `rescue` clause still has no test coverage. Running the test locally a few times reveals that rails caches rendered files. So if a spec before already rendered the `404.html`, renaming it does not hurt. I'm not sure what to do. Stubbing `render` also feels dirty. We could take this as an excuse to make the path to the error page configurable. That way, if the host app already has a `public/404.html`, Pageflow could be configured to reuse that. In our test though, we could point the config option to a non-existing path and check for the `head` response. What do you think?

Regarding testing the response body: The current test does not really ensure the template is used. For all we know this could also be the `head` response. Maybe we can at least test that the response is not empty or looks like an html page.

Finally, I'm not so fond of using `rand` in tests. If the id really does not matter, we can just as well choose one or use a string like `"not_there"`.

tilsammans:
@tf I took out the test and left it at that for the moment.

Adding the config feels a little scope-creepy to me, but not a bad idea in principle. If we're doing that only to satisfy the test case, that's probably not the right reason though.

I will be on vacation for three weeks, with little access to GitHub.

tf:
Agreed. Have a nice vacation.

Issue #539, opened 2016-06-21
Generate Procfile


tilsammans:
Would you be interested if I tweaked the generator, to generate a Procfile in the root?

It would look something like this:

```
web: bundle exec rails server
worker: bundle exec rake resque:work QUEUE=*
scheduler: bundle exec rake resque:scheduler QUEUE=*
```

We can get rid of the verbose instructions in the README then.

tf:
One would still need to mention foreman, right? Or would we also install that via the generator somehow?

tilsammans:
It would still have to be mentioned and installed manually, for local development. I'm cool with that.

tf:
Sounds fine to me.

PR #574, opened 2016-06-29
Generator unit tests

tilsammans:
In order to implement #539 I wanted to add unit tests for the generator. Even though the `bin/rspec` helper generates a dummy app and initializes Pageflow for it, I think testing the generators directly is valuable too.

Do you agree @tf? If so I will create more unit tests for the remaining generators first.

### DIFF
--- a/pageflow.gemspec
+++ b/pageflow.gemspec
@@ -161,4 +161,7 @@ Gem::Specification.new do |s|
 
   # Semantic versioning rake tasks
   s.add_development_dependency 'semmy', '~> 0.2.0'
+
+  # A gem that makes it easy to write specs for your Rails 3 Generators.
+  s.add_development_dependency 'ammeter', '~> 1.1'
 end
--- a/spec/generators/pageflow/active_admin_initializer/active_admin_initializer_spec.rb
+++ b/spec/generators/pageflow/active_admin_initializer/active_admin_initializer_spec.rb
@@ -1,0 +1,24 @@
+require 'spec_helper'
+require 'support/shared_contexts/generator'
+require 'fileutils'
+
+require 'generators/pageflow/active_admin_initializer/active_admin_initializer_generator'
+
+module Pageflow
+  module Generators
+    describe ActiveAdminInitializerGenerator, type: :generator do
+      before do
+        FileUtils.mkdir_p "#{destination_root}/config/initializers"
+
+        FileUtils.touch "#{destination_root}/config/initializers/active_admin.rb"
+      end
+
+      it 'adds Pageflow load path to active_admin load path' do
+        run_generator
+        initializer = file('config/initializers/active_admin.rb')
+        expect(initializer)
+          .to contain("ActiveAdmin.application.load_paths.unshift(Pageflow.active_admin_load_path)\n\n")
+      end
+    end
+  end
+end
--- a/spec/generators/pageflow/assets/assets_generator_spec.rb
+++ b/spec/generators/pageflow/assets/assets_generator_spec.rb
@@ -1,0 +1,45 @@
+require 'spec_helper'
+require 'support/shared_contexts/generator'
+require 'fileutils'
+
+require 'generators/pageflow/assets/assets_generator'
+
+module Pageflow
+  module Generators
+    describe AssetsGenerator, type: :generator do
+      before do
+        FileUtils.mkdir_p "#{destination_root}/app/assets/javascripts"
+        FileUtils.mkdir_p "#{destination_root}/app/assets/stylesheets"
+        FileUtils.mkdir_p "#{destination_root}/config"
+
+        FileUtils.touch "#{destination_root}/app/assets/javascripts/active_admin.js.coffee"
+        FileUtils.touch "#{destination_root}/app/assets/stylesheets/active_admin.scss"
+        FileUtils.touch "#{destination_root}/config/application.rb"
+      end
+
+      it 'generates editor assets' do
+        run_generator
+        expect(file('app/assets/javascripts/pageflow/editor.js')).to exist
+        expect(file('app/assets/stylesheets/pageflow/editor.scss')).to exist
+      end
+
+      it 'generates application assets' do
+        run_generator
+        expect(file('app/assets/javascripts/pageflow/application.js')).to exist
+        expect(file('app/assets/stylesheets/pageflow/application.scss')).to exist
+      end
+
+      it 'requires pageflow in active_admin JavaScript' do
+        run_generator
+        asset = file('app/assets/javascripts/active_admin.js.coffee')
+        expect(asset).to contain("#= require pageflow/admin\n")
+      end
+
+      it 'requires pageflow in active_admin SCSS' do
+        run_generator
+        asset = file('app/assets/stylesheets/active_admin.scss')
+        expect(asset).to contain("@import \"pageflow/admin\";\n")
+      end
+    end
+  end
+end
--- a/spec/generators/pageflow/cancan/cancan_generator_spec.rb
+++ b/spec/generators/pageflow/cancan/cancan_generator_spec.rb
@@ -1,0 +1,14 @@
+require 'spec_helper'
+require 'support/shared_contexts/generator'
+require 'generators/pageflow/cancan/cancan_generator'
+
+module Pageflow
+  module Generators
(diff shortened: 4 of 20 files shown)


tf:
Very interesting. I didn't know `ammeter` before. Yes, definitely go got it! I have some minor suggestions which I'll post as line comments.

pageflow.gemspec
```diff
@@ -161,4 +161,7 @@ Gem::Specification.new do |s|
 
   # Semantic versioning rake tasks
   s.add_development_dependency 'semmy', '~> 0.2.0'
+
+  # A gem that makes it easy to write specs for your Rails 3 Generators.
+  s.add_development_dependency 'ammeter', '~> 1.1.3'
```
tf: For versions `>= 1.0` I tend to only depend on the minor, i.e. `~> 1.1`, assuming Semver is being followed. Or is there a reason to stay on `1.1` stable?

spec/generators/pageflow/cancan/cancan_generator_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'spec_helper'
+
+# Generators are not automatically loaded by Rails
+require 'generators/pageflow/cancan/cancan_generator'
```
tf: Let's remove the repeated comment. The require pretty much speaks for itself.

spec/generators/pageflow/cancan/cancan_generator_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe Pageflow::Generators::CancanGenerator, type: :generator do
```
tf: Let's follow the convention found elsewhere in the codebase:

```
module Pageflow
  module Generators
    describe CancanGenerator, type: :generator do
```

spec/generators/pageflow/cancan/cancan_generator_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  # Tell the generator where to put its output (what it thinks of as Rails.root)
+  destination File.expand_path("../../../../../tmp", __FILE__)
+
+  before { prepare_destination }
```
tf: Can we move this and the `destination` call above to some shared context for `type: generator`. I don't think this setup code provides enough useful context to repeat it in all files.

spec/generators/pageflow/assets/assets_generator_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  module Generators
+    describe AssetsGenerator, type: :generator do
+      setup_default_destination
+
+      before do
+        FileUtils.mkdir_p "#{destination_root}/app/assets/javascripts"
+        FileUtils.mkdir_p "#{destination_root}/app/assets/stylesheets"
+        FileUtils.mkdir_p "#{destination_root}/config"
+
+        FileUtils.touch "#{destination_root}/app/assets/javascripts/active_admin.js.coffee"
```
tilsammans: @tf your thoughts about how deep these tests should go?

In this example I've just touched the file that is required by this generator. This is cool with me, as it allows us to run this test in the first place. I would be reluctant to actually re-create the entire contents of these files. It seems like that is very brittle.

So I propose that as far as required files by these generators, I just `touch` them, but never recreate the contents for them.

tilsammans:
This is as far as I think it valuable to go. Review?

spec/generators/pageflow/assets/assets_generator_spec.rb
tf: Sounds fine. This could only become a problem, when a generator tries to insert a line after an existing one. But I don't think we are doing this at the moment.

tilsammans: I broke my own promise with `user.rb` further down. :) But it seemed fair.

spec/support/generators.rb
```diff
@@ -0,0 +1,22 @@
...
+          # Tell the generator where to put its output (what it thinks of as Rails.root)
+          destination File.expand_path('../../tmp', __FILE__)
+
+          before { prepare_destination }
+        end
+      end
+
+      def self.included(klass)
+        klass.extend(Macros)
+      end
```
tf: What do you think of turning this into a shared context with metadata `type: generator`. That way we get rid of alle the `setup_default_destination` calls.

spec/support/generators.rb
```diff
@@ -0,0 +1,22 @@
+module Pageflow
+  module Specs
+    module Generators
+      module Macros
```
tf: Move this file to `spec/support/helpers` and it will get required automatically in the `spec_helper` no need to require it separately then.

spec/generators/pageflow/resque/resque_generator_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+require 'support/generators'
+
+# Generators are not automatically loaded by Rails
+require 'generators/pageflow/resque/resque_generator'
+
+module Pageflow
+  module Generators
+    describe Pageflow::Generators::ResqueGenerator, type: :generator do
```
tf: Duplicate module names. Also happened in some of the other files

spec/generators/pageflow/active_admin_initializer/active_admin_initializer_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+require 'support/generators'
+require 'fileutils'
+
+# Generators are not automatically loaded by Rails
```
tf: I still think these comments do not add much value.

spec/generators/pageflow/routes/routes_generator_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+require 'support/generators'
+
+# Generators are not automatically loaded by Rails
+require 'generators/pageflow/routes/routes_generator'
+
+module Pageflow
+  module Generators
+    describe RoutesGenerator, type: :generator do
+      # intentionally left empty since the routes generator depends on ActiveAdmin-generated routes.
```
tf: Why not also add a fixture file with the basic active admin routes? You also do this for the user model. Otherwise let's delete this file, since it does nothing.

spec/support/generators.rb
tilsammans: I believe that style has fallen out of fashion; it's too magical. These days the RSpec dogma is that explicit is better. But since the rest of the code is like that, I'll do it that way.

spec/support/generators.rb
tilsammans: The automatic loading of files is also something frowned upon these days ... makes everything slow. That's why I required it just in the tests.

tilsammans:
@tf all feedback implemented ✅

spec/support/generators.rb
tf: Good point. Probably something to consider moving forward. But for now I also prefer consistency.

spec/support/helpers/shared_contexts.rb
```diff
@@ -0,0 +1,4 @@
+RSpec.shared_context 'generators', type: :generator do
+  destination File.expand_path('../../../tmp', __FILE__)
+  before { prepare_destination }
+end
```
tf: At first I was thinking this file should be called `generator_shared_context.rb` since the current name suggests that all future shared contexts should live inside this file, which I'd like to avoid. But looking through the other helper files, I realized that this is actually the first shared context. So what do you think of moving the file to a new directory: `spec/support/shared_contexts/generator.rb`. We could then consider _not_ adding `spec/support/shared_contexts` to the list of magically required file and start following your suggestion to require support files explicitly. In separate PRs we could start doing the same for the other sub directories of `support`.

Sorry, for suggesting to undo my last change request, but I really think your argument for explicitness makes a lot of sense and a new support sub directory gives us a good opportunity to start moving in that direction without giving up the current structure of `spec/support`.

Other than that, everything looks really great! Please squash the commits and I'll merge.

tilsammans: No worries, it makes a lot of sense to me. Will try to wrap this up today.

tilsammans:
All done; squashed and rebased.

PR #578, opened 2016-07-13
remove unneeded loop

tilsammans:
It seems this can be optimised this way, or am I not understanding something?


tf:
I guess you are right. Still not completely sure if the order of CSS rules could be important here. I'll check  and report back.

tilsammans:
This one is from July and seems pretty small, chance it can go in before I need to rebase?

tf:
Yeah, sorry for the delay. Been meaning to test this and did not get around to it. CSS changes with possible consequences for rule precedence always make me scared. But this should be fine.

PR #588, opened 2016-08-16
Main color changes

kiddinho1:


tf:
* PR contains unrelated commits
* New main color makes for a pretty poor contrast especially for the links inside the credits menu and the active color of progress navigation bar button labels. Maybe Markus has another idea how this can be tweaked?

PR #599, opened 2016-08-17
prevent account with entries being deleted

### LINKED ISSUE #562: Account delete should deal with entries
When an account is deleted, something has to be done with its [entries](https://github.com/codevise/pageflow/blob/master/app/models/pageflow/account.rb#L6).

Either `nillify` or destroy, or the account destroy should be halted. Otherwise the admin pages can't deal with the missing account for the existing entry.

tilsammans:
Instead, ActiveRecord::DeleteRestrictionError will be raised. Entries
without accounts leave the system in an unstable state, so it must be
prevented.

Fixes #562

### DIFF
--- a/app/models/pageflow/account.rb
+++ b/app/models/pageflow/account.rb
@@ -2,8 +2,8 @@ module Pageflow
   class Account < ActiveRecord::Base
     include FeatureTarget
 
-    has_many :users
-    has_many :entries
+    has_many :users, dependent: :restrict_with_exception
+    has_many :entries, dependent: :restrict_with_exception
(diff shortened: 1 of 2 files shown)


app/models/pageflow/account.rb
```diff
@@ -3,7 +3,7 @@ class Account < ActiveRecord::Base
     include FeatureTarget
 
     has_many :users
```
tf: I think it's the same situation for users, right? Maybe we can handle that in one go?

tilsammans:
Agreed it should also look at users. PR updated.

PR #603, opened 2016-08-20
increase the test coverage

tilsammans:
* removed unused tests
* introduced test file for files_helper
* added tests for video_files_helper


tilsammans:
👍 

This was the low-hanging fruit. Eventually I will try to address all the missing coverage, but due to time constraints that won't be soon.

tf:
Thanks. While I don't think coverage should be a goal in itself as a quality measure, having 100% coverage is a great tool for dead code detection. Especially in `spec/support/dominos` there is some unused stuff for example. Just FYI, some of the front end related helpers might not be worth tackling since they will probably go away soon, when the built in page types will be ported to  use `pageflow-react`.

tilsammans:
The work on `pageflow-react` is partly why I didn't want to go too far with this. It should be a nice and performant update, looking forward to what will happen.

PR #615, opened 2016-08-31
Switch to cancancan from cancan

tf:
`cancan` is no longer maintained.


tilsammans:
CanCan is still mentioned in the README.

PR #616, opened 2016-08-31
Remove cargo-culted @body_class

tilsammans:
content_tag with a nested yield does not work completely, because the
end of the tag is swallowed unless you use `capture` inside the block.

This change replaces the content_tag helper with the regular HTML tag.
Now the </body></html> appears in the page.

### DIFF
--- a/app/views/layouts/pageflow/application.html.erb
+++ b/app/views/layouts/pageflow/application.html.erb
@@ -20,11 +20,12 @@
 
   <%= render 'layouts/pageflow/ie_include_tags' %>
 </head>
-<%= content_tag :body, :class => [@body_class, 'load_all_images has_no_mobile_platform non_js'].compact.flatten.join(' ') do %>
+
+<body class="load_all_images has_no_mobile_platform non_js">
   <script>
     jQuery('body').removeClass('load_all_images');
     jQuery("body").removeClass('non_js').addClass('js');
   </script>
   <%= yield %>
-<% end %>
+</body>
 <% end %>


tilsammans:
We were not seeing `</body></html>` in our productions, and using `yield` inside a block causes that. After this change the end tags are in the page again.

tilsammans:
Also I'm not sure if `@body_class` is actually used. It's not present in this repo, nor in my app itself. It might be used by a page type maybe? If that's the case, if could be worthwhile to make an official API out of it or mention it somewhere.

tf:
That's really weird. I checked and could not find any development or production system that does not render closing body or html tags. Can you tell me the configuration with which you are experiencing this problem?

I'm not a big fan of ERB interpolation inside of attribute values. It does not really look like idiomatic Rails to me.

Regarding `@bodyclass`, I also wasn't able to find any usage and wasn't even aware of its existence. So I guess it must have been cargo-culted in the early days and is safe to remove.

tilsammans:
Strange. All our stories have it, for example [this one](https://verhalen.volkskrant.nl/opluchting-in-manbij#3194). The `</body></html>` tags are missing. So I couldn't tell you the exact configuration. I do know that it's related to the yield inside the block.

I'll look into this further to see where the problem originates exactly.

tf:
Hm. Doing

    $ curl https://verhalen.volkskrant.nl/opluchting-in-manbij

prints the closing tags on my machine.

tilsammans:
It does appear I've been sleeping.

Well, at least we uncovered the `@body_class`. I'll amend this PR to nuke it.

tilsammans:
@tf dropped `@body_class`. That should be all for this PR.

app/views/layouts/pageflow/application.html.erb
```diff
@@ -20,7 +20,8 @@
 
   <%= render 'layouts/pageflow/ie_include_tags' %>
 </head>
-<%= content_tag :body, :class => [@body_class, 'load_all_images has_no_mobile_platform non_js'].compact.flatten.join(' ') do %>
+
+<%= content_tag :body, class: "load_all_images has_no_mobile_platform non_js" do %>
```
tf: Since there is no dynamic aspect left here, we can go all the way back to simply using `<body ...>`, right?

tilsammans: As far as I am concerned yes! You seemed to like the content_tag so I left it in. :)

Will replace with a regular body tag.

tf: I just don't fancy ERB interpolations inside of HTML attributes, but I sure like me some good old HTML ;)

tilsammans:
Done!

Issue #628, opened 2016-09-08
JSON.parse error on account admin page


tf:
When going to `/admin/account/n`, a [piece of JS code runs](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/admin/accounts.js#L4) that was originally designed to run only on the account edit form. But since the selector `.admin_accounts form.pageflow_account` also applies on the `show` page, the code runs and tries to parse JSON from `script#theme_options` which is not there. Trying to parse the empty string then results in the error.

The code needs to be changed to ensure that it only runs on `/admin/account/n/edit`.

silvestrijonathan:
Hi @tf!

I think I have a way to prevent this error from occurring, but I have a quick question regarding your request:

Do you want the variable to not even be set if a user is not on the `/admin/account/n/edit` page, or would you like for the `updateThemeFeatures` function on line 7 to not execute?

From your link in the issue it sounds like it's a case of the former, but I just wanted to check.

DanielKuhn:
Fixed in https://github.com/codevise/pageflow/pull/1072

PR #633, opened 2016-09-21
Add regex to check if current url is the admin accounts edit page.

silvestrijonathan:
Addressing #628.

themeOptions should not be set if the current url isn't the edit page.


silvestrijonathan:
The regex works as intended, but throws a new error as the variable is used in a function and will now no longer be defined. I think I can edit this to still work as intended, however.

silvestrijonathan:
Need to re-work my approach, will close this PR and re-open once I have something concrete.

PR #640, opened 2016-10-05
Have account filter respect permissions

aviav:
- Only users with account manager or admin permissions will see the
  account filter
- The account filter will only display accounts within the scope of the
  respective user

### DIFF
--- a/admins/pageflow/entry.rb
+++ b/admins/pageflow/entry.rb
@@ -49,7 +49,9 @@ module Pageflow
     end
 
     filter :title
-    filter :account
+    filter :account,
+           if: ->(_) { authorized?(:index, :accounts) },
+           collection: -> { Account.accessible_by(current_ability, :read) }
     filter :created_at
     filter :edited_at
     filter :first_published_at
--- a/spec/controllers/admin/entries_controller_spec.rb
+++ b/spec/controllers/admin/entries_controller_spec.rb
@@ -3,6 +3,31 @@ require 'spec_helper'
 describe Admin::EntriesController do
   render_views
 
+  describe '#index' do
+    describe 'account filter' do
+      it 'is displayed for account managers' do
+        account_manager = create(:user)
+        create(:account, with_manager: account_manager)
+
+        sign_in(account_manager)
+        get :index
+
+        expect(response.body).to have_selector('#q_account_id')
+      end
+
+      it 'is not displayed for account publishers and entry managers' do
+        underprivileged_user = create(:user)
+        account = create(:account, with_publisher: underprivileged_user)
+        create(:entry, with_manager: underprivileged_user, account: account)
+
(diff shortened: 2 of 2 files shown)


admins/pageflow/entry.rb
```diff
@@ -49,7 +49,9 @@ module Pageflow
     end
 
     filter :title
-    filter :account
+    filter :account,
+           if: ->(_) { authorized?(:index, :accounts) },
+           collection: -> { indexable_accounts }
```
tf: As long as we are depending on `cancancan` this seems like the perfect use case for `Account.accessible_by(current_ability)`. Should we ever switch to `pundit`, it would become `pundit_scope(Account)`. That way we can skip the helper. I'd also skip `indexable` since that basically is what `resolve` is used for in the context of Pundit/ActiveAdmin. The helper also does not use the alias at moment.

Since we are coding some logic in these ActiveAdmin DSL calls, I think it would have been justified to add a controller spec that checks the `select` tag inside the rendered view. That way we would be sure that nothing breaks during ActiveAdmin updates. But since I'm sure there are many cases where we haven't done it either, I'm also ok with leaving it untested.

admins/pageflow/entry.rb
```diff
@@ -49,7 +49,9 @@ module Pageflow
     end
 
     filter :title
-    filter :account
+    filter :account,
+           if: ->(_) { authorized?(:index, :accounts) },
+           collection: -> { Account.accessible_by(current_ability, :see_in_list) }
```
tf: Sorry for being so nitpicky, but in the `Folder` case in line 67 below we are also reusing `:read` which maps to `#resolve`. Why not do the same thing here? Also there is one Hound warning left.

aviav: While `can :read, Account` maps to `#resolve`, it also has a block. If we wanted to use it here, we could for that reason not use `accessible_by`. Additionally, `read?` in the account policy aliases `manage?`, which might require further changes or could at least lead to misunderstandings, since we do not want to show only accounts with at least `manage`-Permissions here.

The Hound warning should by now be obsolete, since `account` is actually used in the line below the one mentioned in the warning.

tf: I'm not aware of any restriction regarding `accessible_by` and actions that have a block. I think the block will simply be ignored and the scope returned by `AccountPolicy::Scope#resolve` will be used. The block is only used when the ability is inquired for individual records.

The fact that the block actually checks for manager permissions while the scope only looks for arbitrary membership, is in fact confusing, but is in the end caused by ActiveAdmin using `:read` to authorize `show` as well as `index`. But since the accounts displayed in the select should exactly be those displayed in the accounts index page, I think it's ok to mimic ActiveAdmin behavior here.

PR #642, opened 2016-10-06
extract h2 in various places into a partial

tilsammans:
We've been having some pain in styling the separate elements in this h2.

Ideally we would like to change the markup into:

```
h3.tagline
h1.title
h2.subtitle
```

which would be more semantically correct and easier to style as block elements.

Extracting the duplicated code into a partial would give us a way to easily override just this partial. But implementing more semantic markup instead of the one big `h2` would be cool with me too.

### DIFF
--- a/app/views/pageflow/pages/_page_headers.html.erb
+++ b/app/views/pageflow/pages/_page_headers.html.erb
@@ -1,0 +1,5 @@
+<h2>
+  <span class="tagline"><%= tagline %></span>
+  <span class="title"><%= title %></span>
+  <span class="subtitle"><%= subtitle %></span>
+</h2>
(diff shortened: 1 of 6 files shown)


tf:
Maybe to give some context, the admittedly questionable DOM structure resulted from a change request by one of our customers during an accessibility check in the early days of Pageflow. I am therefore a bit cautious of change in this area, especially since it would have to be performed in all if the page type gems to keep things consistent

Please also not, that the templates changed by this PR will go away soon when the default page types are rewritten as React components. Examples of the way things will look can already now be found in the [timeline page type](https://github.com/codevise/pageflow-timeline-page/blob/master/app/assets/javascripts/pageflow/timeline_page/components/page.jsx#L17). So if this should become an interface you plan to depend on, we would have to think about how to recreate it in the context of React components.

tf:
For the reasons outlines above I'm closing this PR.

PR #656, opened 2016-10-21
Use only one jshint config for Hound

tf:
There is no way to tell hound about the separate config in `spec`. Add
test lib globals to main jshint to prevent tons of warnings about
`describe` and `expect`.

[skip ci]


tf:
Travis does not start

PR #660, opened 2016-10-24
Nested file types, text track files

aviav:
Add support for nesting file types in parent file types.

As a first example, add text track files as a nested file type with parent video file. Includes menus and CRUD workflows, but not yet the actual display of subtitles etc.


app/models/concerns/pageflow/uploaded_file.rb
```diff
@@ -9,11 +9,34 @@ module UploadedFile
...
+        unless models_of_nested_file_types.include?(self.class)
+          errors[:base] << 'File type of provided parent file does not permit nesting files of '\
+                           "type #{self.class.name}"
+        end
+      end
+    end
+
+    def parent_belongs_to_same_entry
+      if parent_file.present?
+        unless parent_file.entry == entry
```
tf: Also there is no test for this validation yet.

app/controllers/pageflow/editor/files_controller.rb
```diff
@@ -25,6 +25,17 @@ def create
         respond_with(:editor, @file)
       end
 
+      def destroy
+        file = file_type.model.find(params[:id])
+
+        if file.parent_file_id?
+          authorize!(:destroy, file)
+          file.destroy
+        end
```
tf: Controller spec is still missing.

app/views/pageflow/editor/text_track_files/_text_track_file.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.(text_track_file, :configuration)
```
tf: Is this needed? The [configuration should be included by default](https://github.com/aviav/pageflow/blob/2525795da25baec58f70a6ffd25ff87a1920574d/app/views/pageflow/editor/files/_file.json.jbuilder#L1).

lib/pageflow/file_types.rb
```diff
@@ -7,13 +7,16 @@ def initialize(page_types)
     end
 
     def each(&block)
-      @page_types.map(&:file_types).flatten.uniq(&:model).each(&block)
+      first_level_file_types = @page_types.map(&:file_types).flatten
+      lower_level_file_types = search_for_nested_file_types(first_level_file_types)
+      (first_level_file_types + lower_level_file_types).uniq(&:model).each(&block)
     end
```
tf: So this is only tested indirectly, by making sure nested file types show up in the config seed, right? Might be worth a test of its own.

app/assets/javascripts/pageflow/editor/api/file_types.js
```diff
@@ -29,26 +29,42 @@ pageflow.FileTypes = pageflow.Object.extend({
...
+        _(this.clientSideConfigModifications[serverSideConfig.collectionName])
+          .each(function(modification) {
+            this.lintModification(modification, serverSideConfig.collectionName);
+            this.applyModification(clientSideConfig, modification);
+          }, this);
+
+        return new pageflow.FileType(_.extend({}, serverSideConfig, clientSideConfig));
+      }, this));
+
+    _.map(this.fileTypes.fileTypes, function(fileType) {
```
tf: Why `map`? `fileTypes` is enumerable itself. No need to reach into its internal state.

app/assets/javascripts/pageflow/editor/api/file_types.js
```diff
@@ -29,26 +29,42 @@ pageflow.FileTypes = pageflow.Object.extend({
...
+      if(fileType && fileType.nestedFileTypes.fileTypes){
+        fileType.nestedFileTypes = new pageflow.FileTypesCollection(
+          _.map(fileType.nestedFileTypes.fileTypes, function(nestedFileType) {
+            var clientSideConfig = clientSideConfigs[nestedFileType];
+            var serverSideConfig = _.find(serverSideConfigs, function(config) {
+              return config.collectionName === nestedFileType;
+            });
+            return new pageflow.FileType(_.extend({}, serverSideConfig, clientSideConfig));
```
tf: We already have created all of there file types above. Let's find them `this.filesTypes` and reuse instances. Otherwise, the modification features break for nested file types.

app/assets/javascripts/pageflow/editor/api/file_types.js
```diff
@@ -29,26 +29,42 @@ pageflow.FileTypes = pageflow.Object.extend({
...
+            this.applyModification(clientSideConfig, modification);
+          }, this);
+
+        return new pageflow.FileType(_.extend({}, serverSideConfig, clientSideConfig));
+      }, this));
+
+    _.map(this.fileTypes.fileTypes, function(fileType) {
+      if(fileType && fileType.nestedFileTypes.fileTypes){
+        fileType.nestedFileTypes = new pageflow.FileTypesCollection(
+          _.map(fileType.nestedFileTypes.fileTypes, function(nestedFileType) {
```
tf: This relies on parsing string representations of the nested files types given in the client side configuration into a `FileTypesCollection` which is then overwritten by a new `FileTypesCollection`. This change of types seems overly confusing to me. 

Moreover, I'd suggest including the names of the nested file types in the config seed data generated by the server. That way there is no need to specify the nested file types again when registering the parent file type in the JS code.

app/assets/javascripts/pageflow/editor/models/text_track_file.js
```diff
@@ -0,0 +1,23 @@
...
+    configuration: {
+      kind: 'captions'
+    }
+  },
+
+  initialize: function(attributes, options) {
+    pageflow.UploadedFile.prototype.initialize.apply(this, arguments);
+    if(this.isNew()) {
+      this.configuration.set('srclang', this.extractLanguageCodeFromFilename());
+      if(this.configuration.get('srclang' !== 'null')){
```
tf: Missing space after `if`.

app/assets/javascripts/pageflow/editor/models/text_track_file.js
```diff
@@ -0,0 +1,23 @@
...
+        this.configuration.set('label',
+                               I18n.t('pageflow.languages.' + this.configuration.get('srclang')));
```
tf: Storing translations in the database, breaks localization for multi-lang teams and prevents switching entry locale. We can use a `placeholder` in the form view and use a I18n generated label as a default value.

app/assets/javascripts/pageflow/ui/views/table_cells/text_table_cell_view.js
```diff
@@ -8,6 +8,13 @@ pageflow.TextTableCellView = pageflow.TableCellView.extend({
...
   update: function() {
-    this.$el.text(this.attributeValue());
+    if(this.attributeValue() === undefined ||
+       this.attributeValue() === null ||
+       this.attributeValue() === '') {
+      this.$el.text(this.options.column.default);
+    }
+    else {
+      this.$el.text(this.attributeValue());
+    }
```
tf: This is equivalent to `this.$el.text(this.attributeValue() || this.options.column.default);`. Please add a spec for this feature.

app/assets/javascripts/pageflow/editor/models/text_track_file.js
tf: No spec yet.

spec/javascripts/collections/files_collection_spec.js
```diff
@@ -36,6 +37,19 @@ describe('FileCollection', function() {
     });
   });
 
+  describe('.createForFileType', function() {
+    it('passes fileType to files', function() {
+      var fileType = new pageflow.FileType({collectionName: 'image_files',
+                                            model: Model,
+                                            matchUpload: /^image/});
```
tf: There is now a factory to create a fixture image file type. See other specs in this file.

app/assets/javascripts/pageflow/editor/initializers/setup_file_types.js
```diff
@@ -21,7 +32,44 @@ pageflow.app.addInitializer(function(options) {
...
+      },
+      {
+        name: 'kind',
+        inputView: pageflow.SelectInputView,
+        inputViewOptions: {
+          values: pageflow.config.availableTextTrackKinds,
+          texts: _.map(pageflow.config.availableTextTrackKinds, function(kind) {
+            var translation = 'pageflow.config.text_track_kind.' + kind;
+            return I18n.t(translation);
+          })
```
tf: We can use the [translationKeyPrefix](https://doclets.io/codevise/pageflow/master#dl-module-pageflow/ui-pageflow-SelectInputView) option here.

app/assets/javascripts/pageflow/editor/initializers/setup_file_types.js
```diff
@@ -21,7 +32,44 @@ pageflow.app.addInitializer(function(options) {
...
+      {
+        name: 'srclang',
+        inputView: pageflow.TextInputView
+      }
+    ],
+    nestedFileTableColumns: [
+      {
+        name: 'label',
+        cellView: pageflow.TextTableCellView,
+        default: I18n.t('pageflow.editor.templates.text_track_files.label.missing')
```
tf: The key prefix makes it look like it comes from a template.

app/assets/javascripts/pageflow/editor/views/nested_files_view.js
```diff
@@ -0,0 +1,81 @@
...
+
+    if(typeof(this.collection.at(fileIndex + 1)) !== 'undefined') {
+      this.options.selection.set('nextFile', this.collection.at(fileIndex + 1));
+    }
+    else if(typeof(this.collection.at(fileIndex - 1)) !== 'undefined') {
+      this.options.selection.set('nextFile', this.collection.at(fileIndex - 1));
+    }
+    else {
+      this.options.selection.set('nextFile', undefined);
+    }
```
tf: Space missing after `if`.

app/assets/javascripts/pageflow/ui/views/table_cells/icon_table_cell_view.js
```diff
@@ -0,0 +1,21 @@
...
+
+  removeExistingIcons: function() {
+    var iconClasses = pageflow.config.availableTextTrackKinds.join(' ');
+    this.$el.removeClass(iconClasses);
+  },
+
+  setTitle: function() {
+    return I18n.t('pageflow.config.text_track_kind.' + this.icon);
+  }
+});
```
tf: "Leaf views" like this a easy to test. Since this is a highly reusable view, I'd say it's worth it. You can look at the other table cells for examples.

app/assets/javascripts/pageflow/ui/views/table_cells/delete_row_table_cell_view.js
```diff
@@ -0,0 +1,28 @@
...
+  },
+
+  update: function() {
+    this.ui.removeButton.toggle(!this.isUploading());
+  },
+
+  destroy: function() {
+    this.getModel().destroy();
+  }
+});
```
tf: "Leaf views" like this a easy to test. Since this is a highly reusable view, I'd say it's worth it. You can look at the other table cells for examples.

app/assets/javascripts/pageflow/editor/views/nested_text_track_item_view.js
```diff
@@ -0,0 +1,20 @@
...
+  onRender: function() {
+    this.update();
+  },
+
+  update: function() {
+    this.ui.label.text(this.model.get('label'));
+    this.ui.language.text(this.model.get('srclang'));
+    this.ui.kind.text(this.model.get('kind'));
+  }
+});
```
tf: Unused?

app/assets/javascripts/pageflow/editor/views/nested_files_view.js
```diff
@@ -0,0 +1,81 @@
+pageflow.NestedFilesView = Backbone.Marionette.ItemView.extend({
```
tf: This could also have a domino based test, since it interacts with editor api. We could test that the specified columns actually appear, like we do in the `UploadableFilesView`.

app/assets/javascripts/pageflow/editor/models/file_uploader.js
```diff
@@ -15,13 +15,23 @@ pageflow.FileUploader = pageflow.Object.extend({
...
-      this.trigger('new:batch');
+    if (pageflow.editor.nextUploadTargetFile) {
+      deferred.resolve();
+    }
+    else {
+      this.deferreds.push(deferred);
+      if(this.deferreds.length == 1) {
+        this.trigger('new:batch');
+      }
     }
```
tf: This needs tests.

app/assets/javascripts/pageflow/editor/models/uploaded_file.js
```diff
@@ -10,6 +10,13 @@ pageflow.UploadedFile = Backbone.Model.extend({
 
     this.configuration.i18nKey = this.i18nKey;
 
+    if (pageflow.editor.nextUploadTargetFile){
+      this.set({parent_file_id: pageflow.editor.nextUploadTargetFile.get('id'),
+                parent_file_model_type: pageflow.editor.nextUploadTargetFile.fileType().typeName});
+      this.collection = pageflow.editor.nextUploadTargetFile
+        .nestedFiles(this.fileType().collectionName);
+    }
+
```
tf: This dependency on global editor state is ugly. This functionality should live inside of the `FilesCollection` which already knows the parent file and should set the collection property for new files all by itself.

app/assets/javascripts/pageflow/editor/templates/file_item.jst.ejs
```diff
@@ -16,6 +16,16 @@
...
+      <tbody>
+        <tr>
+          <th><%= I18n.t('pageflow.editor.templates.file_item.credits') %></th>
+          <td><input type="text" class="rights"/></td>
+        </tr>
+        <tr>
+          <th><%= I18n.t('pageflow.editor.templates.file_item.alt_text') %></th>
+          <td><input type="text" class="alt_text"/></td>
+        </tr>
+      </tbody>
```
tf: When you created the mock up, we decided to keep inputs out of the file info area. This can be added to the edit files view instead.

app/assets/javascripts/pageflow/editor/models/uploaded_file.js
aviav: How does a FilesCollection know the parent file other than via `pageflow.editor.nextUploadTargetFile`? We create any nestedFiles collection in this model and do not store direct information about the parent in such a collection

tf: That's true, but it should be easy to pass a reference to the parent file when creating the `nestedFiles` collection. [`createForFiletype`](https://github.com/aviav/pageflow/blob/5ff59e547c3d24cd6c54d682155a4bfa6ab0f27e/app/assets/javascripts/pageflow/editor/collections/files_collection.js#L77) passes all options to the collection.

db/migrate/20160824115000_add_parent_file_to_video_files.rb
```diff
@@ -0,0 +1,6 @@
+class AddParentFileToVideoFiles < ActiveRecord::Migration
+  def change
+    add_column :pageflow_video_files, :parent_file_id, :integer
+    add_column :pageflow_video_files, :parent_file_model_type, :string
+  end
```
aviav: This means we need to add the index to every file type table including panoramas, right?

app/assets/javascripts/pageflow/editor/templates/file_item.jst.ejs
tf: Ok, then let's remove all alt text related changes from this PR. Given the PR's title, they do no belong here anyway.

db/migrate/20160824115000_add_parent_file_to_video_files.rb
tf: Yes.

lib/pageflow/file_types.rb
aviav: I think in this case we already do what you asked for here: https://github.com/codevise/pageflow/pull/660/files#diff-f26ff2c97b6893fedefc1ce6225aabd3L17

app/assets/javascripts/pageflow/editor/api/file_types.js
aviav: The names of the nested file types are already included in the config seed data generated by the server, because in FileTypes#each, we recursively search for file types of nested files, starting at the top level file types.

This is what the editor receives:
```
"fileTypes":[{"collectionName":"image_files","typeName":"Pageflow::ImageFile","paramKey":"image_file","i18nKey":"pageflow/image_file"},{"collectionName":"video_files","typeName":"Pageflow::VideoFile","paramKey":"video_file","i18nKey":"pageflow/video_file"},{"collectionName":"audio_files","typeName":"Pageflow::AudioFile","paramKey":"audio_file","i18nKey":"pageflow/audio_file"},{"collectionName":"pageflow_panorama_packages","typeName":"Pageflow::Panorama::Package","paramKey":"package","i18nKey":"pageflow/panorama/package"},{"collectionName":"text_track_files","typeName":"Pageflow::TextTrackFile","paramKey":"text_track_file","i18nKey":"pageflow/text_track_file"}],
```

However, this data is not structured according to the nesting hierarchy. Maybe we can make this easier to read by also passing the editor this kind of nesting information, or by extracting the nested file types in a way that is more similar to top level file type extraction.

app/assets/javascripts/pageflow/ui/views/mixins/input_view.js
```diff
@@ -147,8 +157,8 @@ pageflow.inputView = {
 
     this.$el.data('inputPropertyName', this.options.propertyName);
 
-    this.ui.labelText.text(this.labelText());
-    this.ui.inlineHelp.text(this.inlineHelpText());
+    this.ui.labelText.html(this.labelText());
```
tf: `labelText` returns text. No need for HTML in label, I think.

app/views/pageflow/editor/files/_file.json.jbuilder
```diff
@@ -18,3 +25,14 @@ if file.respond_to?(:thumbnail_url)
...
+
+file_type.nested_file_types.each do |nested_file_type|
+  json.set! 'nested_files' do
+    json.set! nested_file_type.collection_name.pluralize do
+      json.partial!(partial: 'pageflow/editor/files/file',
+                    collection: file.nested_files(nested_file_type.model),
+                    locals: {file_type: nested_file_type},
+                    as: :file)
+    end
+  end
```
tf: I think this is not needed anymore.

app/assets/javascripts/pageflow/editor/models/file_uploader.js
```diff
@@ -15,13 +17,23 @@ pageflow.FileUploader = pageflow.Object.extend({
       fileType: fileType
     });
 
+    if (editor.nextUploadTargetFile){
+      file.set({parent_file_id: editor.nextUploadTargetFile.get('id'),
+                parent_file_model_type: editor.nextUploadTargetFile.fileType().typeName});
+    }
+
```
tf: What happens when a top level file upload is started (via drag and drop) while a `nextUploadTargetFile` is set?

config/locales/new/video_subtitles.de.yml
```diff
@@ -0,0 +1,95 @@
...
+  pageflow:
+    config:
+      text_track_kind:
+        subtitles: Übersetzungs-Textspur
+        captions: Ton-Ersatz-Textspur
+        descriptions: Bild-Ersatz-Textspur
+    editor:
+      errors:
+        nested_type_error: Der Typ '%{type}' des Uploads '%{name}' ist nur für genestete Dateien erlaubt.
+        top_level_type_error: Der Typ '%{type}' des Uploads '%{name}' ist nur für nicht-genestete Dateien erlaubt.
```
tf: These error messages are too technical IMO. We cannot expect the user to understand what a nested file is.

app/assets/javascripts/pageflow/editor/models/file_uploader.js
```diff
@@ -15,13 +17,31 @@ pageflow.FileUploader = pageflow.Object.extend({
...
+      if (fileType.topLevelType) {
+        throw(new pageflow.TopLevelTypeError(upload));
+      }
+      file.set({parent_file_id: editor.nextUploadTargetFile.get('id'),
+                parent_file_model_type: editor.nextUploadTargetFile.fileType().typeName});
+    }
+    else if (!fileType.topLevelType) {
+      throw(new pageflow.NestedTypeError(upload));
+    }
+
```
tf: This still does not check whether the the uploaded file is allowed as a nested file of the current targe.

aviav: That is true. However, we have a validation in the Rails backend for that https://github.com/codevise/pageflow/pull/660/files#diff-647fc1f26572e8132240adef057a5ebcR22

spec/javascripts/models/file_uploader_spec.js
```diff
@@ -1,103 +1,209 @@
...
+        editor.setUploadTargetFile(targetFile);
+        fileUploader.add(nestedFileUpload, {editor: editor});
+        sinon.stub(this.entry, 'getFileCollection').withArgs(this.textTrackFileType)
+          .returns(new pageflow.FilesCollection.createForFileType(this.textTrackFileType,
+                                                                  {},
+                                                                  {entry: this.entry,
+                                                                   parentFile: targetFile}));
+        expect(editor.nextUploadTargetFile.nestedFiles(
+          this.entry.getFileCollection(this.textTrackFileType)
+        ).length).to.eq(1);
```
tf: What does this stub do? The only one calling `this.entry.getFileCollection` after it has been stubbed, is the test itself, to get the parameter for the `nestedFiles` call. If we want to pass in a newly created collection, we can just do that directly. Moreover, far as I can see, the collection returned by the stub is empty (we pass `{}` as `files` paramter, which I guess should be `[]`). So I'm not sure why the test is passing.

PR #665, opened 2016-10-28
Embed entries as iframes

### LINKED ISSUE #622: Add route to allow displaying entry inside iframe
At the moment entries cannot be embedded on another website via an iframe since Rails sets an `X-Frame-Options: SAMEORIGIN` header by default. It would be nice if for a published entry with route `/my-entry` there was an additional route `/my-entry/embed` that is handled by the same controller action, but does not include the `X-Frame-Options` header.

To achieve this, we can add an `after_action` for  [the `show` action of `Pageflow::EntriesController`](https://github.com/codevise/pageflow/blob/master/app/controllers/pageflow/entries_controller.rb#L23) which removes the header when an `embed` param is present.

Then we can [add a route](https://github.com/codevise/pageflow/blob/master/config/routes.rb#L66) for `/:id/embed` that also goes to `entries#show` and sets the `embed` param [via a default](http://guides.rubyonrails.org/routing.html#defining-defaults).

We should also add [a controller spec](https://github.com/codevise/pageflow/blob/master/spec/controllers/pageflow/entries_controller_spec.rb#L151) to test that removing the header by adding the `embed` param works.

tf:
* Add embed route to display entry inside iframe
* Add embed code field to entry admin attributes table

fixes #622

### DIFF
--- a/admins/pageflow/entry.rb
+++ b/admins/pageflow/entry.rb
@@ -155,6 +155,7 @@ module Pageflow
     controller do
       helper FoldersHelper
       helper EntriesHelper
+      helper EmbedCodeHelper
       helper Admin::EntriesHelper
       helper Admin::FeaturesHelper
       helper Admin::FormHelper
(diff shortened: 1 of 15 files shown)


tf:
Did not know about this option. Might be something to consider in the future.

PR #670, opened 2016-11-07
move disabled into the initializer

tilsammans:
This way, sites have the option to force this value after the Pageflow
javascript-code is done initialising.

See #554 for previous discussion.

### DIFF
--- a/app/assets/javascripts/pageflow/slideshow/atmo.js
+++ b/app/assets/javascripts/pageflow/slideshow/atmo.js
@@ -5,7 +5,7 @@
     initialize: function(slideshow, events, multiPlayer) {
       this.slideshow = slideshow;
       this.multiPlayer = multiPlayer;
-      this.disabled = false;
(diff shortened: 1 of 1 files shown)


tilsammans:
@tf what do you think of this change?

It would allow us to enable the value in the application.js, thereby ignoring the default.

tf:
I guess that's fine. It should then be safe to call `pageflow.atmo.enable()` when the `pageflow.ready` promise resolves.

tf:
The help texts will still claim that atmo is disabled on mobile, though. But the medium term goal has to be making atmo work on mobile anyway since technically it should be possible by reusing audio elements.

tilsammans:
Just to be sure, that would be this code, right?

```javascript
pageflow.ready.then(function(result) {
  pageflow.atmo.enable();
});
```

tf:
I think so, yes.

PR #671, opened 2016-11-09
Improve HTML translation support in I18nUtils

aviav:
- add html option to findTranslation, which will expand searches to also
  look for keys with _html suffix
- HTML-escape translations whose keys do not end in _html

### DIFF
--- a/app/assets/javascripts/pageflow/ui/utils/i18n_utils.js
+++ b/app/assets/javascripts/pageflow/ui/utils/i18n_utils.js
@@ -12,15 +12,30 @@ pageflow.i18nUtils = {
    *   Translation key candidates.
    *
    * @param {string} [options.defaultValue]
-   *   Value to return if non of the keys has a translation.
+   *   Value to return if none of the keys has a translation. Is
+   *   treated like an HTML translation if html flag is set.
+   *
+   * @param {boolean} [options.html]
+   *   If true, also search for keys ending in '_html' and HTML-escape
+   *   keys that do not end in 'html'
    *
    * @return {string}
    */
   findTranslation: function(keys, options) {
     options = options || {};
 
+    if (options.html) {
+      keys = this.translationKeysWithSuffix(keys, 'html');
+    }
+
     return _.chain(keys).reverse().reduce(function(result, key) {
-      return I18n.t(key, _.extend({}, options, {defaultValue: result}));
+      var unescapedTranslation = I18n.t(key, _.extend({}, options, {defaultValue: result}));
+      if (!options.html || key.match(/_html$/) || result == unescapedTranslation) {
+        return unescapedTranslation;
+      }
+      else {
+        return $('<div />').text(unescapedTranslation).html();
+      }
     }, options.defaultValue).value();
   },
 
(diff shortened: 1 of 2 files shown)


spec/javascripts/pageflow/ui/utils/i18n_utils_spec.js
```diff
@@ -31,6 +33,33 @@ describe('pageflow.i18nUtils', function() {
 
       expect(result).to.eq('Value interpolated');
     });
+
+    it('escapes html', function() {
+      var result = pageflow.i18nUtils.findTranslation(
+        ['html.key.without.suffix']
+      );
+
+      expect(result).to.eq('&lt;div /&gt;');
```
tf: This is an unwanted breaking change imo. Escaping should only take place, if the caller signals via the `html` option that he intents to treat the result as HTML. If we expect text, we never want to see HTML entities in the result.

If I read the implementation correctly the default value is also not escaped, which I think is correct behavior. We should also have tests for the treatment of the default value.

app/assets/stylesheets/pageflow/themes/default/logo/variant/watermark.scss
```diff
@@ -5,6 +5,11 @@
 /// - `"all-pages"`: on all pages
 $logo-watermark-variant-fade-in-near-top: "first-page" !default;
 
+/// Fade in logo when header is active. If the header logo is
+/// inverted on the current page, the non-inverted logo variant
+/// is displayed while the header is active.
+$logo-watermark-variant-fade-in-with-header: false !default;
+
```
tf: The branch includes a rebased version of a commit already on master.

app/assets/javascripts/pageflow/ui/utils/i18n_utils.js
```diff
@@ -6,21 +6,34 @@
...
     options = options || {};
 
+    if (options.html) {
+      keys = this.translationKeysWithSuffix(keys, 'html');
+    }
+
     return _.chain(keys).reverse().reduce(function(result, key) {
-      return I18n.t(key, _.extend({}, options, {defaultValue: result}));
+      var unescapedTranslation = I18n.t(key, _.extend({}, options, {defaultValue: result}));
+      if (key.substr(-5) === '_html' || result == unescapedTranslation) {
```
tf: `match(/_html$/)` would have been an alternative.

spec/javascripts/pageflow/ui/utils/i18n_utils_spec.js
aviav: I think this is the behavior that we would expect from Rails translations http://guides.rubyonrails.org/i18n.html#using-safe-html-translations

I agree on testing default escaping.

app/assets/javascripts/pageflow/ui/utils/i18n_utils.js
```diff
@@ -6,21 +6,35 @@
 pageflow.i18nUtils = {
   /**
    * Find the first key for which a translation exists and return the
-   * translation.
+   * translation. Escape HTML unless key ends in '_html'.
```
tf: This is no longer true, is it?

PR #679, opened 2016-12-13
Render storylines as array in seed data

tf:
Unify interface to allow generic handling of storylines just like any
other collection.


tf:
Too many Hound warnings.

PR #692, opened 2017-01-06
backport e2f36f4433a1df8ced235b2bf0db1381fac312f9

tilsammans:
This way, sites have the option to force this value after the Pageflow
javascript-code is done initialising.


tf:
Thanks, but I do not plan to release another patch level version before the next minor comes out. So I'll say just use the fork for now.

PR #694, opened 2017-02-03
Widget configuration

tf:
Add serialized configuration hash to widget model and allow editing inside the editor.


tilsammans:
This one is very welcome!! Always handy to inspect the entry when a WidgetType doesn't do what I want it to. 👍

PR #737, opened 2017-04-19
Add theme options for first page top padding

tf:
Give more fine grain control of top padding on first page to allow
making space for logo.


tf:
Duplicate.

PR #738, opened 2017-04-19
Add theme options for dimension of background logo

tf:
Allow specifying custom mobile height and top position on phone.


tf:
Duplicate.

PR #753, opened 2017-05-11
add Cache-Control header for S3 uploads

tilsammans:
In addition to the existing Expires header _for now_. I am actually proposing we get rid of the Expires header completely. I'll write this in a comment on the pull request.

Expires is HTTP/1.0 (!)
Cache-Control is HTTP/1.1 and supersedes Expires.
Cache-Control offers everything from Expires and a lot more.

In our case, it's important to mark the files public, so that they will
be cached by proxies. Also give them the same high expiry time (1 year), because they
have timestamps in the URL, which invalidates the cache.

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -6,31 +6,31 @@ Pageflow.configure do |config|
 
   if Rails.env.test?
     config.paperclip_s3_default_options.merge!({
-      :storage => :filesystem,
-      :path => ':rails_root/tmp/attachments/test/s3/:class/:attachment/:id_partition/:style/:filename'
+      storage: :filesystem,
+      path: ':rails_root/tmp/attachments/test/s3/:class/:attachment/:id_partition/:style/:filename'
     })
   else
     config.paperclip_s3_default_options.merge!({
-      :storage => :s3,
-      :s3_headers => {'Expires' => 1.year.from_now.httpdate},
-      :s3_options => {:max_retries => 10},
+      storage: :s3,
+      s3_headers: {'Cache-Control' => "public, max-age=31536000"},
+      s3_options: {max_retries: 10},
 
-      :url => ':s3_alias_url',
-      :path => ':host/:class_basename/:attachment/:id_partition/:pageflow_attachments_version:style/:filename',
+      url: ':s3_alias_url',
+      path: ':host/:class_basename/:attachment/:id_partition/:pageflow_attachments_version:style/:filename',
 
       # Paperclip deletes and reuploads the original on
       # reprocess. Sometimes S3 seems to change the order of commands
       # causing the image to be deleted. This is fixed on paperclip
       # master, but for us not deleting old files is good enough. They
       # might be in the CDN anyway.
-      :keep_old_files => true
+      keep_old_files: true
     })
   end
 
   config.paperclip_filesystem_default_options.merge!({
-    :storage => :filesystem,
-    :path => ':pageflow_filesystem_root/:class/:attachment/:id_partition/:style/:filename',
-    :url => 'not_uploaded_yet'
+    storage: :filesystem,
+    path: ':pageflow_filesystem_root/:class/:attachment/:id_partition/:style/:filename',
+    url: 'not_uploaded_yet'
   })
 
   config.thumbnail_styles = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,19 @@
 
 [Compare changes](https://github.com/codevise/pageflow/compare/12-0-stable...master)
 
-None so far.
+Media stack:
+
+- Switch from `Expires` to `Cache-Control` header for media uploads.
+  ([#753](https://github.com/codevise/pageflow/pull/753))
+
+It’s recommended you update the files currently stored on S3:
+
+```
+s3cmd --recursive modify --add-header="Cache-Control: public, max-age=31536000" s3://yourbucket/
+s3cmd --recursive modify --remove-header=Expires s3://yourbucket/
+```
+
+Tread carefully when you do this! As noted in [this StackExchange answer](http://stackoverflow.com/questions/22501465/how-to-add-cache-control-in-aws-s3), we have experienced that some public read permissions were lost after running this script. Test first using just a single object. In the AWS Management Console, you might want to grant public read access on the entire bucket again to be safe.
 
 See
 [12-0-stable](https://github.com/codevise/pageflow/blob/12-0-stable/CHANGELOG.md)


tilsammans:
I wanted to get consensus about the Expires header. My vote would be to remove it. It is the least confusing. Cache-Control works everywhere, there is really no reason to keep using Expires.

This is made worse by the fact ...

The `1.year.from_now` is calculated at *parse* time (app boot) and after that never changes! I have soooo many files in S3 with a Expires time in the past! Thus ensuring they will never be cached, which is totally not what we want.

Cache-Control works by calculating seconds from the http request, so the header field value can remain constant. If we really need the Expires in there, we have to put this line in a lambda.

Changing headers on your S3 files works easiest using s3cmd. I will paste the command lines to use in a new comment.

tf:
I agree. This nonsense must have been cargo-culted from somewhere. I guess removing and replacing it with Cache-Control is the best option.

tilsammans:
Okay, I will update this pull request accordingly.

tilsammans:
The command to update your existing S3 uploads:

```
s3cmd --recursive modify --add-header="Cache-Control: max-age= 31536000" s3://yourbucket/
```

Then get rid of all the existing Expires headers, since there will be a lot in the past:

```
s3cmd --recursive modify --remove-header=Expires s3://yourbucket/
```

config/initializers/paperclip.rb
```diff
@@ -6,31 +6,31 @@
 
   if Rails.env.test?
     config.paperclip_s3_default_options.merge!({
-      :storage => :filesystem,
-      :path => ':rails_root/tmp/attachments/test/s3/:class/:attachment/:id_partition/:style/:filename'
+      storage: :filesystem,
```
tilsammans: I don't understand this one. Looks fine to me.

config/initializers/paperclip.rb
```diff
@@ -6,31 +6,31 @@
...
-      :s3_headers => {'Expires' => 1.year.from_now.httpdate},
-      :s3_options => {:max_retries => 10},
+      storage: :s3,
+      s3_headers: {'Cache-Control' => "public, max-age=31536000"}, # Cache-Control preferred over Expires
+      s3_options: {max_retries: 10},
 
-      :url => ':s3_alias_url',
-      :path => ':host/:class_basename/:attachment/:id_partition/:pageflow_attachments_version:style/:filename',
+      url: ':s3_alias_url',
+      path: ':host/:class_basename/:attachment/:id_partition/:pageflow_attachments_version:style/:filename',
```
tilsammans: No quick wins here.

tilsammans:
It's all done @tf 

I can squash if you want.

tf:
Looks good. The remaining Hound indentation warnings are invalid IMO. Not sure the comment  "Cache-Control preferred over Expires" is needed, though. If you just read the code the way it is now, you never know there was an obsolete Expires header there before.

Otherwise, just squash the commits and I'll merge. I'll probably backport the change to `12-stable`. Makes sense to have it in the 12.0 release.

config/initializers/paperclip.rb
tf: I think Hound wants us to indent all the way beyond the parenthesis. If this should come up more often, we have to tweak the Rubocop config.

tilsammans:
I see. Relative to the parenthesis. That looks not right to me.

tilsammans:
Squashed and rebased @tf

tilsammans:
Just now read your comment. I'll amend and push again today.

I know you're right but I felt uneasy ripping out something that has had direct influence on people's S3 since version 0.0.0. But now I updated the CHANGELOG so the record is there.

tf:
Two questions regarding the `s3cmd` call:

1. In the default options you include `public` in the `Cache-Conrol` header which makes a lot of sense to me judging from the MDN docs on `Cache-Control`. Any reason this is not set for existing files on S3?

2. In my tests the `--acl-public` flag was no op in combination with `modify`. Only when used with the `setacl` command I saw an effect. But also only if I revoked public read access manually first since all files were marked with "Everyone: Read" by Paperclip anyway. Anything I am missing here?

tilsammans:
> Any reason this is not set for existing files on S3?

No good reason; the command line was made yesterday by me when I didn't optimize the settings yet. I agree that is should be included. I'll update the pull request.

I chose to have two commands because running it takes a long time; and I don't think we can guarantee that a header is added and functioning, before we remove another, all in the same command. I am also not aware of a better option but if there is, I would be happy to include that.

> Anything I am missing here?

I ran the script yesterday and the public read access for everyone was removed. This is absolutely true. I cannot be positive that it was the operation itself or just some side-effect. Also don't think it affected all files. But it did affect many and we were getting permission denied.

I agree with your assessment on the docs, but then again [someone warned about this in the StackExchange answer](http://stackoverflow.com/questions/22501465/how-to-add-cache-control-in-aws-s3). I am not sure what is the best approach. Maybe drop the command line option but add a warning that people should just bulk-update the entire bucket to be sure? After I did that, all permissions were fixed.

tf:
That sounds scary. In the StackExchange answer there are also reports of lost `Content-Type` headers when using `s3cmd` 1.5.0-rc1. That would be even worse, because information might be lost. What version of `s3cmd` are you using?

tilsammans:
I am using master, the version number is 1.6.1.

Our Content-Types were unchanged.

tf:
Not sure. I never used neither of those tools before. Still I found [an issue discussion](https://github.com/s3tools/s3cmd/issues/37#issuecomment-65021193) about the content type bug and it appears to have been fixed some time ago.

I'd imagine `cp` to be a lot slower. So maybe, as you said, just including a word of warning to use the newest `s3cmd` version might be ok.

tf:
And [here the author actually says](https://github.com/s3tools/s3cmd/pull/406#issuecomment-65011029) that the `--acl-public` option might actually be required (or has been at some point).

tf:
And in the PR request referenced in the very next comment [there is a commit](https://github.com/s3tools/s3cmd/pull/428/commits/da091eee9876632d9581e79ce0718a9bdba17d76) that says the option is now no longer required. Really strange that you experienced the problem still, though.

tilsammans:
Okay, I am printing a warning and I am not changing it to AWS CLI. The docs say that when using the REPLACE directive, all meta data from the original object is discarded. Which would obviously be a disaster. At least I know that those two commands work.

tilsammans:
Good to go @tf

tf:
Backported to `12-0-stable`.

tilsammans:
I am running the commands now. So far, any file that's been done still has the public read permissions. So it's definitely not affecting everything. The command takes hours to run for us. I will do more checks and report back if there are any causes I can identify.

PR #763, opened 2017-05-18
Fix page title hysteresis

aviav:
On choosing a page link, properly update title when it should be an
empty string. Without this, e.g. in ReferenceInputView, `title` will
not update in certain situations, because page#title() returns
`'undefined'` due to `''` falsiness in Javascript.

To reproduce:
```
this.configuration.get('title') === ''
this.configuration.get('additional_title') === undefined
```

### DIFF
--- a/app/assets/javascripts/pageflow/editor/models/page.js
+++ b/app/assets/javascripts/pageflow/editor/models/page.js
@@ -48,7 +48,9 @@ pageflow.Page = Backbone.Model.extend({
   },
 
   title: function() {
-    return this.configuration.get('title') || this.configuration.get('additional_title');
+    return this.configuration.get('title') ||
+      this.configuration.get('additional_title') ||
+      '';
   },
 
   thumbnailFile: function() {


app/assets/javascripts/pageflow/editor/models/page.js
```diff
@@ -48,7 +48,14 @@ pageflow.Page = Backbone.Model.extend({
...
   title: function() {
-    return this.configuration.get('title') || this.configuration.get('additional_title');
+    var configurationTitle = this.configuration.get('title');
+    var title = configurationTitle || this.configuration.get('additional_title');
+
+    if(!title && configurationTitle === '') {
+      title = configurationTitle;
+    }
+
+    return title;
```
tf: So we want to ensure that `title` always returns a string and never `undefined`. Wouldn't it be easier to say:

```
return this.configuration.get('title') || this.configuration.get('additional_title') || '';
```

aviav: Agreed

PR #765, opened 2017-05-19
Themes as features

aviav:
Integrate themes with feature flags

### DIFF
--- a/app/models/pageflow/revision.rb
+++ b/app/models/pageflow/revision.rb
@@ -46,7 +46,7 @@ module Pageflow
 
     validates_inclusion_of(:theme_name,
                            in: lambda do |revision|
-                             Pageflow.config_for(revision.entry.account).themes.names
+                             Pageflow.config_for(revision.entry).themes.names
                            end)
 
     def main_storyline_chapters
--- a/app/models/pageflow/revision.rb
+++ b/app/models/pageflow/revision.rb
@@ -44,6 +44,10 @@ module Pageflow
     validate :published_until_unchanged, :if => :published_until_was_in_past?
     validate :published_until_blank, :if => :published_at_blank?
 
+    validates_inclusion_of(:theme_name, in: lambda do |revision|
+                             Pageflow.config_for(revision.entry.account).themes.names
+                           end)
+
     def main_storyline_chapters
       main_storyline = storylines.first
       main_storyline ? main_storyline.chapters : Chapter.none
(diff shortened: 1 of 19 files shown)


aviav:
Is it the reasoning for including the commit that we create entries from the admin?

spec/models/pageflow/home_button_spec.rb
```diff
@@ -5,15 +5,17 @@ module Pageflow
     describe '#url' do
       it 'returns home_url of revision' do
         revision = Revision.new(home_url: 'http://example.com')
-        theming = Theming.new
+        theming = Theming.new(account: create(:account))
```
tf: Using the `:theming` factory would be easier since it would automatically create the account.

app/models/pageflow/revision.rb
```diff
@@ -44,6 +44,11 @@ class Revision < ActiveRecord::Base
     validate :published_until_unchanged, :if => :published_until_was_in_past?
     validate :published_until_blank, :if => :published_at_blank?
 
+    validates_inclusion_of(:theme_name,
+                           in: lambda do |revision|
+                             Pageflow.config_for(revision.entry.account).themes.names
```
tf: Didn't we mean to allow entry level theme feature flags?

tf:
The PR includes a migration that copies the theme names to the revisions. Without the mentioned commit there is no way to create new entries with theme names in their revision, though.

I hadn't noticed it before, but I think it would even make sense to include the changes  to `entry_theme_stylesheet_link_tag` ([[1]](https://github.com/codevise/pageflow/commit/8b96e34389aff71efed6cc4bca2ab670be06c599#diff-9a3cc3cd2b3cd73a5ba47d1641d650cbR45), [[2]](https://github.com/codevise/pageflow/commit/eb0d7c0e9a977cf8cf6ee87a95dbe2c7198c11f3)) since the PR already goes have the way of moving theme_name to Revision. With these changes the moved data would also be used.

The other PR would then mainly be "allow switching themes inside the editor".

tf:
Alternatively, [this validation](https://github.com/aviav/pageflow/commit/490d835b94c33a768ece325342f506e46951ef52#diff-607f7d5ab79a21c625c3946fd664c33bR47) should not have been part of the PR in the first place. Then the migration would not have been required and the PR would really just have been "Themes as (account) features".

app/models/pageflow/revision.rb
aviav: I think I had that in some other branch. Yes, that is what we want here

tf:
Specs are failing

PR #767, opened 2017-05-22
Fix exception when sorting user accounts table by role

tf:
When eager loading memberships entities, we need to ensure (using
`references`) that ActiveRecord always does a join instead of separate
query. Otherwise the conidition inside the `account`/`entry`
associations references a missing column.

Before the table only worked when sorting by entity name since the
order requires a join.

Also make sure to actually use the preloaded association to prevent
N+1 queries.


tf:
Travis got confused by reordering commits.

PR #770, opened 2017-05-23
Pull in jquery-ui/draggable

tilsammans:
This is being used in [admin/entries.js](https://github.com/codevise/pageflow/blob/d88a1f619a02c8e9f070be5c77af78f45c344031/app/assets/javascripts/pageflow/admin/entries.js#L22)

### DIFF
--- a/app/assets/stylesheets/pageflow/jquery_ui.scss
+++ b/app/assets/stylesheets/pageflow/jquery_ui.scss
@@ -12,6 +12,7 @@
 @import "jquery-ui/datepicker";
 @import "jquery-ui/menu";
 @import "jquery-ui/selectmenu";
+@import "jquery-ui/draggable";
 
 div.ui-datepicker{
  font-size: 12px;


tilsammans:
Stand by for an updated pull request; I believe it's enough when this is in `admin.js`.

tilsammans:
On second thought; this PR fixes the issue and it's at home along all the other jquery-ui imports.

I am leaving this as-is unless @tf tells me otherwise.

tilsammans:
Error message without this fix:

```
couldn't find file 'jquery.ui.draggable' with type 'application/javascript'
      Checked in these paths: 
        /Users/joost/Rails/scrollytelling/app/assets/fonts
        /Users/joost/Rails/scrollytelling/app/assets/images
        /Users/joost/Rails/scrollytelling/app/assets/javascripts
        /Users/joost/Rails/scrollytelling/app/assets/stylesheets
        /Users/joost/Rails/scrollytelling/vendor/assets/javascripts
        /Users/joost/Rails/scrollytelling/vendor/assets/stylesheets
        ... (lots more paths)
```

tilsammans:
I had a bit more time to investigate this issue, and it was actually caused by an outdated version of `pageflow-before-after`. Since I switched to master for that one, the correct requires are used for jquery-ui-rails 5.

PR #771, opened 2017-05-24
mark %pageflow_widget_margin_right optional

tilsammans:
widgets don't *have* to provide it.
it will raise an error in the default player otherwise.


tilsammans:
@tf note that I temporarily included the compiled webpack output in 4b66d8e. I don't know if sprockets was supposed to compile this automatically, but for us it didn't, neither development or production.

If you're on board with this pull request I will clean up my branch first.

tf:
Agreed. Guess we never faced this situation before since all our widgets actually define the placeholder. But you are right.

Regarding Webpack: There currently is no integration with Sprockets. If you run your app with a local checkout of Pageflow you need [to run webpack](https://github.com/codevise/pageflow/blob/master/doc/contributing/rails_engine_development.md) at least once. Travis [also does this](https://github.com/codevise/pageflow/blob/master/.travis.yml#L27).

Deploying from a branch is a bit painful at the moment since you need to force add the ignores assets. But I hope all of this is just an intermediate step to a Rails 5.1/Webpacker setup with a real `pageflow` npm/yarn package.

PR #774, opened 2017-05-30
move player controls to host pageflow initialization

tilsammans:
This exposes the player controls, making them configurable from the host
application. It's possible to do this already, directly in the pageflow
initializer:

```
config.widget_types.register(Pageflow::BuiltInWidgetType.slim_player_controls,
default: true)
```

The classic controls would still be available, though. To allow removing
them altogether, I see different options:

- The simplest thing would be to move the registration calls for the
  widget types [1] to the host application initializer. This would be a
  breaking change, but maybe even more aligned with the way things work
  for built in page types already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,13 @@
 
 #### Breaking Changes
 
-- Move player controls initialization to the generated Pageflow initializer. Add these lines to `config/initializers/pageflow.rb` in your host application:
+- The built-in widget types must now be registered in the host application. To keep existing functionality from previous Pageflow versions, add these lines to `config/initializers/pageflow.rb` in your host application:
 
 ```
-# Configure the classic or slim player controls.
+# Register the built-in widget types.
+# You can remove these or add different versions with the same name.
+config.widget_types.register(Pageflow::BuiltInWidgetType.navigation, default: true)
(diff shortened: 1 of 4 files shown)


tilsammans:
Also move the navigation widget types?

```
  config.widget_types.register(Pageflow::BuiltInWidgetType.navigation, default: true)
  config.widget_types.register(Pageflow::BuiltInWidgetType.mobile_navigation, default: true)

```

tf:
Also moving the navigation widget types probably be the most consequent thing to do, right?

tilsammans:
Yes, agree. I will update the PR.

It does mean breakage when people don't add this back in. But I guess a version jump from 0.x to 12.x is pretty significant. If this is too much of a change, we could begin with deprecation notices, and make the change in 13.x?

tf:
We could. But I'd also be ok with adding another breaking change to 12.0 since there already are a couple of major ones.

For the 12.0 release, I plan to collapse the sections for the different release candidates in the changelog into one section. So there will be one list of breaking changes to work through for other's making the update.

tilsammans:
Sounds good to me.

I'll update this PR right now.

tilsammans:
@tf updated

PR #775, opened 2017-05-30
generate a secure password in the seeds

tilsammans:
most like people will change this file before deplooying, but you never
know if one slips through. this change will result in a more secure
installation out of the box, should anyone just run it unchanged.


tf:
This is a very good idea. It is possible to call generator instance methods from the the template (see [the theme generator](https://github.com/codevise/pageflow/blob/master/lib/generators/pageflow/theme/theme_generator.rb#L8) and [its template](https://github.com/codevise/pageflow/blob/master/lib/generators/pageflow/theme/templates/themes/%25name%25.scss.tt#L1) for an example. That way we can get around the `gsub` and let good old `!Pass123` have its well deserved retirement ;)

tilsammans:
@tf updated

tilsammans:
Just noticed the [static password in seeds.rb](https://github.com/codevise/pageflow/blob/94ca556bfa1397ea01dfdde141a893d3e13a9898/lib/pageflow/seeds.rb#L20) by coincidence. What to do with this one, @tf?

tf:
Well, if a user is using the generated `seeds.rb` file this is not used. I'm fine with leaving it in as a fallback. The other option that I see would be raising an exception if you try to use the seed helper that creates a user without having specified a default password.

tilsammans:
Agreed, seems like overkill to raise here, maybe even a little patronising.

PR #776, opened 2017-05-30
Move register calls for built-in widget types to plugin

tf:
In #774, widget type registrations have been moved to the host application, to
allow disabling built in widget types.

Extracting the register calls into a plugin has a couple of
advantages:

* Should there ever be a need to add a built-in widget type, host
  applications which follow the default setup need not be modified.

* More in line with the way built in page types are registered.

* Same flexibility: If you do not wish to use the default set of built
  in widget types, simply replace the plugin with a bunch of custom
  register calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,10 @@
 
 #### Breaking Changes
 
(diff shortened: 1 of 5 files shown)


tf:
@tilsammans what do you think of this idea?

tilsammans:
This idea actually crossed my mind as well. I think it's a great idea.

PR #778, opened 2017-05-31
Entry theme selection

aviav:


app/assets/javascripts/pageflow/editor/models/theme.js
```diff
@@ -0,0 +1,13 @@
...
+  initialize: function(attributes, options) {
+    this.options = options || {};
+  },
+
+  title: function() {
+    return I18n.t('pageflow.' + this.get('name') + '_theme.name');
+  },
+
+  thumbnailUrl: function() {
+    return this.options.thumbnailUrl || this.get('preview_thumbnail_url');
```
tf: This option is only used by [the theme factory](https://github.com/codevise/pageflow/pull/778/files#diff-2b3aab320c738f1a3c0e0a4451449e45R16). But the factory could just pass the `preview_thumbnail_url` attribute directly. Then all code referring to `options` in this file could be removed.

app/assets/javascripts/pageflow/editor/views/model_thumbnail_view.js
```diff
@@ -17,24 +17,37 @@ pageflow.ModelThumbnailView = Backbone.Marionette.View.extend({
...
+        this.newThumbnailView = new pageflow.FileThumbnailView({
+          model: file,
+          className: 'thumbnail file_thumbnail',
+          imageUrlPropertyName: this.options.imageUrlPropertyName
+        });
+      }
+      else {
+        this.newThumbnailView = new pageflow.StaticThumbnailView({
+          model: this.model
+        });
```
tf: The `StaticThumbnailView` is recreated each time update is called, which happens whenever the model configuration changes. If we want to ensure that that the current `model.thumbnailUrl()` is used after a configuration change, `StaticThumbnailView` could listen to that event itself and call its own update method.

app/assets/javascripts/pageflow/editor/views/theme_item_view.js
```diff
@@ -0,0 +1,51 @@
...
+    inUseRegion: '.theme_in_use'
+  },
+
+  events: {
+    'click .use_theme': function(event) {
+      var newThemeName = this.model.get('name');
+
+      var stylesheetPath = this.options.themes.findByName(newThemeName)
+          .get('stylesheet_path');
+      pageflow.stylesheet.update('theme', stylesheetPath);
```
tf: This is the wrong place to handle stylesheet reloading since it couples the otherwise quite generic view to a very specific use case. There already is [an initializer concerned with reloading stylesheets](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/editor/initializers/stylesheet_reloading.js). Let's move this functionality there to decoupled "how the theme is changed" (this view/ThemeInputView etc) from "what has to happen when the theme is changed" (mentioned initializer)

app/assets/javascripts/pageflow/editor/views/theme_item_view.js
```diff
@@ -0,0 +1,51 @@
...
+    );
+  },
+
+  inUse: function() {
+    return this.model.get('name') === this.options.themeInUse;
+  },
+
+  translateThemeName: function() {
+    var name = this.model.get('name');
+    return I18n.t('pageflow.' + name + '_theme.name');
```
tf: Duplication of `Theme#title`

app/assets/javascripts/pageflow/editor/views/change_theme_dialog_view.js
```diff
@@ -0,0 +1,75 @@
...
+    this.ui.previewImage.attr('src', selectedTheme.get('preview_image_url'));
+    this.ui.previewHeaderThemeName.text(that.translateThemeName({
+      name: selectedTheme.get('name')
+    }));
+  },
+
+  translateThemeName: function(options) {
+    var name = options.name;
+    return I18n.t('pageflow.' + name + '_theme.name');
+  }
```
tf: Duplication of `Theme#title`

app/assets/stylesheets/pageflow/editor/base.scss
```diff
@@ -15,6 +15,7 @@
 @import "./indicators";
 @import "./disabled_atmo_indicator";
 @import "./unsupported_browser_hint";
+@import "./change_theme";
```
tf: Should also be inside `.editor` block to ensure no styles can apply inside entry preview. `ChangeThemeDialog` aleady has the `editor` CSS class.

app/assets/stylesheets/pageflow/editor/change_theme.scss
```diff
@@ -0,0 +1,106 @@
...
+  .content {
+    position: absolute;
+    top: 20px;
+    right: 20px;
+    bottom: 50px;
+    left: 20px;
+    overflow: auto;
+    background-color: #fff;
+
+    &>div {
```
tf: Hound is right.

app/assets/stylesheets/pageflow/editor/change_theme.scss
```diff
@@ -0,0 +1,106 @@
...
+
+  li {
+    position: relative;
+    line-height: 60px;
+    background-color: #eeeeee;
+    margin: 10px;
+  }
+
+  li.active {
+    background-color: #d2e5fa;
```
tf: Offtopic: Since color literals simply are a reality inside the Pageflow codebase, I'd be open to adding disabling this rule.

aviav:
See https://github.com/codevise/pageflow/pull/779

app/assets/stylesheets/pageflow/editor/base.scss
aviav: The ReferenceInputView's button also is styled by this. Should it have the editor class also?

PR #783, opened 2017-06-02
Images for Default Theme & max width variable for content

Modularfield:


tf:
Cherry picked into #780

PR #784, opened 2017-06-04
[WIP] Paperclip :pageflow_filesystem_root interpolation

tilsammans:
Previously, this interpolation would only be defined in development, test and production environments. A staging environment did not have this interpolation which would lead to an error. Changed the existing code to a switch inside the definition, so that an interpolation value always exists.

The special cases for Rails.env.test should probably move into a support file under spec.

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -1,8 +1,9 @@
 require 'socket'
+require 'rails'
 
 Pageflow.configure do |config|
   config.paperclip_attachments_version = 'v1'
-  config.paperclip_filesystem_root = Rails.root.join('tmp/attachments/production/')
+  config.paperclip_filesystem_root = Rails.root.join("tmp/attachments/#{Rails.env}/")
 
   if Rails.env.test?
     config.paperclip_s3_default_options.merge!({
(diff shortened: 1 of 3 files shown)


tf:
Agreed. This initializer is a mess. Moving test related things to `spec/support` sounds like a good idea. Some random thoughts:

* Changing the default of `paperclip_filesystem_root` should not change much since the [generated initializer](https://github.com/codevise/pageflow/blob/master/lib/generators/pageflow/initializer/templates/pageflow.rb#L28) probably overrides it for most applications. 
* Always using the config value in the interpolation is probably a good idea. But then the value set in the generator template (`"tmp/attachments/production"`) probably also has to be changed.
* The proposed change removes the `filesystem` directory from the path in test and development, but leaves in [the `s3` directory](https://github.com/scrollytelling/pageflow/blob/0d056707600ca51cf2c687e167acf358ab2862b6/config/initializers/paperclip.rb#L11) in test environment. Changing the path should not be too much of a breaking change for development setups since the filesystem paths are only used until files have been uploaded to S3. Still, it might make sense to be consistent here.
* `PAGEFLOW_ATTACHMENTS_SCOPE_NAME` is also a bit of a strange special case. Should it also be a config option?

So far I've not faced these problems since I prefer to also run staging systems in production environment by [separating behavior and credentials](https://www.schneems.com/2017/03/21/config-behavior-versus-credentials/).

tilsammans:
Sure, I'll work on a bigger cleanup in this area.

I don't disagree in principle with separating credentials and behaviour, and in fact this applies fully to our infrastructure. But ENV is not the safe haven for our secrets we once thought. In fact, [ENV for secrets is down right dangerous](https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/).

I am tantalisingly close to launching with Pageflow 12, but need a few days to look at some weird permission errors that arose on our staging. I will relaunch it under production and see if that clears up the issues. After the launch of that I can start focusing on the changes you mentioned.

tilsammans:
I have assigned this to 12.1 in which I want to move the `test`-specific stuff out of the way.

I am thinking the bigger refactoring can go into 12.2 but open to other thoughts on this subject.

tf:
Sounds fine. :+1:

tf:
Can this one be closed then now that there is #895?

tilsammans:
> Can this one be closed then now that there is #895?

Yes

PR #791, opened 2017-06-30
replace simple jQuery with native JavaScript

tilsammans:
element.classList is quite widely supported:
http://caniuse.com/#search=classlist

Legacy IE does not support multiple parameters for the add() and
remove() methods. I could solve this by using two lines for the removal.

I know there is a solid dependence on jQuery but for these simple class
manipulations we shouldn't need to load all of jQuery?

### DIFF
--- a/app/views/layouts/pageflow/application.html.erb
+++ b/app/views/layouts/pageflow/application.html.erb
@@ -23,8 +23,8 @@
 
 <body class="load_all_images has_no_mobile_platform non_js">
   <script>
-    jQuery('body').removeClass('load_all_images');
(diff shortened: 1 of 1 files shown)


tf:
It's not made explicit anywhere, but we sort of still support IE 9 so far (or at least did not deliberately break support until now). [Can I use](http://caniuse.com/#feat=classlist) says `classList` is >= IE 10, though. While I would not go with jQuery again if I had to start over, I guess, getting rid of jQuery is not going to happen any time soon. So, since it is loaded anyway, I see no reason to switch to DOM API here.

PR #792, opened 2017-07-06
improved handling of worker race conditions

tilsammans:
When an attachment already is on s3, it seems another worker has already
performed the task. otherwise perform it.

### DIFF
--- a/app/jobs/pageflow/upload_file_to_s3_job.rb
+++ b/app/jobs/pageflow/upload_file_to_s3_job.rb
@@ -5,21 +5,20 @@ module Pageflow
     extend StateMachineJob
 
     def self.perform_with_result(file, options = {})
-      if file && file.attachment_on_filesystem.file?
-        if File.exists?(file.attachment_on_filesystem.path)
-          file.attachment_on_s3 = file.attachment_on_filesystem
-          file.save!
-
-          file.attachment_on_filesystem.destroy unless file.keep_on_filesystem_after_upload_to_s3?
-          :ok
-        else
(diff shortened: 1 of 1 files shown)


tilsammans:
We recently tried to scale by having a separate process server. And on this server, we had multiple worker and scheduler processes. This did not work as planned. These workers assume they can manipulate uploaded files directly.

On a separate process server, they obviously can't reach those files. The files are on some random front end server. But it seems that even with the workers on the front ends, we have a problem with horizontal scaling anyway. Files uploaded on server A cannot be reached by a worker starting on server B. Even just assuming one server, our worker processes cannot remove files uploaded by the web server (nginx) process. So this would still fail at line 13.

The pull request I submit here deals with the race condition we saw when using multiple processes. I do not claim to have solved all the horizontal scaling challenges, but it's a start.

tf:
I've added a [wiki page](https://github.com/codevise/pageflow/wiki/Distributing-Resque-Workers) describing our current setup for distributing Resque workers accross multiple instances. I'm not aware of any issues. Resque's usage of Redis queues should prevent situations where multiple workers pick up the same upload job simultaneously. Files (and also deleting of files) can be synchronized between instances using NFS or EFS.

tilsammans:
Thanks! I could use private block storage for this. Not sure if I will, because it introduces a dependency. I will try to find out why we are seeing the race condition; something in our setup maybe.

tf:
Ok. I'll close this PR then and we can continue the discussion in a separate issue once you found out more.

PR #793, opened 2017-07-09
[WIP] expand some sections in the Creating Page Types doc

tilsammans:
Added some sections with guesses for the boilerplate.

Since this probably doesn't apply to Pageflow 12; this targets the `0-11-stable` branch.

### DIFF
--- a/doc/creating_page_types.md
+++ b/doc/creating_page_types.md
@@ -22,6 +22,7 @@ While you're in there, add `pageflow` as a dependency:
 ```
 # pageflow-before-after.gemspec
 spec.add_dependency "pageflow" # you should use ~> to lock to a release
+spec.add_dependency 'pageflow-public-i18n', '~> 1.0'
 ```
 
 ### Using the new PageType in the host application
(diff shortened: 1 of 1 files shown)


tf:
Thanks for working on this! So far I procrastinated on completing these guides since the API is so blurry, including a lot of Sprockets and Rails template path conventions. When `pageflow-react` still was a separate thing, I started to write [some documentation on creating page types](https://github.com/codevise/pageflow-react/blob/master/docs/guides/creating_page_types.md) there. But that is also out of date again already. The best reference implementations right now probably are [`pageflow-timeline-page`](https://github.com/codevise/pageflow-timeline-page) and [`pageflow-vr`](https://github.com/tf/pageflow-vr/).

Even for Pageflow 0.11, I'd strongly recommend building new page types with React to make things more future proof. Therefore, I'm not sure about the worth of documenting the legacy way of doing things, even on the `0-11-stable` branch. Maybe we can combine the new parts from this PR and the `pageflow-react` docs into something that will work for Pageflow 12 and above?

tilsammans:
I wrote this documentation for the most part to try to understand what was going on. I agree that the React way should be used going forward.

I am trying to wrap my mind around what to do but it's very difficult. The "official" instructions in the doc/ directory all mention the old way of doing things. And when I use our 0.11-era widgets on a pageflow-12 install, they mostly seem to work. Only the oEmbed-plugin doesn't.

It's hard to write plugins when there is no official spec/doc/way to do things.

How would I go about making [pageflow-oembed](https://github.com/scrollytelling/pageflow-oembed) compatible with Pageflow 12? It's a reasonably small JS. Am I required to create an object in window and go the React route? Or should the old way of writing widgets still work?

Really in need of some more guidelines at this point. :|

tilsammans:
@tf the reason `pageflow-oembed` doesn't work is because the page content is now rendered by ReactDOM and cannot be modified outside of it. I don't know how to solve this one. Is unmounting React acceptable? Our plugin relies on changing the DOM—turning anchors into embeds—and it's not clear to me how this should mesh with the React components.

tf:
That's really a tough one. The API for widget types did not change with Pageflow 12 (as far as I'm aware). Still, by [reaching into the DOM of pages](https://github.com/scrollytelling/pageflow-oembed/blob/48040897069f9db4c5e1782fdb8dff0788d50fe0/app/assets/javascript/pageflow/oembed.js#L5) `pageflow-oembed` does a lot more than the usual widget type and sort of breaks the encapsulation of page types. As you said, with React we are no longer in control of the DOM. So that approach no longer works.

I guess, there would need to be an API to insert widgets into pages, having pages define outlets or insert points. Or there would need to be some way to override standard components, e.g. to replace the `ContentText` component with a `ContentTextWithOembed` component. Or maybe the `ContentText` component could provide an extension point itself to register embed components that are supposed to be inserted into the page.

A lot of options with different trade-offs. This definitely needs some more thought.

tilsammans:
I think the solution has to be using React components. There seems to be a well-maintained at https://www.npmjs.com/package/react-twitter-widgets

Some documentation about the React setup, required components etc would be appreciated. Also, I prefer to not introduce a new Page Type for it. Possible?

tilsammans:
Closing because this does not apply to React.

As to the discussion that ensued; we will use the Redux store to modify page contents.

PR #794, opened 2017-07-10
move configuration into a concern

tilsammans:
This makes it obvious where to document this module.
And it DRYs things up.

### DIFF
--- a/app/models/pageflow/chapter.rb
+++ b/app/models/pageflow/chapter.rb
@@ -1,6 +1,6 @@
 module Pageflow
   class Chapter < ActiveRecord::Base
-    include Configuration
+    include SerializedConfiguration
 
     belongs_to :storyline, touch: true
     has_many :pages, -> { order('position ASC') }
--- a/app/models/pageflow/chapter.rb
+++ b/app/models/pageflow/chapter.rb
@@ -1,11 +1,11 @@
 module Pageflow
   class Chapter < ActiveRecord::Base
+    include Configuration
+
     belongs_to :storyline, touch: true
     has_many :pages, -> { order('position ASC') }
 
     delegate :entry, to: :storyline
-
-    serialize :configuration, JSON
 
     def configuration
       super || {}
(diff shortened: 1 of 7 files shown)


app/models/pageflow/chapter.rb
```diff
@@ -1,12 +1,12 @@
...
     belongs_to :storyline, touch: true
     has_many :pages, -> { order('position ASC') }
 
     delegate :entry, to: :storyline
 
-    serialize :configuration, JSON
-
     def configuration
       super || {}
     end
```
tf: Shouldn't this be removed as well?

tilsammans:
I'll clean it up.

I should have realised that configuration was taken. Ahhh, naming things...

tilsammans:
How about `Pageflow::Attributes` or a variation thereof? `Pageflow::PageAttributes` perhaps?

tf:
`PageAttributes` is confusing since the mixin is also used for storylines and chapters. `Pageflow::Attributes`  sounds a bit too unspecific. How about `Pageflow::ConfigurationAttribute`?

tilsammans:
One more: since it's basically a key/value or object store, is `Pageflow::ConfigurationStore` or `Pageflow::ConfigurationStorage` a good name?

tf:
Technically, though, the page/chapter/storyline that includes the mixin is not really the key/value store itself, but rather provides an attribute to access the key/value store. Maybe even more specific, something like `WithSerializedConfigruation`. But maybe that's too technical.

tilsammans:
Waiting for #803 to continue.

Maybe just `Pageflow::SerializedConfiguration` as a concern name? It's descriptive at least. We are still sort of abusing configuration to store headers and such, but that is for another time.

tf:
Since #694 `Pageflow::Widget` now also has a serialzed configuration and could therefore also use the mixin.

tilsammans:
I implemented the changes.

PR #795, opened 2017-07-17
rename the section ID to "page-nnnn"

tilsammans:
In CSS, selectors cannot begin with a numeric ID.

There are [workarounds](https://benfrain.com/when-and-where-you-can-use-numbers-in-id-and-class-names/) but following the spec seems better.

### DIFF
--- a/app/views/pageflow/pages/_page.html.erb
+++ b/app/views/pageflow/pages/_page.html.erb
@@ -1,5 +1,10 @@
 <%= cache page do %>
-  <%= content_tag :section, :id => page.perma_id, :class => page_css_class(page), :data => {:template => page.template, :id => page.id, :chapter_id => page.chapter_id} do %>
+  <%= content_tag :section, id: "page-#{page.perma_id}", class: page_css_class(page),
+    data: {
+      template: page.template,
+      id: page.id,
+      chapter_id: page.chapter_id
+    } do %>
     <%= render_page_template(page, entry: entry) %>
     <a class="to_top" href="#content"><%= t('pageflow.public.goto_top') %></a>
   <% end %>


tf:
[There](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/slideshow.js#L47) [is](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/slideshow.js#L88) [lots](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/slideshow/page_widget.js#L14) [of](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/editor/views/page_preview_view.js#L34) [code](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/history.js#L6) that depend on the current shape of the id. The ids also act as anchors for non-js deep links (not sure if that is really used). Why do you want to reference ids from CSS?

tilsammans:
I see. Maybe introduce a class for it?

I have used it to [style some pages—not all—selectively](https://github.com/scrollytelling/app/commit/d3c4a1389dd83a2722a19f845ff660c2c0a62e7a). Theses pages needed a media query too.

(My finest attempt at a passive aggressive dig towards wysihtml follows 😇:) if the html editor was more powerful, I could maybe have editors handle this themselves. Either in the html editor, via some inline element, or the html source. Or by allowing us to extend editor functions, or tag the page with free-form key/values, that end up targetable by CSS.

A bold, italic and underline button isn't enough, really.

tf:
While I totally agree with your criticism of the current rich text editor capabilities, I still think styling individual pages inside the theme should be a last resort. And for a last resort, I feel, the current workaround `.page[id=15]` works well enough, or at least does not justify including an additional css class on each and every page.

And while I have gone through a fair share of aggression regarding wysihtml myself, it has to be said, that it actually allows for a way bigger feature set than we currently use. Looking into enabling at least lists and headings would probably be a great first step.

tilsammans:
That's fair enough.

And I would welcome an upgrade to wysihtml, because I think that at least fixes the bug where you cannot get rid of empty `<b></b>`. (I hope)

Pageflow still uses the 0.4.17 version from November 2014. Are you open to using https://github.com/Voog/wysihtml-rails instead, either `0.5.5` or `0.6.0.beta2`? (the latter works fine, btw)

tf:
Yes, I'd be open to give the new version a try.

PR #796, opened 2017-07-17
[WIP] Provide consistent interface for entry and account policies

aviav:
Since we did not need some of the account policies, we did not provide them up to this point. To provide a more consistent policy interface, this commit adds three methods that were undefined so far.

Also, the standard entry and account policies now always include admin. This was also inconsistent.


aviav:
Will try roles as an interface that is possibly cleaner

PR #799, opened 2017-07-19
Allow specifying opacity of image variant logo

tf:
Provide a separate variable for the logo on the first page, to allow
e.g. having a transparent logo on all but the first page.

### DIFF
--- a/app/assets/stylesheets/pageflow/themes/default/logo/variant/background_image.scss
+++ b/app/assets/stylesheets/pageflow/themes/default/logo/variant/background_image.scss
@@ -1,3 +1,9 @@
+// Opacity of the logo
+$logo-background-image-variant-opacity: null !default;
+
+/// Opacity of the logo on the first page
+$logo-background-image-variant-first-page-opacity: null !default;
+
 @mixin logo-variant-background-image(
   $first-page-only,
   $top,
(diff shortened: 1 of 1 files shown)


tf:
In a way Pageflow already supports DOM element based headers. Those are used for the so called ["watermark" logo variant](https://github.com/codevise/pageflow/blob/master/app/assets/stylesheets/pageflow/themes/default/logo/variant/watermark.scss#L19). There is also [some JS](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/widgets/header.js#L2) to update CSS classes and [a helper method](https://github.com/codevise/pageflow/blob/master/app/helpers/pageflow/entries_helper.rb#L76). Still the header partial itself has to be provided at the moment by overwriting [`app/views/pageflow/entries/_header.html.erb`](https://github.com/codevise/pageflow/blob/master/app/views/pageflow/entries/_header.html.erb). This was meant to honor the fact that headers (including their navigation menu) are often really customized a lot. Probably, the right thing to do would be to introduce a new widget type role `header` and allow providing custom headers via the widget mechanism, which did not exist yet when the current solution was developed.

tilsammans:
A new widget type role sounds like a good step forward!

PR #802, opened 2017-07-25
Introduce config option to allow multiaccount users

aviav:
Defaults to true. Keeps previous default behavior. Depends on changes from https://github.com/codevise/pageflow/pull/815


app/views/components/pageflow/admin/users_tab.rb
```diff
@@ -38,7 +38,8 @@ def build(theming)
               end
             end
           end
-          add_membership_button_if_needed(current_user, account, 'account')
+          add_membership_button_if_needed(current_user, account, 'account') if
+            Pageflow.config.allow_multiaccount_users
```
tf: `add_membership_button_if_needed` already contains a lot of logic. I'd suggest moving this piece of logic there as well. It's also easier to test logic like the on the view component level.

lib/pageflow/configuration.rb
```diff
@@ -252,6 +252,11 @@ class Configuration
     # take. Defaults to `[:captions, :subtitles, :descriptions]`.
     attr_reader :available_text_track_kinds
 
+    # Allow one user to be member of multiple accounts. Defaults to
+    # true.
+    # @since 12.1
```
tf: Please use `@since edge`. Tags will automatically be rewritten once a Pageflow version is released. That way there is no need to guess which minor version the feature will be in.

app/models/pageflow/membership.rb
```diff
@@ -11,6 +11,9 @@ class Membership < ActiveRecord::Base
 
     validates :user, :entity, :role, presence: true
     validates :user_id, uniqueness: {scope: [:entity_type, :entity_id]}
+    validates :user_id,
+              uniqueness: {scope: :entity_type},
+              if: :on_account? && !Pageflow.config.allow_multiaccount_users
```
tf: How does the error message look when trying to invite an already existing user?

tilsammans:
It is necessary to keep the old behaviour as a config option? I wouldn't mind allowing multi-account across the board. It seems to be more in line with the new permissions model.

tf:
@tilsammans we mostly need this as a temporary solution for Hosted Pageflow since much of its UI assumes that users only have one account. Moreover, we plan to add an invitations feature to prevent users from being added to accounts without their consent.

tilsammans:
Is everyone getting these coveralls mails too? They're too noisy. If I disable them in the repo, I fear that they will be turned off for everyone. How to disable them just for me?

tf:
Yeah, I hate them, too. I think you can block the coveralls user in your user settings. Disabling them in the repo, blocks coveralls from commenting on the PR completely, I think.

aviav:
Specs should be passing on Travis Rebuild

app/views/admin/users/_form.html.erb
```diff
@@ -1,35 +1,47 @@
-<%= admin_form_for([:admin, @user.becomes(User)]) do |f| %>
-  <%= f.inputs "Details" do %>
-    <% if @user.new_record? %>
-      <% Pageflow::AccountPolicy::Scope
-           .new(current_user, Pageflow::Account).member_addable.each do |account| %>
-        <%= quota_state_description(:users, account) %>
+<% if @user.new_record? &&
+      Pageflow.config.quotas.get(:users, @user.initial_account).exhausted? %>
+  <%= render 'admin/users/quota_exhausted' %>
```
tf: * Let's pass an `account` local to the partial. This will make implementing the partial itself much easier.
* Let's rename the partial to `pageflow/admin/users/quota_exhausted`. The way we use it, it is API surface. So I think staying in our own namespace makes sense.
* Given that the partial is API surface, let's make sure it cannot get lost again and add a controller spec. Maybe we can use something like [stub_template](https://github.com/codevise/pageflow/blob/0f417b64d04c447cc5f0045ef7d695a4a0a62e32/spec/support/helpers/stub_template_in_helper_specs.rb) to actually make sure the correct locals are passed. Otherwise we should at least be able to use [the render_template matcher](https://relishapp.com/rspec/rspec-rails/docs/matchers/render-template-matcher)

tf:
see #848

PR #803, opened 2017-07-25
Test workflow improvements

tilsammans:
With these changes, testing locally should become a lot more comfortable, especially on macOS.

- [x] document the Rubies we are using for tests
- [x] document the need for `bundler` gem
- [ ] ~~commit `package-lock.json` file for reproducible tests~~ to be done separately
- [x] switch to Poltergeist for headless tests
- [x] test for existence of socket file
- [x] ease the requirement for tempfile location


tilsammans:
I was done fighting to install Qt for the umpteenth time. It's so heavy a dependency, whereas Poltergeist simply provides the correct binaries. No more need to install the full Xcode app install either, saves about 5 GB!

I eased the strict test for a Tempfile path, as these are in `/var` on macOS and in general, this decision is for the OS, it cannot be relied upon in specs.

Unsure if we are to ignore the npm lockfile or include it; I included it here.

spec/support/pageflow/dummy/templates/database.yml
```diff
@@ -8,8 +8,10 @@ development:
 <% if ENV['PAGEFLOW_DB_HOST'].present? %>
   host: <%= ENV['PAGEFLOW_DB_HOST'] %>
   port: <%= ENV.fetch('PAGEFLOW_DB_PORT', '3306') %>
-<% else %>
+<% elsif File.socket?('/var/run/mysqld/mysqld.sock') %>
   socket: /var/run/mysqld/mysqld.sock
+<% elsif File.socket?('/tmp/mysql.sock') %>
+  socket: /tmp/mysql.sock
 <% end %>
```
tf: I'd suggest adding an `else` case here and failing with a helpful error message. Maybe we can use a YAML anchor to prevent duplication of these blocks since are starting to get more complex.

tilsammans:
Great suggestions. I will work on this more tonight.

tilsammans:
> Which version of npm do you use?

I was on 5.1, which explains the difference in lock files. I will remove this bit from the pull request.

spec/support/pageflow/dummy/templates/database.yml
tilsammans: I will add the macro and fall back to `localhost:3306`, which should cover everything.

tilsammans:
> Now there are a bunch bandwidth detection outputs in the logs.

This is when you run the test locally and on Travis, right? I don't mind those, and the output doesn't seem to affect test results. I would say it can stay as is. (Maybe even move this output away from console.log and into the `pageflow` object).

Other changes done.

tilsammans:
I am considering [using PhantomJS 2.1.1 on Travis](https://github.com/ruby-bench/ruby-bench-web/pull/190/files), what do you think? I am running 2.1.1 locally...

tf:
@aviav can you make sure the tests still pass on your machine after these changes? I vaguely remember their was something special about your setup.

aviav:
Everything passing with flying colors, even some specs that were previously failing in my development. Of course that looked like it was related to some Qt-webkit stuff.

It also looks like I cannot set my review to 'approve' since @tf already approves, but I would approve if I could :-)

PR #806, opened 2017-07-26
Call hook on entry publication

tf:
Allow plugins to register custom functionality to be performed when an
entry has been published.


tilsammans:
So many new features! It's hard to keep track! :)

PR #810, opened 2017-07-26
read the attribute instead of super

tilsammans:
super is a bit counter-intuitive since the method might not exist on the superclass, but we're just relying on the message being picked up anyway. By reading the attribute the intent is made more clear.

We are not using `read_attribute` but the brackets, because the latter raises an exception when the attribute cannot be found. The rails style guide recommends so. https://github.com/bbatsov/rails-style-guide#read-attribute

### DIFF
--- a/app/models/concerns/pageflow/feature_target.rb
+++ b/app/models/concerns/pageflow/feature_target.rb
@@ -49,7 +49,7 @@ module Pageflow
     end
 
     def features_configuration
-      super || {}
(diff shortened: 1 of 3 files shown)


tilsammans:
I am submitting this, since after the merge of #794 some standalone controller specs fail their tests for me now. Not sure why yet, it seems to be related to autoloading. The `super` threw RSpec off; this change helps.

PR #813, opened 2017-07-28
[WIP] Allow HTTP caching based on Last-Modified

tilsammans:
Uses `fresh_when` to both set `Last-Modified` on a client's initial request, and respond with `304 Not Modified` if the content hasn't changed. For this, we added delegation of the ActiveRecord timestamps on Entry to Revision.

This change speeds up browsers. By adding `public: true` we also allow caching proxies to store the response. This setup even prevents the application server being hit at all. It can be accomplished using a simple cache like rack-cache, but also works great with cdn-based reverse proxy servers, nginx, Varnish, CloudFlare, etc.

### DIFF
--- a/app/models/pageflow/published_entry.rb
+++ b/app/models/pageflow/published_entry.rb
@@ -25,6 +25,7 @@ module Pageflow
              :author, :publisher, :keywords,
              :theme,
              :password_protected?,
+             :created_at, :updated_at,
              :to => :revision)
 
     def initialize(entry, revision = nil)
--- a/app/controllers/pageflow/entries_controller.rb
+++ b/app/controllers/pageflow/entries_controller.rb
@@ -39,6 +39,8 @@ module Pageflow
           else
             @entry.share_target = @entry
           end
+
+          fresh_when(@entry, public: true)
         end
         format.json do
           authenticate_user!
--- a/spec/controllers/pageflow/entries_controller_spec.rb
+++ b/spec/controllers/pageflow/entries_controller_spec.rb
@@ -167,6 +167,24 @@ module Pageflow
           expect(response.status).to eq(200)
         end
 
+        it 'includes a last-modified response header' do
+          entry = create(:entry, :published)
+
+          get(:show, id: entry)
+
+          expect(response.headers).to include('Last-Modified')
+          expect(response.headers['Last-Modified']).to eq(entry.updated_at.httpdate)
+        end
+
+        it 'sets Cache-Control to public' do
+          entry = create(:entry, :published)
+
+          get(:show, id: entry)
+
+          expect(response.headers).to include('Cache-Control')
+          expect(response.headers['Cache-Control']).to include('public')
+        end
+
         it 'responds with not found for non-published entry' do
           entry = create(:entry)
 
@@ -280,6 +298,28 @@ module Pageflow
           end
         end
 
+        context 'with if-modified-since header' do
+          let(:entry) { create(:entry, :published) }
+
+          it 'responds with success if the entry was modified' do
+            time = entry.updated_at.advance(days: -1)
+
+            request.env['HTTP_IF_MODIFIED_SINCE'] = time.httpdate
+            get(:show, id: entry)
+
+            expect(response.status).to eq(200)
+          end
+
+          it 'responds with not modified if the entry was not modified' do
+            time = entry.updated_at.advance(days: 1)
+
+            request.env['HTTP_IF_MODIFIED_SINCE'] = time.httpdate
+            get(:show, id: entry)
+
+            expect(response.status).to eq(304)
+          end
+        end
+
         context 'with page parameter' do
           it 'renders social sharing meta tags for page' do
             entry = create(:entry, :published)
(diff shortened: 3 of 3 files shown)


tilsammans:
Etag generation in Pageflow isn't reliable, as it uses implicit template rendering. This was only perfected in Rails 5.0.4 and 5.1.3. See a host application's log:

```
Couldn't find template for digesting: entries/show
```

This means http caching server-side isn't reliable either. But the `Last-Modified` already helps in clients. To fix the Etag issue, we could switch to explicit template rendering. Which just means replacing `render 'template name'` with `render partial: 'template name'` anyway.

app/models/pageflow/published_entry.rb
```diff
@@ -25,6 +25,7 @@ class PublishedEntry
              :author, :publisher, :keywords,
              :theme,
              :password_protected?,
+             :created_at, :updated_at,
```
tilsammans: Perhaps delegate `created_at` to Entry instead? 🤔

tf: Since `Revision` is `belongs_to :entry, touch: true`, both are probably safer to delegate to `Entry`, right?

tf:
Interesting suggestion. I haven't worked with `fresh_when` so far. So I'm not sure about implicit vs explicit template rendering or whether this has to wait until the Rails 5 upgrade. We need to make sure ETags cannot cause display of obsolete responses after a deploy etc. There is also stuff like server rendered React components, which probably also makes this more complicated.

app/models/pageflow/published_entry.rb
tilsammans: Indeed. If `touch: true` is ever removed, or it stops working, it won't invalidate the cache then. I guess it's a fair tradeoff, but if you'd prefer I can write two methods that are more bullet proof than a simple delegation. The methods could take a published revision, or fallback to the entry when there is none. It's probably the best solution, but more LOC.

tilsammans:
The React views only make it better, since they don't use Rails partials.

A web server should return both Etag and Last-Modified. [This is a great writeup](https://www.keycdn.com/blog/http-cache-headers/) and it also has a header check tool.

We are using this header in production and we see a positive effect; i.e. lower egress bandwidth for the same number of visitors. And no complaints about stale cache. This header does not influence CDN/proxy anyway since there is no max-age, they won't cache it.

```
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Connection: close
Status: 200 OK
Last-Modified: Mon, 14 Aug 2017 19:20:43 GMT
Cache-Control: public
```

I have already taken a quick look at moving towards explicit template rendering. It updates a _lot_ of view files but it's quite doable. It should stabilise the Etag with potentially more bandwidth savings than this change alone.

If you're 🆗 with this one, I can submit a PR for that. Let me know if there are private branches queued up that touch a lot of view files. Because mine definitely will change many dozens of them.

tilsammans:
> We need to make sure ETags cannot cause display of obsolete responses after a deploy etc.

This PR doesn't touch ETag. And I think fixing the warning in the logs will make that possibility _less_ likely, since it then considers all the partials that make up an entire response. Change a partial (after a deploy), will for sure change the Etag. Right now many partials are either ignored or Rails makes a guess at them.

tf:
Can you give a short example how this explicit template would look? I still not sure I understand how this will work with React based views. If we change the JavaScript the server side rendered results change, but non of the partials. How does Rails know when to invalidate ETags?

I am not aware of any unmerged branches that touch a lot of view code. So from that perspective it should not be a problem.

tilsammans:
@tf a switch to explicit template rendering is changing this:

``` ruby
<%= render 'pageflow/entries/skip_links' %>
```

to this:

``` ruby
<%= render partial: 'pageflow/entries/skip_links' %>
```

The former makes many assumptions, including local vars, while the latter doesn't. So in some cases it also means adding a locals hash. These are quite easy to find by looking at the variables used in the partials. The upside is that Rails is able to consistently calculate the Etag.

That's it really. React does not come into view for this change; since the only controller action that will change is the initial request for the entry. Which is rendered server-side.

tilsammans:
To reiterate; I don't think we should improve partial rendering in this patch, but I would like to have the last-modified header in the response. It bugs me that we don't send it, when in fact this timestamp is very much part of published stories.

app/models/pageflow/published_entry.rb
tf: Either way, we should make sure that `PublishedEntry#updated_at` changed when the underlying `Entry` changes (i.e. a feature flag is enabled or the title is changed). The way it is now, only `Revision` changes are accounted for.

tf:
If I read the docs and code of `fresh_when` correctly, it sets last modified and etag headers. If the etag generation issues have not been sorted out yet, doesn't that mean risking displaying stale content after a deploy where templates changed but timestamps did not?

Also, I'm still really confused how a different etag can be generated automatically and only server side rendered React changed. Shouldn't this cause Rails to look at the templates, come to the conclusion that nothing has changed, generate the same etag and respond with "Not Modified", even though the changed React components would have resulted in a different HTML response?

tilsammans:
> If the etag generation issues have not been sorted out yet, doesn't that mean risking displaying stale content after a deploy where templates changed but timestamps did not?

Yes, in fact.

Okay. If you don't mind I will change all partial rendering to use the explicit form, which will result in a changed Etag, which alleviates your objections.

tilsammans:
@tf I have great news! The fixes in Rails 5 were backported to Rails 4 and I have just confirmed that it works on 4.2.10 at least. The `Couldn't find template for digesting` error is completely gone from the log as well.

I have changed a little something in the `_entry.html.erb` partial and the ETag did in fact change!

```
ETag: W/74742a5e06f0acfaddf2baa524b6ce14 # pageflow master
ETag: W/70dc14374cd504600898d02185fb0e7a # small change inside a partial
```

This means that changes in templates _will_ invalidate the http cache because the ETag changes too.

I have this patch running in production since July and it has no side-effects. I really would like this one included, not only because of cache benefits but also because we're making better use of http headers. Last-Modified by itself should really be a no-brainer I think.

If you're 👎  on this pull request, then let's close it. I can live with the monkey patch for us.

tf:
Yes, I'm very much in favor of this PR. The remaining open questions mainly concern server side JS. Can you make the same test and just make change to `pageflow/dist/react.js` which is used in server side rendering. I'm wondering if we need something along the lines of the second part of [this](https://gist.github.com/jeremy/4211803). 

Also [this little part](http://guides.rubyonrails.org/caching_with_rails.html#external-dependencies) of the Rails caching guide sounds a bit scary to me.

PR #815, opened 2017-07-28
Extend query interface

aviav:
- Introduce AccountRoleQuery
- Introduce AccountMemberQuery
- Make AccountMemberQuery and EntryRoleQuery depend on ApplicationQuery

### DIFF
--- a/app/models/pageflow/account_member_query.rb
+++ b/app/models/pageflow/account_member_query.rb
@@ -1,0 +1,62 @@
+module Pageflow
+  class AccountMemberQuery
+    class Scope
+      # Account whose members we scope
+      # @return {Pageflow::Account}
+      attr_reader :account
+
+      # Base scope which is further scoped according to account role
+      # @return [ActiveRecord::Relation<User>]
+      attr_reader :scope
+
+      # Create scope that can limit base scope to account members at
+      # or above a given role
+      #
+      # @param {Pageflow::Account} account
+      #   Required. Membership account to check.
+      # @param [ActiveRecord::Relation<User>] scope
+      #   Optional. Membership entity to check.
+      def initialize(account, scope = User.all)
+        @account = account
+        @scope = scope
+      end
+
+      # Scope to those members from scope on account who have at least
+      # role
+      #
+      # @param {String} role
+      #   Required. Minimum role that we compare against.
+      # @return [ActiveRecord::Relation<User>]
+      def with_role_at_least(role)
+        scope.joins(memberships_for_account_with_at_least_role(role))
+             .where(membership_is_present)
+      end
+
+      private
+
+      # @api private
+      def memberships_for_account_with_at_least_role(role)
+        options = {roles: Roles.at_least(role), account_id: account.id}
+
+        sanitize_sql(<<-SQL, options)
+          LEFT OUTER JOIN pageflow_memberships as
+            pageflow_account_memberships ON
+          pageflow_account_memberships.user_id = users.id AND
+          pageflow_account_memberships.role IN (:roles) AND
+          pageflow_account_memberships.entity_id = :account_id AND
+          pageflow_account_memberships.entity_type = 'Pageflow::Account'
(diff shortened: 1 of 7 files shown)


app/models/pageflow/account_member_query.rb
```diff
@@ -0,0 +1,57 @@
+module Pageflow
+  class AccountMemberQuery < ApplicationQuery
+    class Scope < Scope
+      # Account whose members we scope
+      # @return {Pageflow::Account}
```
tf: In the YARD doc they use `[]` for types. This is inconsistent in other files as well, but let's try to be consistent.

app/models/pageflow/account_member_query.rb
```diff
@@ -0,0 +1,57 @@
+module Pageflow
+  class AccountMemberQuery < ApplicationQuery
+    class Scope < Scope
```
tf: This works but does not read well. Let's be explicit and write `ApplicationQuery::Scope`.

app/models/pageflow/account_member_query.rb
```diff
@@ -0,0 +1,57 @@
...
+      #   Required. Minimum role that we compare against.
+      # @return [ActiveRecord::Relation<User>]
+      def with_role_at_least(role)
+        scope.joins(memberships_for_account_with_at_least_role(role))
+             .where(membership_is_present)
+      end
+
+      private
+
+      # @api private
```
tf: `private` stuff (in the Ruby sense) is `@api private` by default.

app/models/pageflow/application_query.rb
```diff
@@ -0,0 +1,12 @@
+module Pageflow
+  class ApplicationQuery
+    class Scope
+      protected
+
+      # @api protected
```
tf: `@api protected` is not a thing so far. Let's make `ApplicationQuery` `@api private` completely since we do not expect anyone to inherit from it.

app/policies/pageflow/account_policy.rb
```diff
@@ -123,13 +124,13 @@ def destroy_membership_on?
     end
 
     def admin?
-      @user.admin?
+      user.admin?
     end
 
     def see_badge_belonging_to?
-      (@account.entries & @user.entries).any? ||
-        @user.memberships.on_accounts.as_previewer_or_above.where(entity: @account).any? ||
```
tf: I think, `FolderPolicy` could also benefit from using the query. Please check if some these `_or_above` methods on `Membership` are unused then and can be removed.

app/models/pageflow/account_member_query.rb
```diff
@@ -0,0 +1,57 @@
+module Pageflow
+  class AccountMemberQuery < ApplicationQuery
```
tf: Please also add a one line doc comment to the query classes themselves.

app/policies/pageflow/account_policy.rb
aviav: Would this be okay in principle (not tested yet)? Would we want to catch the error/throw something else?
```
    def has_at_least_role?(role, on_as_many_accounts_as: 1)
      case on_as_many_accounts_as
      when 0
        does_not_have_role?(role)
      when 1
        has_role_at_least_once?(role)
      else
        if on_as_many_accounts_as > 1
          has_role_at_least?(role, on_as_many_accounts_as)
        else
          raise ArgumentError('Integer 0 or greater expected for '\
                              'on_as_many_accounts_as')
        end
      end
    end
```

aviav: (We would need `on_as_many_accounts_as: 2` for now, but doing it for `1` and `2` only seems excessively lazy)

aviav: Adding the Scope requires specs and maybe thought, so I will leave it for the next apparent opportunity. I agree, however, that we can probably use the extended query interface in other policies as well and thus improve readability, design and such

PR #817, opened 2017-08-02
Extend EncodingConfirmation by public account attribute

aviav:


aviav:
Reasoning: account of encoding confirmation's entry can be interesting for apps that use pageflow. Thus, it makes sense to add it to the interface of encoding confirmation from a design perspective.

We could use specs here, but stick to established practice (no specs) due to resource constraints

PR #825, opened 2017-08-09
Preload page background images

tf:
Setting the `load_image` css class to make sure the `background-image`
rule applies does not trigger image loading (e.g. in Chrome). Detect
image url via computed styles to account for media queries and ensure
image is loaded via `new Image()`.

Originally similar functionality was part of most page types'
`preload` hooks, but has not been ported to React based page types so
far.

### DIFF
--- a/node_package/src/components/BackgroundImage.jsx
+++ b/node_package/src/components/BackgroundImage.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
+import {preloadBackgroundImage} from 'utils';
 
 /**
  * Display an element with a background image referenced by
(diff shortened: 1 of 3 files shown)


tilsammans:
Are you open to including `link rel="preload"` as well? It will be the standard and Chrome already supports it.

It's [coming to Firefox too](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content) and needn't go into the head, these links work fine in the body.

https://developers.google.com/web/updates/2016/03/link-rel-preload

tf:
Interesting. To be backwards compatible we would still need to keep the `new Image` part for now, right? It appears to be rather hard to feature detect [[1]](https://github.com/w3c/preload/issues/7), [[2]](https://github.com/w3c/preload/pull/52), [[3]](https://gist.github.com/igrigorik/a02f2359f3bc50ca7a9c). And the nasty computed styles bit would also have to stay since this does not give us a better way to find the correct image based on the matching media query. So all in all, not yet convinced the gains are worth the effort here.

tilsammans:
Well, it's at least a move towards a standard, which seems the correct approach in itself.

The Rack specification does not allow multiple `Link` headers. The HTTP spec has evolved in that regard, where Link is one of the few exceptions that do. To be rack-compatible (and thus Rails) the images need to be comma-separated in the one Link header.

It does seem like too much effort at this time. It's something to consider maybe when we're using an updated Rack spec (if ever).

PR #834, opened 2017-08-12
destroy widgets when parent subject is destroyed

tilsammans:
leaves invalid rows in the database otherwise.


tilsammans:
@tf updated

tilsammans:
The failure on Travis seems unrelated.

PR #835, opened 2017-08-13
Fix memberships

aviav:
- remove redundant `permitted_params`
- change form: entry/account for query params which are visible in URLs, 'Pageflow::Entry'/'Pageflow::Account' elsewhere to minimize confusion

### DIFF
--- a/admins/pageflow/membership.rb
+++ b/admins/pageflow/membership.rb
@@ -14,12 +14,6 @@ module Pageflow
       helper Pageflow::Admin::MembershipsHelper
       helper Pageflow::Admin::FormHelper
 
-      def permitted_params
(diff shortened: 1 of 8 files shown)


tf:
@aviav Are there any functional changes here? Or is it just about the removal of unused code and simplification?

aviav:
In membership admin, we had two `permitted_params` methods. I think I got failures with both of them around. In any case, that was faulty code.

As for the other changes, I am not sure whether they fix anything functional. After I extracted them in this way, I'd guess they mainly change internal naming and passed data and would not be surprised if the failures I encountered were fixed by stuff in the other PRs. I can test for this if it's important

tf:
I'll give it a spin in Dedicated.

PR #836, opened 2017-08-13
Allow :create Entry only for publishers on accounts, not on entries

aviav:


app/policies/pageflow/entry_policy.rb
```diff
@@ -87,7 +87,8 @@ def publish?
     end
 
     def create?
-      publish? && AccountPolicy::Scope.new(user, Account).entry_creatable.any?
+      query.has_at_least_account_role?(:publisher) &&
+        AccountPolicy::Scope.new(user, Account).entry_creatable.any?
```
tf: So the idea is that an entry publisher should not be able to create an entry? I agree with that. But shouldn't [this spec](https://github.com/aviav/pageflow/blob/0c7019b7a49ff2b139d1b27b80662d8aaf03f957/spec/policies/pageflow/entry_policy_spec.rb#L48) fail then?

I don't really get the second part of the condition. Doesn't that say, that I'm allowed to create an entry if there is at least one account where i can create an account (which also just means [being publisher in at least one account](https://github.com/aviav/pageflow/blob/0c7019b7a49ff2b139d1b27b80662d8aaf03f957/app/policies/pageflow/account_policy.rb#L23))

tf:
Not sure why I am assigned here. What is then answer to the question?

aviav:
I thought I understood that you wanted to look into those questions yourself. Maybe we need to clarify something here

PR #839, opened 2017-08-13
Show new entry button only to publisher or above

aviav:


tf:
Test

tf:
Test 2

tf:
I'll test this in Dedicated.

aviav:
I just pushed a fix

tf:
Needs to be Ruby 2.1 compatible.

aviav:
I see. For how long do we support Ruby 2.1, anyway? It does not even receive security fixes anymore

https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

tf:
Not decided yet. We can discuss dropping it in the next major version.

tf:
see #847

PR #840, opened 2017-08-14
Generate admins with account membership in specs

aviav:


spec/policies/pageflow/folder_policy_spec.rb
```diff
@@ -22,7 +22,7 @@ module Pageflow
     it 'does not permit admins when there is no more than one account' do
       Account.delete_all
       folder = create(:folder)
-      admin = create(:user, :admin)
+      admin = create(:user, :admin, on: folder.account)
```
tf: Why is this needed now? Given admins an account makes sense, but why does the folder policy have to change?

aviav: This is only about the spec: Changing the factory creates a second account when the admin is created. Providing `folder.account` with `on` prevents that and makes the spec pass again

PR #841, opened 2017-08-15
Permissions config

aviav:


app/policies/pageflow/entry_policy.rb
```diff
@@ -120,15 +120,17 @@ def update_account_on?
...
 
     def manage_account_of?
       query.has_at_least_account_role?(:manager)
     end
 
     def update_feature_configuration_on?
-      manage_account_of?
+      Pageflow.config.permissions.show_entry_features_to_non_admin &&
+        manage_account_of?
     end
```
tf: These policy changes need specs. Let's add a protected `permissions_config` method to `ApplicationPolicy`:

    def permissions_config
      Pageflow.config.permissions
    end

This way we can avoid duplicating the concrete look up logic.

app/policies/pageflow/user_policy.rb
```diff
@@ -62,6 +62,14 @@ def set_admin?
...
     end
 
+    def see_admin_status?
+      if Pageflow.config.permissions.show_admin_boolean_to_non_admin
+        read?
+      else
+        admin?
+      end
+    end
+
```
tf: see above.

admins/pageflow/user.rb
```diff
@@ -41,7 +41,9 @@ module Pageflow
           row :locale do
             I18n.t('language', locale: user.locale)
           end
-          boolean_status_tag_row(:admin?, 'admin warning')
+          if authorized?(:see_admin_status_of, user)
+            boolean_status_tag_row(:admin?, 'admin warning')
+          end
         end
```
tf: Let's add a [controller spec](https://github.com/aviav/pageflow/blob/007e4a08566642a6131a6996d2ba0ad326446997/spec/controllers/admin/users_controller_spec.rb#L5), that makes sure this code continues to work.

config/initializers/admin_resource_tabs.rb
```diff
@@ -14,3 +10,17 @@
...
+                                        name: :features,
+                                        component: Pageflow::Admin::FeaturesTab,
+                                        required_account_role: :manager)
+  else
+    config.admin_resource_tabs.register(:entry,
+                                        name: :features,
+                                        component: Pageflow::Admin::FeaturesTab,
+                                        admin_only: true)
+  end
+end
```
tf: We already [have a test](https://github.com/aviav/pageflow/blob/007e4a08566642a6131a6996d2ba0ad326446997/spec/controllers/admin/entries_controller_spec.rb#L45) for original case. Let's also cover the other one.

lib/pageflow/configuration/permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+
+      # Allow non-admin users to see entry features. Defaults to true.
+      # @since edge
+      attr_accessor :show_entry_features_to_non_admin
+
+      # Allow non-admin users to know about admin boolean. Defaults to
+      # true.
+      # @since edge
+      attr_accessor :show_admin_boolean_to_non_admin
```
aviav: I think it is appropriate. Maybe decoupling our naming scheme from a cancan-like syntax has its own benefits (such as thinking more freely about authorization), but I thought about this as well and am myself undecided

aviav: Do you think it makes sense to treat user_account_badge_list in this PR? I am trying to weigh merge urgency vs completeness in terms of benefits

tf: Feel free to skip it. It should not come up in our case anyway.

lib/pageflow/configuration/permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+          options.fetch(:show_entry_features_to_non_admin, true)
+        @show_admin_boolean_to_non_admin =
+          options.fetch(:show_admin_boolean_to_non_admin, true)
+        @show_theming_dropdown_to_non_admin =
+          options.fetch(:show_theming_dropdown_to_non_admin, true)
+      end
+
+      # Allow non-admin users to see entry features. Defaults to true.
+      # @since edge
+      attr_accessor :show_entry_features_to_non_admin
```
aviav: In case I hear no objection, I shall go with `non_admin_may_update_features` here. Reasoning: - name is more likely to remain valid even if we adjust roles and permissions so that `account_managers` prefix became a misnomer
- `may` is consistent with this being a permission. `can` implies that we assume our authorization layer is impervious to attacks -- which we do not
- `may` may unlock our thinking from CanCan-like abilities

tf: I do not really care about "may" vs "can" as long as we stay consistent.

Regarding the first bullet point, I disagree. If we ever adjust permissions such that other roles besides account managers may update features (which I have a hard time imagining at the moment), simply changing the meaning of this option would be a rather nasty breaking change IMO. We should strive to make this permission interface as explicit as possible.

tf: Given that letting account managers edit features will be the default, we could also consider negating the name: `only_admins_may_update_features`. That way it would be clear that these options can be used to restrict permissions above what is the default.

aviav: This one I like very much. Will do another renaming pass.

I do not have a hard time imagining permission adjustments. That may be due to my imagination or the fact that I implemented many change requests regarding permissions already -- before we merged the permissions branch

config/initializers/admin_resource_tabs.rb
aviav: This should be the only thing still missing for this PR

tf:
see #849

PR #851, opened 2017-08-31
Fix file type detection for vtt uploads on Safari

tf:
Safari gives an empty string as `type` for vtt file uploads. Test for
file extension instead. We already do this for `srt` files anyway.


tilsammans:
It's true, Safari is the new IE.

PR #857, opened 2017-09-06
Switch off file meta data edit links when uploading

aviav:


app/assets/javascripts/pageflow/editor/views/file_meta_data_item_value_view.js
```diff
@@ -45,5 +49,10 @@ pageflow.FileMetaDataItemValueView = Backbone.Marionette.ItemView.extend({
 
   getText: function() {
     throw new Error('Not implemented');
+  },
+
+  toggleEditLink: function() {
+    this.ui.editLink.toggle(!!this.options.settingsDialogTabLink &&
+                            !this.model.isNew());
   }
```
tf: We should also call `toggleEditLink` from `onRender` to make sure the state is set up correctly initially.

PR #860, opened 2017-09-07
split up base JS into vendor/app

tilsammans:
This pattern is quite common for Webpack and it makes a lot of sense to me too.

We split up application.js into a vendor part and an app part. The engine precompiles both. They are meant to be included separately into the html <head>. Vendor won't change nearly as often and so will remain cached for longer.

It's backwards compatible with the current way. I'll update docs etc if you think this is a good idea.

### DIFF
--- a/app/assets/javascripts/pageflow/vendor.js
+++ b/app/assets/javascripts/pageflow/vendor.js
@@ -1,0 +1,16 @@
+// vendor JavaScript manifest for Pageflow.
+// This file is meant to hold code we don't develop and/or rarely change.
+// It means this asset will stay cached for much longer.
+
+//= require i18n
+//= require jquery
+//= require jquery-ui/widget
+//= require jquery_ujs
+//= require jquery.fullscreen
+//= require jquery.placeholder
+
+//= require iscroll
+//= require audio5.min
+
+//= require react
+//= require backbone-rails
--- a/lib/pageflow/engine.rb
+++ b/lib/pageflow/engine.rb
@@ -57,25 +57,6 @@ module Pageflow
       config.i18n.reload!
     end
 
-    # Precompile additional assets. pageflow/editor.* has to be
-    # provided by the main app.
-    config.assets.precompile += %w(pageflow/editor.js pageflow/editor.css
-                                   pageflow/application_with_simulated_media_queries.css
-                                   pageflow/print_view.css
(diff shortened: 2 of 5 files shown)


tilsammans:
The way HTTP/2 streams all assets together makes the enormous asset bundles obsolete IMO.

tilsammans:
@tf changes done & rebased

lib/pageflow/engine.rb
```diff
@@ -88,5 +69,27 @@ class Engine < ::Rails::Engine
...
+         Pageflow.config.themes.any? do |theme|
+           path == theme.stylesheet_path
+         end
+       end
+
+       app.config.assets.precompile << lambda do |path, filename|
+         filename.start_with?(Engine.root.join('app/assets').to_s) &&
+           !['.js', '.css', ''].include?(File.extname(path))
+       end
+    end
```
tf: What difference does wrapping the code in an `initializer` block make here? Can we fix indentation here?

app/assets/javascripts/pageflow/vendor.js
```diff
@@ -0,0 +1,16 @@
+// vendor stylesheet for Pageflow.
```
tf: "stylesheet" is a typo.

lib/pageflow/engine.rb
tilsammans: Just following the documented API for engine asset precompilation. Pretty sure this will handle a tweaked asset pipeline better.

Indentation fixed.

tilsammans:
@tf 

> Specs are failing since Teaspoon does not load the new vendor.js file.

I am not too familiar with Teaspoon, how do I fix that?

Other changes are done.

tf:
Tried to push a commit to the PR, but maybe the option is not set. Here's the diff, that makes teaspoon pass: https://github.com/scrollytelling/pageflow/compare/vendor-app-javascript...tf:vendor-app-javascript

tilsammans:
I've merged in your commits.

tf:
Travis appears to be stuck somehow...

tilsammans:
@tf squashed all the commits (including yours).

tf:
Thanks. I'd totally merge, but I want the green check mark on the PR first ;) I don't know why Travis stopped picking up changes, but worst case we'll have to create a fresh PR from the same branch.

tilsammans:
I'll create a new PR tomorrow if it doesn't go through today.

PR #861, opened 2017-09-07
Add Rake task and Resque job to delete old auto snapshots

tf:
Prevent database from growing too much.


tilsammans:
What's the deal with this Rake task @tf ?

We run it once and it's all set? Or run it every day?

aviav:
The task can be run as needed, periodically or once. Running the task once will delete auto snapshots only once. There is no built-in auto-repeat of this

PR #868, opened 2017-09-12
If multiaccounts off, users cannot gain/lose account memberships

aviav:
Only by invitation or registration is it possible to determine a
user's account when allow_multiaccount_users = false

### DIFF
--- a/app/policies/pageflow/account_policy.rb
+++ b/app/policies/pageflow/account_policy.rb
@@ -99,24 +99,20 @@ module Pageflow
       publish?
     end
 
-    def manage?
+    def read?
       user.admin? ||
         (query.has_at_least_role?(:manager) &&
          Pageflow.config.allow_multiaccount_users)
     end
 
-    def read?
-      manage?
-    end
-
     def update?
-      manage?
+      read?
     end
 
     def update_feature_configuration_on?
       user.admin? ||
         (!permissions_config.only_admins_may_update_features &&
-         manage?)
+         read?)
     end
 
     def add_member_to?
@@ -147,12 +143,6 @@ module Pageflow
       admin? ||
         (Pageflow.config.allow_multiaccount_users &&
          @user.memberships.on_accounts.as_manager.any?)
-    end
-
-    private
-
-    def allows?(roles)
-      user.memberships.where(role: roles, entity: @account).any?
     end
   end
 end
--- a/app/policies/pageflow/account_policy.rb
+++ b/app/policies/pageflow/account_policy.rb
@@ -120,7 +120,9 @@ module Pageflow
     end
 
     def add_member_to?
-      manage?
+      Pageflow.config.allow_multiaccount_users &&
+        (user.admin? ||
+         query.has_at_least_role?(:manager))
     end
 
     def edit_role_on?
@@ -128,7 +130,7 @@ module Pageflow
     end
 
     def destroy_membership_on?
-      edit_role_on?
+      add_member_to?
     end
 
     def admin?
--- a/app/policies/pageflow/membership_policy.rb
+++ b/app/policies/pageflow/membership_policy.rb
@@ -98,7 +98,7 @@ module Pageflow
     end
 
     def destroy_for_account?
-      @user.admin? ||
+      (Pageflow.config.allow_multiaccount_users && @user.admin?) ||
         AccountPolicy.new(@user, @membership.entity).destroy_membership_on?
     end
   end
--- a/app/policies/pageflow/membership_policy.rb
+++ b/app/policies/pageflow/membership_policy.rb
@@ -98,8 +98,7 @@ module Pageflow
     end
 
     def destroy_for_account?
-      (Pageflow.config.allow_multiaccount_users && @user.admin?) ||
-        AccountPolicy.new(@user, @membership.entity).destroy_membership_on?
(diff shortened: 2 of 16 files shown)


app/policies/pageflow/membership_policy.rb
```diff
@@ -98,7 +98,7 @@ def destroy_for_entry?
     end
 
     def destroy_for_account?
-      @user.admin? ||
+      (Pageflow.config.allow_multiaccount_users && @user.admin?) ||
         AccountPolicy.new(@user, @membership.entity).destroy_membership_on?
```
tf: The first part of the conditional appears to be redundant. `add_membership_on?` (which `destroy_membership_on?` delegates to) will always be true anyway if the first part of the conditional is true.

app/views/components/pageflow/admin/add_membership_button_if_needed.rb
```diff
@@ -4,39 +4,45 @@ class AddMembershipButtonIfNeeded < ViewComponent
...
-          path = new_admin_user_membership_path(
-            parent, entity_type: to_class_name(entity_type)
-          )
-          data_tooltip = I18n.t("pageflow.admin.#{entity_type}.none_addable_tooltip")
-          rel = "add_#{entity_type}_membership"
-        elsif parent.is_a?(Entry)
-          path = new_admin_entry_membership_path(
-            parent, entity_type: to_class_name(entity_type)
-          )
+        if !Pageflow.config.allow_multiaccount_users &&
```
tf: This sort of duplicated permission logic in the view layer. We could try to use `authorize?(:create, membership)` with an unsaved example membership instead. The logic to construct that example membership would be closely related to what is already used below to find out if another membership would be possible. Then we could also remove the `authorize?` call in the  [members tab](https://github.com/codevise/pageflow/blob/b6610062f7a37bcdd0463cfef2f5159c6a0ae219/app/views/components/pageflow/admin/members_tab.rb#L49). 

Maybe take stab at moving in that direction. But if it turns out to take more than one or two hours, I'm not sure it's worth it.

aviav: As discussed, I'll go in the direction of putting another `authorize?` call into UserAccountsTab and rename `AddMembershipButtonIfNeeded` to `AddMembershipButton`

app/policies/pageflow/account_policy.rb
```diff
@@ -120,15 +120,17 @@ def update_feature_configuration_on?
     end
 
     def add_member_to?
-      manage?
```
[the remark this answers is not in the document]
aviav: I'll point the other two methods to `read?` then, and remove `manage?`. I agree that it is an artifact of the transition from CanCan

app/views/components/pageflow/admin/add_membership_button_if_needed.rb
tf: And `UsersTab` which is displayed in the account admin show page.

app/policies/pageflow/membership_policy.rb
aviav: It can still make sense from a perspective of asking questions that are inexpensive to answer early on. However, I agree that it makes more sense to settle on one layer where we answer this, as long as we have no performance issue here

app/policies/pageflow/account_policy.rb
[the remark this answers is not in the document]
tf: ok

app/policies/pageflow/membership_policy.rb
tf: But `add_membership_on?` will short circuit under the exact same conditions. So there really is no performance win here.

aviav: I agree. When I wrote this, it was more challenging to me than it is now to keep different layers of code in mind at the same time. In the meantime, learning happened

aviav:
FTR, the cause of what took me so long was not reading thoroughly on first interactive rebase, and then searching in the wrong places, because I did not notice that my `git rerere` settings overwrote what I thought I had fixed

app/views/components/pageflow/admin/add_membership_button_if_needed.rb
aviav: I only saw this last comment now while reading my e-mail. Looking into it

app/views/components/pageflow/admin/users_tab.rb
```diff
@@ -40,7 +40,9 @@ def build(theming)
               end
             end
           end
-          add_membership_button_if_needed(current_user, account, 'account')
+          if authorized?(:add_account_to, :users)
+            add_membership_button(current_user, account, 'account')
+          end
```
tf: Shouldn't this check `:add_member_to, account`? In the remaining `:add_account_to` case (in the `UserAccountsTab`), we can then pass the `user` instead of the symbol `:users`. Even if the `UserPolicy` does not rely on the concrete user at the moment, I see no reason to not allow this for the future.

aviav: For now, I'll leave the symbol, since we still have `can :manage, ::User` in the `ability_mixin`, which covers :add_account_to for sysadmins. I agree with `:add_member_to, account`

tf: ok

PR #871, opened 2017-09-13
add Resque::Server to install generator

### LINKED ISSUE #856: Resque Web
What do you think about adding Resque Web to the install generator, @tf?

It's three very helpful lines:

``` ruby
  authenticate :user, lambda {|user| user.admin?} do
    mount Resque::Server.new, at: "/background_jobs"
  end
```

If you don't like it, I will add it to the Wiki.

tilsammans:
This will mount Resque::Web in the routes at /background_jobs.
Fixes #856

### DIFF
--- a/lib/generators/pageflow/routes/routes_generator.rb
+++ b/lib/generators/pageflow/routes/routes_generator.rb
@@ -7,8 +7,18 @@ module Pageflow
 
       def add_route
         inject_into_file 'config/routes.rb', after: "  ActiveAdmin.routes(self)\n" do
-          "  Pageflow.routes(self)\n"
+          <<-HEREDOC
+  Pageflow.routes(self)
+
+  authenticate :user, lambda { |user| user.admin? } do
+    mount Resque::Server.new, at: "/background_jobs"
+  end
+          HEREDOC
         end
+      end
+
+      def require_resque_server
+        prepend_to_file 'config/routes.rb', "require 'resque/server'\nrequire 'resque_scheduler/server'\n\n"
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 - Generate a secure password in the seeds
   ([#775](https://github.com/codevise/pageflow/pull/775))
 
+- Add Resque::Server to the generated routes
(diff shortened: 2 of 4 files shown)


lib/generators/pageflow/routes/routes_generator.rb
```diff
@@ -7,9 +7,19 @@ class RoutesGenerator < Rails::Generators::Base
...
+
+  authenticate :user, lambda {|user| user.admin?} do
+    mount Resque::Server.new, at: "/background_jobs"
+  end
+          HEREDOC
         end
       end
+
+      def require_resque_server
+        prepend_to_file 'config/routes.rb', "require 'resque/server'\n\n"
```
tf: Let's also require `resque_scheduler/server`. It adds another tab to the Resque web interface.

lib/generators/pageflow/routes/routes_generator.rb
```diff
@@ -7,9 +7,19 @@ class RoutesGenerator < Rails::Generators::Base
 
       def add_route
         inject_into_file 'config/routes.rb', after: "  ActiveAdmin.routes(self)\n" do
-          "  Pageflow.routes(self)\n"
+          <<-HEREDOC
+  Pageflow.routes(self)
+
+  authenticate :user, lambda {|user| user.admin?} do
```
tf: Where is Hound? :wink: Let's add spaces inside the curly brackets here.

tilsammans:
TIL about `rescue/scheduler/server`!

All changes done.

tilsammans:
@tf I am going to back out the resque/scheduler require.

It doesn't load in our app: `LoadError: cannot load such file -- resque/scheduler/server`

There are various issues in resque/scheduler that make it seem that this is not possible in the version tied to Pageflow. And lots of people pointing out that the bundled web server in Resque 1.x will be gone in 2.x anyway.

tf:
I think it has to be `resque_scheduler/server` instead of `resque/scheduler/server`. We use it in all our production apps. Resque 2 has been abandoned and will not happen AFAIK.

tilsammans:
Got it. PR updated.

We're Sidekiq people around here...

PR #880, opened 2017-09-28
Reduce editor lock polling interval

rafaelteixeira:


spec/helpers/pageflow/config_helper_spec.rb
```diff
@@ -38,6 +38,12 @@ module Pageflow
 
         expect(result['availablePublicLocales']).to eq(['fr'])
       end
+
+      it 'includes edit lock interval' do
+        result = JSON.parse(helper.editor_config_seeds)
+
+        expect(result['editorLockInterval']).to eq(Pageflow.config.editor_lock_interval)
```
tf: On second thought, I'm not sure how Rails' `15.seconds` actually transform into JavaScript numerals. So just to be sure everything works as expected, I'd suggest assigning the config option at the top of the test to, say, `5.seconds` and then assert that the rendered value actually is `5`. Since the default unit of interval duration in JS is milliseconds, I'd even consider calling the the JS seed property `editLockIntervalInSeconds`. That will also make the `* 1000` above much clearer.

lib/pageflow/configuration.rb
```diff
@@ -266,6 +266,11 @@ class Configuration
     # @since edge
     attr_reader :permissions
 
+    # Defines the editor lock polling interval.
+    # @return [number]
+    # @since 0.12
+    attr_reader :editor_lock_interval
```
tf: It'c called "edit lock" in the rest of the code. Let's stick with that. Can you still fix the `@since` tag?

app/models/pageflow/edit_lock.rb
```diff
@@ -33,7 +33,7 @@ def matches?(id)
     end
 
     def timed_out?
-      Time.now > self.updated_at + TIME_TO_LIVE
+      Time.now > updated_at + Pageflow.config.editor_lock_interval
```
tf: `timed_out?` should be consistent with `active` above.

app/models/pageflow/edit_lock.rb
```diff
@@ -1,8 +1,8 @@
 module Pageflow
   class EditLock < ActiveRecord::Base
-    TIME_TO_LIVE = 10.seconds
-
-    scope :active, ->(time = Time.now - TIME_TO_LIVE) { where('pageflow_edit_locks.updated_at >= ?', time) }
+    scope :active, ->(time = Time.now - 2.3*Pageflow.config.editor_lock_interval) {
```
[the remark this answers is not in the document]
tf: Maybe we can turn `2.3` into a constant and give it an intention revealing name? Not sure what that could be though.

lib/pageflow/configuration.rb
```diff
@@ -266,6 +266,11 @@ class Configuration
     # @since edge
     attr_reader :permissions
 
+    # Defines the editor lock polling interval.
+    # @return [number]
+    # @since edge
+    attr_accessor :edit_lock
+
```
tf: Sorry for being so nitpicky, but when I suggested to rename `editor_lock_interval` and use `edit_lock` instead, I did not mean to drop the `interval` suffix. I'd even go for a more explicit naming like `edit_lock_polling_interval`.

spec/models/pageflow/edit_lock_spec.rb
```diff
@@ -4,7 +4,8 @@ module Pageflow
   describe EditLock do
     describe 'active' do
       it 'excludes timed out EditLocks' do
-        lock = create(:edit_lock, updated_at: (EditLock::TIME_TO_LIVE + 1.minute).ago)
+        time_lock = Pageflow.config.edit_lock * EditLock::TIMER_TOLERANCE
+        lock = create(:edit_lock, updated_at: (Time.now - time_lock - 1.minute))
```
tf: The piece of logic `Pageflow.config.edit_lock * EditLock::TIMER_TOLERANCE` is now duplicated a lot. What do you think about extracting a static `EditLock.time_to_live` method?

lib/pageflow/configuration.rb
rafaelteixeira: Oh sure, I'll changed it right now

spec/models/pageflow/edit_lock_spec.rb
rafaelteixeira: Yeah, that looks better.

PR #885, opened 2017-10-21
Split vendored code from our own

tilsammans:
These will be compiled separately. Since they change less often,
the idea is that these assets will remain in cache for longer.

Continuation of #860 where Travis refused to play nice.


tilsammans:
Travis still didn't receive the webhook. Amended and force-pushed the last commit and then it worked.

PR #889, opened 2017-10-30
Introduce ApplicationRecord

tilsammans:
All models inherit from it instead of `ActiveRecord::Base`.
This is the default in Rails too and it provides a standard approach to
put things applicable to all models.

### DIFF
--- a/app/models/pageflow/account.rb
+++ b/app/models/pageflow/account.rb
@@ -1,5 +1,5 @@
 module Pageflow
-  class Account < ActiveRecord::Base
+  class Account < ApplicationRecord
     include FeatureTarget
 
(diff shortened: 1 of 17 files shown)


tilsammans:
In the near future I am going to need `ApplicationRecord`, because I want to make `Pageflow::VERSION` part of the cache key. (When Pageflow updates I think we should purge the cache just to be sure)

tilsammans:
@tf all green

tf:
Our deploy scripts actually purge the whole cache on each deploy to be safe. Including the version in the cache key might be a nice safety net for people only ever using released versions. For cases where we are running off of feature branches, we'd still need explicit cache cleaning.

PR #891, opened 2017-10-30
update contents of gemspec

tilsammans:
- add MIT license
- add lib as the standard require path
- don't put lib on the load path if it's already there
- make the bundler and rake dependencies explicit

### DIFF
--- a/pageflow.gemspec
+++ b/pageflow.gemspec
@@ -12,6 +12,7 @@ Gem::Specification.new do |s|
   s.license     = 'MIT'
 
   s.files = Dir['{admins,app,config,db,lib,vendor,spec/factories,spec/fixtures}/**/*', 'MIT-LICENSE', 'Rakefile', 'README.md', 'CHANGELOG.md']
+
   s.require_paths = ['lib']
 
   s.add_dependency 'rails', '~> 4.2.6'
@@ -119,9 +120,6 @@ Gem::Specification.new do |s|
 
   # Used by the dummy rails application
   s.add_development_dependency 'mysql2', '~> 0.3.16'
-
-  s.add_development_dependency 'bundler'
-  s.add_development_dependency 'rake'
 
   # Testing framework
   s.add_development_dependency 'rspec-rails', '~> 3.4'
--- a/pageflow.gemspec
+++ b/pageflow.gemspec
@@ -1,5 +1,5 @@
-$:.push File.expand_path('../lib', __FILE__)
-
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'pageflow/version'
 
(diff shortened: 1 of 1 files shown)


pageflow.gemspec
```diff
@@ -9,8 +9,10 @@ Gem::Specification.new do |s|
   s.email       = ['info@codevise.de']
   s.homepage    = 'http://www.pageflow.io'
   s.summary     = 'Multimedia story telling for the web.'
+  s.license     = 'MIT'
 
   s.files = Dir['{admins,app,config,db,lib,vendor,spec/factories,spec/fixtures}/**/*', 'MIT-LICENSE', 'Rakefile', 'README.md', 'CHANGELOG.md']
+  s.require_paths = ['lib']
```
tf: This appears to [be the default](http://guides.rubygems.org/specification-reference/#require_paths=). But being explicit can't hurt.

pageflow.gemspec
```diff
@@ -118,6 +120,9 @@ Gem::Specification.new do |s|
   # Used by the dummy rails application
   s.add_development_dependency 'mysql2', '~> 0.3.16'
 
+  s.add_development_dependency 'bundler'
+  s.add_development_dependency 'rake'
+
```
tf: The rest of the gemspec goes with pessimistic version constraints. Should we state that we don't know yet whether we are going to be compatible with Bunder 2 or Rake 13?

pageflow.gemspec
tilsammans: I agree. I tend to include it nowadays, so developers are exposed to it. It can explain some weird gem behaviour for example.

pageflow.gemspec
tilsammans: I wasn't sure. I am not aware of their roadmaps and don't want to get into it either.

Restricting these pessimistically felt like busywork in the future, but I guess consistency wins here. For what it's worth, Rails 4 needed bundler < 2 but Rails 5 wants 1.3.0 and higher.

Maybe we'll roll with `rake ~> 11.x` and `bundler >= 1.3.0`?

tf: That would exclude the current Rake version 12. Since there are no known (to me) Rake or Bundler versions that do not work with Pageflow (not the Rails version used by Pageflow) `rake < 13` and `bundler < 2` would probably be the most consistent pessimistic choices. But I'm also open to using open ranges given that both are development dependencies and core tools that should be pretty stable going forward.

tilsammans: I've been wrestling with this and I came to think that I should just remove them from `pageflow.gemspec`. Rails is a dependency and Rails will do a fine job of pulling in Rake and Bundler.

tf: Fine with me. We also use a lot of other stuff - even at runtime, not just for development - that comes with Rails (e.g. ActiveRecord) without explicitly declaring dependencies.

PR #892, opened 2017-11-02
[WIP] Add HowlerJs lib to autoplay muted videos

rafaelteixeira:
Use the HowlerJs lib to allow autoplay background videos to play sound on non-native players (videos muted by default)

- Replace Audio5js with ProxyPlayer that wraps HowlerJs and keeps the audio player interface stable
- Patch HowlerJs to emit timeupdate HTML5 event 
- Patch HowlerJs to have currentTime method that works on HTML5 mode
- Manually trigger timeupdate events via interval on webAudio mode
- Provide currentTime method for webAudio mode
- Add method to delegate volume fading from proxy player to HowlerJs
- Check if there's already a volume fading method and do nothing if it has
- Remove disabling of ATMO on mobile
- Remove volume control checks from ATMO
- To add an option to video player to use Howler to play the sound of videos
- Use this option on background videos
- Delegate all volume methods to HowlerJs in videos that use it
- Add global Unmute button UI to call HowlerJs unlock method

Additional changes:

- Display a mute button only on mobiles
- Create a global **muted** flag that defaults to **true** on mobile (*only muted autoplay as before*)
- Initialize ATMOs and autoplay videos *muted* according to this global muted flag
- Create a global **unmute** event triggered by clicking the unmute button
- When the unmute event is triggered:
  - Removes the unmute button
  - Sets the muted flag to false
  - Unmute all Howler instances
  - Unmute listener on Howler video player triggers play method to sync audio with the video currentTime


tilsammans:
I heard through the grapevine that this is being worked on. Is there an ETA? We need this baaaaaaadly! 🙏

tf:
I agree. @rafaelteixeira will be back next week. So let's see that we can sort it out then. Far as I know, this PR does not represent the latest version of the changes made so far.

tf:
Regarding first user interaction: We discussed adding a global, one-time unmute button (pretty much like here [1] - which is also where we got the idea for using Howler).

Catching the promise might still be a good idea. Maybe even to re-display the button?

[1] https://www.nytimes.com/interactive/2016/08/05/sports/olympics-gymnast-simone-biles.html

tilsammans:
Yes! That would be excellent. It’s what people have come to expect also. I assume we would be free to style the button using our own themes?

tilsammans:
cc @studiorabota

tilsammans:
Pinging @studiorabota this is about to be merged, any opinions?

.jshintrc
```diff
@@ -14,12 +14,13 @@
     "Cocktail",
     "_",
     "videojs",
-    "Audio5js",
+    "Howl",
+    "Howler",
```
tf: Could not find any usages of `Howler` global. Only `Howl`.

app/assets/javascripts/pageflow/audio_player/proxy_player.js
```diff
@@ -0,0 +1,133 @@
+pageflow.AudioPlayer.proxyPlayer = function(sources, options) {
+  var proxy = {
+    position: 0,
+    duration: 0
+  };
+
+  var _playing = false;
```
tf: No need for `_` in name. This local variable is private anyway.

app/assets/javascripts/pageflow/audio_player/proxy_player.js
```diff
@@ -0,0 +1,133 @@
...
+    }
+    return result;
+  };
+
+  proxy.fadeVolume = function(newVolume, duration) {
+    var currentVolume = this.volume();
+
+    this.fade(currentVolume, newVolume, duration);
+
+    return new $.Deferred().resolve().promise();
```
tf: Promise should only be resolved after fading is done. Howler emits an event that can be used to do this.

app/assets/javascripts/pageflow/slideshow/atmo.js
```diff
@@ -96,7 +86,7 @@
         crossFade: true,
         playFromBeginning: false,
         rewindOnChange: true,
-        pauseInBackground: true
+        pauseInBackground: false
```
tf: This is intended to stop audio playback when mobile browser is sent to background. Is this no longer needed?

app/assets/javascripts/pageflow/video_player/use_howler_for_audio_output.js
```diff
@@ -0,0 +1,36 @@
...
+    return originalPause.apply(this, arguments);
+  };
+
+  player.fadeVolume = function(newVolume, duration) {
+    var currentVolume = playerAudio.volume();
+
+    playerAudio.fade(currentVolume, newVolume, duration);
+
+    return new $.Deferred().resolve().promise();
+  };
```
tf: Calls to `volume` need to be intercepted as well.

node_package/src/media/components/Page.jsx
```diff
@@ -70,15 +70,9 @@ export default connect(combineSelectors({
...
     defaultTextTrackFileId: prop('page.defaultTextTrackFileId')
-  }),
-  hasAutoplaySupport: has('autoplay support')
+  })
 }))(MediaPage);
 
-function willAutoplay(props) {
-  return props.page.autoplay !== false &&
-         props.hasAutoplaySupport;
-}
```
tf: The old logic enabled autoplay also when `page.autoplay` is undefined.

app/assets/javascripts/pageflow/audio_player/proxy_player.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+
+  var timeupdateInterval;
+  function startInterval() {
+    if (!timeupdateInterval) {
+      if (player._html5) {
+        timeupdateInterval = setInterval(function() {
+          player._emit('timeupdate');
+        }, 250);
```
tf: Doesn't the patched version of Howler emit `timeupdate` events by itself in Html 5 mode? There is code further up in this file handling such an event at least.

.jshintrc
rafaelteixeira: It was for the global unlock (Howler._enableMobileAudio()).
But if not necessary, I'll remove.

app/assets/javascripts/pageflow/audio_player/proxy_player.js
rafaelteixeira: You're right. Removing the HTML5 emitter in the next commit.

node_package/src/media/components/Page.jsx
rafaelteixeira: Old logic restored

rafaelteixeira:
There's still some strange behavior on iOS.
The auto unlock from Howler doesn't seen to work, and it didn't raise any error in the onloaderror and onplayerror events.
It needs further investigation for this case.

tilsammans:
> There's still some strange behavior on iOS.
> The auto unlock from Howler doesn't seen to work, and it didn't raise any error in the onloaderror and onplayerror events.

I think I understand why this does not work on iOS, because browser vendors _really_ don't want this to happen. So any shenanigans will be dealt with. I am already completely happy if we can interact with this Promise that is supposed to come from the browser.

I don't mind surfacing a button with a click event. It beats the alternative: that nothing happens and no explanation is given. But what you're saying is that the promise doesn't come either? 😢 😭

rafaelteixeira:
@tilsammans I added a mute button widget to test this, but am still having some weird behaviours. The unmute action seems to unlock the audios playback, but the ATMO feature still require an touch event to play OR that I access a background video page (on my test case).

app/assets/javascripts/pageflow/built_in_widget_types.js
```diff
@@ -8,4 +8,10 @@ pageflow.widgetTypes.register('default_mobile_navigation', {
   enhance: function(element) {
     element.navigationMobile();
   }
+});
+
+pageflow.widgetTypes.register('default_mute_toogle', {
+  enhance: function(element) {
+    element.muteToggle();
+  }
 });
```
tf: The new widget is just a jQuery UI widget, not a Pageflow widget type that can be activated or deactivated by the user. This code is not needed.

app/assets/stylesheets/pageflow/base.scss
```diff
@@ -66,6 +66,7 @@ a {
 @import "./page_types";
 @import "./page_transitions";
 @import "./slideshow";
+@import "./mute_toggle";
```
tf: Styles should live in default theme to make them configurable.

lib/pageflow/built_in_widget_type.rb
```diff
@@ -24,6 +24,10 @@ def self.mobile_navigation
       new('default_mobile_navigation', ['mobile_navigation'], 'pageflow/entries/mobile_navigation')
     end
 
+    def self.mute_toggle
+      new('default_mute_toggle', ['player_controls'], 'pageflow/widgets/mute_toggle')
+    end
+
```
tf: Not needed.

lib/pageflow/built_in_widget_types_plugin.rb
```diff
@@ -3,6 +3,7 @@ class BuiltInWidgetTypesPlugin < Plugin
     def configure(config)
       config.widget_types.register(Pageflow::BuiltInWidgetType.navigation, default: true)
       config.widget_types.register(Pageflow::BuiltInWidgetType.mobile_navigation, default: true)
+      config.widget_types.register(Pageflow::BuiltInWidgetType.mute_toggle, default: true)
```
tf: Not needed.

app/assets/javascripts/pageflow/widgets/mute_toggle.js
```diff
@@ -0,0 +1,52 @@
+(function($) {
+  $.widget('pageflow.muteToggle', {
+    _create: function() {
+      var audio = new Audio();
+      audio.setAttribute('src', '/assets/pageflow/unmute.mp3');
```
tf: Won't work with precompiled assets.

tf:
> Initialize ATMOs and autoplay videos muted according to this global muted flag

In particular, we decided to deactivate the atmo initially and only activate it once the unmute event is dispatched. This way it should be possible to start playing the atmo immediately once the button is clicked.

tf:
Continued in #949

PR #894, opened 2017-11-03
Work around breaking change in resque_mailer 2.4.3

tf:
Stating with [1] a `HashWithIndifferentAccess` is passed which does
not support the `symbolize_keys!` method which we used before to
ensure compatbility with `resque_mailer` 2.4.1.

[1] https://github.com/zapnap/resque_mailer/pull/102

### DIFF
--- a/app/mailers/pageflow/user_mailer.rb
+++ b/app/mailers/pageflow/user_mailer.rb
@@ -3,7 +3,17 @@ module Pageflow
     include Resque::Mailer
 
     def invitation(options)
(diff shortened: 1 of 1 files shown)


tf:
Well, there is a spec that tests the expected "symbols as keys" case. We could add tests for the "string keys" and HashWithIndifferentAccess cases as well, but that still won't do as any good once `resque_mailer` finds something even weirder to pass as arguments - which will probably happen soon as 2.4.4 is out given their current track record :smiling_imp:

PR #895, opened 2017-11-03
[WIP] Paperclip initializer improvements

tilsammans:
I took the time to clean part of this initializer up. I want to leave it at this; every time I dove deeper I ended up in confusing mess. I took the liberty of yanking out some IMHO dubious distinctions based on RAILS_ENV. I don't think they served a useful function. Less is more in cases like this.

The highlights:

* All `RAILS_ENV==test` code was taken out and moved to spec/support
* config.paperclip_attachments_version default was moved into Configuration itself
* config.paperclip_filesystem_root was moved there as well
* the :host interpolation was inverted and now special-cases development only
* I tidied up the initializer generator and given more emphasis on this setting for multi server
* added very simple smoke tests for all Configuration accessors

The individual commits are scoped to one specific change.


tilsammans:
I created a Wiki entry for the importance of the shared file system: https://github.com/codevise/pageflow/wiki/Configuring-a-shared-folder-for-file-uploads

We have this running live now and everything's cool. I still think direct upload to S3 should be our medium-term goal, as this has the added benefit of sharing nothing (reducing custom config) and freeing up the local web server.

tf:
see #985

PR #902, opened 2017-11-09
Update setting_up_s3_bucket_policies.md

paulmwatson:
The `"Version": "2017-06-20",` is invalid according to the S3 bucket policy editor (error of `The policy must contain a valid version string`.) The policy doesn't need it.

### DIFF
--- a/doc/setting_up_s3_bucket_policies.md
+++ b/doc/setting_up_s3_bucket_policies.md
@@ -50,7 +50,6 @@ Grant read access to main bucket:
 Grant full access to output bucket:
 
     {
-      "Version": "2017-06-20",
       "Id": "PageflowOutputBucketPolicy",
       "Statement": [
         {


tf:
Thanks for pointing this out.

[The AWS docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html) say: 

> If you do not include a Version element, the value defaults to 2008-10-17. However, it is a good practice to always include a Version element and set it to 2012-10-17.

Then again the [Zencoder docs](https://app.zencoder.com/docs/guides/getting-started/working-with-s3) use `"Version": "2008-10-17"`. So not completely sure what to do here.

At any rate, the version string in our guide appears to be bogus.

tf:
Meh. I guess "not good practice" is better than "wrong" :joy:

CleverMobile2811:
Hello,

Thanks for your pageflow source.
I've set up in my side.
- AWS S3 bucket upload failed.
As a result, GET http://page-flow-out.s3.website-us-west-1.amazonaws.com/background_small.png : 
Not Found
I'd added config in AWS S3 bucket and pageflow.rb.
I will appreciate if anyone give me any advise for aws s3 setting for pageflow.
Thank you!

PR #903, opened 2017-11-09
[WIP] Permit custom params for file types that request this

aviav:
Also, file types can now define custom attributes. Custom attributes
serve a different purpose than configuration attributes: Their values
do not change after a file is created. Thus, we can persist them
directly on the file as opposed to on file usages, since constant
attributes need no versioning.


tf:
merged as part of #1024

PR #905, opened 2017-11-12
[WIP] Atom feed

### LINKED ISSUE #585: Atom feed
What would you think about an atom feed?

`GET /entries.atom` would contain the 20 latest published revisions/entries.

tilsammans:
Just one test, but made some awesome progress here. I'm happy with the results.

Still WIP but would love to get your feedback at this point.

Fixes #585


app/views/pageflow/entries/index.atom.builder
```diff
@@ -0,0 +1,32 @@
...
+  feed.title @account.name
+  feed.url pretty_theming_url(@theming)
+  feed.updated @entries.first.updated_at if @entries.present?
+  feed.favicon image_path("pageflow/themes/#{@theming.theme.directory_name}/favicon.ico")
+
+  # we can have next_url once we can upgrade ActiveAdmin->Kaminari
+  # feed.next_url pretty_theming_url(path_to_next_page(@entries))
+
+  @entries.each do |pageflow_entry|
+    published_entry = Pageflow::PublishedEntry.new(pageflow_entry)
```
tf: I'd suggest moving all of the entry query building logic that lives in the controller into a separate query class (something like `FeedEntriesQuery`). This gives us a nice place to test that only the desired entries show up in the feed. We can then also wrap all of the entries in `PublishedEntriy` decorators right there to further simplify this view code.

lib/generators/pageflow/install/install_generator.rb
```diff
@@ -8,6 +8,7 @@ def generate_all
         generate 'active_admin:install User --skip-users'
         generate 'active_admin:devise User --no-default-user'
         generate 'friendly_id'
+        generate 'kaminari:config'
```
tf: What exactly can be configured in the generated file? I'm a bit worried, that changes made in a host application to change Active Admin behavior might unintentionally also change the feed output. Can we somehow decouple from these defaults by making explicit choices in Pageflow (and provide Pageflow config options where necessary)? I'd prefer if Kaminari did not become part of our API surface.

tilsammans: I believe it's just view helpers and defaults, and we can safely omit this step.

However with this change we _are_ pulling Kaminari on board, and the kaminari generator is something people will be able to do if they want. Luckily we don't use paging in any html view and we can sort of pretend that Kaminari isn't really here.

spec/features/visitor/atom_feed_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    <image>pageflow/placeholder_original.jpg</image>
+    <author>
+      <name>Scrollytelling</name>
+    </author>
+  </entry>
+</feed>
```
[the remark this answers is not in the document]
tilsammans: Atom has been cast in concrete for years, I don't expect changes to the spec.

That being said, it doesn't really belong here. But I found myself opening a random example feed every single time, and preferred having the actual output nearby, so I could 'see' what the code is doing.

I don't mind removing it; say the word and it's gone. Until then I find it handy, is all.

tilsammans:
@tf all feedback implemented

spec/features/visitor/atom_feed_spec.rb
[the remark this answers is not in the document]
tf: It does not bother me that much really. Just stood out a bit while I was reading the PR.

app/models/pageflow/feed_entries_query.rb
```diff
@@ -0,0 +1,19 @@
+module Pageflow
+  class FeedEntriesQuery < ApplicationQuery
```
tf: On second thought, a way less technical name for this might just be `EntryFeed`. `resolve` can then be `entries`. Let's return `PublishedEntries` there directly instead of having the decorator creation inside the view code.

app/models/pageflow/feed_entries_query.rb
```diff
@@ -0,0 +1,19 @@
...
+      def initialize(scope, page)
+        @scope = scope
+        @page = page
+      end
+
+      def resolve
+        scope \
+          .published_without_password_protection
+          .order(updated_at: :desc)
+          .page page
```
tf: ActiveAdmin [jumps though some hoops](https://github.com/activeadmin/activeadmin/blob/6a0f2481761ee1387b30b9c89353faa4b4ffb359/lib/active_admin/resource_controller/data_access.rb#L247) to make sure it is compatible with all global Kaminari config choices. How far do we want to go here? At least, I'd hard code the number of items per page here.

CHANGELOG.md
```diff
@@ -9,3 +9,19 @@ None so far.
 See
 [12-1-stable branch](https://github.com/codevise/pageflow/blob/12-1-stable/CHANGELOG.md)
 for previous changes.
+
+##### Manual Update Steps
```
tf: This section is no longer needed, I guess.

tilsammans:
> The main part still missing IMO is tests.

Ouch! It's true, but truth hurts sometimes. xD

Now that the API is stable I feel easier testing everything. Will do that next.

tf:
That came across a bit more harsh than indented :smiley:

tilsammans:
It's all good. It needed to be said. Will work on it, chief!

tilsammans:
This is ready now I think @tf

spec/models/pageflow/entry_feed_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    it 'is paginated with configurable per_page' do
+      3.times { create(:entry, :published, account: account) }
+
+      scope = EntryFeed::Scope.new(account.entries)
+      scope.per_page = 2
+      entries = scope.entries
+
+      expect(entries.length).to eq(2)
+    end
```
tf: Let's test pagination on the controller level instead. Looking at the controller code, I think pagination is not happening at all at the moment. Testing a bit more full stack, we'd be on the safe side.

spec/controllers/pageflow/atom_feed_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+module Pageflow
+  describe EntriesController do
```
tf: At the moment files in `spec/controllers` are more or less one-to-one with files in `app/controllers`. I agree, though, that `entries_controller_spec.rb` is already very long. What do you think about putting this into `spec/controllers/pageflow/entries_controller/atom_feed_spec.rb`? That way, it's still clear which controller we are testing.

spec/controllers/pageflow/atom_feed_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        get(:index, format: 'atom')
+
+        expect(response.content_type).to eq('application/atom+xml')
+      end
+
+      it 'responds with published entries' do
+        get(:index, format: 'atom')
+
+        expect(response.body).to include('<title>Atom-Powered Robots Run Amok</title>')
+      end
```
tf: All of those only test the unscoped case. Let's add at least one positive and one negative test case for filtering entries by account via request scope.

app/views/pageflow/entries/show.html.erb
```diff
@@ -11,6 +11,7 @@
 
     <%= tag :link, :rel => 'icon', :href => image_path("pageflow/themes/#{@entry.theme.directory_name}/favicon.ico"), :type => 'image/ico' %>
     <%= meta_tags_for_entry(@entry) %>
+    <%= auto_discovery_link_tag(:atom, entries_url(format: 'atom')) %>
```
tf: This links to the global, unscoped entries feed, right?

app/controllers/pageflow/entries_controller.rb
```diff
@@ -19,9 +19,23 @@ class EntriesController < Pageflow::ApplicationController
...
 
-      redirect_to(theming.home_url)
+        format.atom do
+          @theming = Theming.for_request(request).first
+
+          if @theming
+            @account = @theming.account
+            @entries = EntryFeed::Scope.new(@account.entries).entries
+          else
+            @entries = EntryFeed::Scope.new(Entry).entries
```
tf: This makes me feel a bit uneasy. For example, in the case of Hosted Pageflow, this would mean that a customer suddenly gets a feed with everybody's entries soon as a he changes the CNAME of his account, but still has the old domain pointing to our server. Is there really a use case for this global feed?

app/models/pageflow/entry_feed.rb
```diff
@@ -0,0 +1,22 @@
+module Pageflow
+  class EntryFeed < ApplicationQuery
+    class Scope < Scope
+      attr_reader :scope, :page
+      attr_accessor :per_page
+
+      def initialize(scope, page=nil)
```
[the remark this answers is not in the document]
tilsammans: M'kay. Looks weird.

app/controllers/pageflow/entries_controller.rb
tilsammans: Yes, ours. 😄 

I see  your point, though. I suggest this atom-feed feature will only work on accounts that have a CNAME, agreed? I'll drop the global feed, and implement this in the host app.

app/views/pageflow/entries/show.html.erb
tilsammans: Nope, it uses the hostname of the Entry.

That being said, I should not include this tag for stories not running under a custom CNAME.

spec/controllers/pageflow/atom_feed_spec.rb
tilsammans: Sure, fine with me.

Actually, the `atom` action has nothing in common with the `html` action. It should really be a separate action or a separate controller, even. Leaving this as-is for now.

tilsammans:
Decided to drop the hard-coded page size. It felt way too restrictive.

People can change this value using Kaminari configuration. Where could we document that?

Other than that, feedback implemented.

app/controllers/pageflow/entries_controller.rb
```diff
@@ -19,14 +19,25 @@ class EntriesController < Pageflow::ApplicationController
...
+          @account = @theming.account
+          @entries = EntryFeed::Scope.new(@account.entries, params[:page]).entries
+        end
+      end
     end
 
     def show
       respond_to do |format|
         format.any(:html, :css) do
+          @theming = Theming.for_request(request).with_cname.first
```
tf: The default [theming request scope used by `Theming.for_request`](https://github.com/codevise/pageflow/blob/6e500d5c821d06764ae6b4f4d52c6d717cf86465/app/models/pageflow/cname_theming_request_scope.rb#L5) already only matches themings with a CNAME. Dropping the `with_cname` would allow using feeds with custom theming request scopes (e.g. Hosted Pageflow supports looking up themings via domains of the form `account-name.pageflow.io`).

app/models/pageflow/theming.rb
```diff
@@ -8,6 +8,7 @@ class Theming < ApplicationRecord
     has_many :entries
 
     scope :with_home_url, -> { where.not(home_url: '') }
+    scope :with_cname, -> { where.not(cname: '') }
```
tf: Maybe no longer needed.

app/views/pageflow/entries/show.html.erb
```diff
@@ -11,6 +11,7 @@
 
     <%= tag :link, :rel => 'icon', :href => image_path("pageflow/themes/#{@entry.theme.directory_name}/favicon.ico"), :type => 'image/ico' %>
     <%= meta_tags_for_entry(@entry) %>
+    <%= auto_discovery_link_tag(:atom, entries_url(format: 'atom')) if @theming %>
```
tf: This looks confusing to me. The context that `@theming` is only set if it could be determined based on the request is hard to get if you don't know the controller code. Maybe we can move this into a helper and make the `Theming.for_request` call there?

There would also be the alternative to use `Configuration#theming_url_options` the way [`pretty_entry_url`](https://github.com/codevise/pageflow/blob/6e500d5c821d06764ae6b4f4d52c6d717cf86465/app/helpers/pageflow/entries_helper.rb#L8) does. That way we could use the canonical entry url (including CNAME etc) even in cases like the preview where the request host name does not match the desired CNAME.

spec/support/config/kaminari.rb
```diff
@@ -0,0 +1,3 @@
+Kaminari.configure do |config|
+  config.default_per_page = 2
+end
```
tf: As mentioned earlier, I find depending on global Kaminari configuration problematic. We might be sharing Kaminari with other libraries in the application. If you are unhappy with hard coding, then let's add an option to `Pageflow::Configuration`. That can then also be adapted here for testing purposes without risking reconfiguring some Kaminari usages in unrelated features.

spec/features/visitor/atom_feed_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'spec_helper'
+
+feature 'account using a custom domain' do
```
tf: Only having a feature spec goes against the testing pyramid approach of the test suite. I know Rails 5 continues to blur the lines between these sorts of tests. But I think there is value in being able to run only the controller specs and already sort of know that *our* code works and having feature specs to ensure everything works together as a whole.

Controller specs are also a good place to focus a bit more on authorization concerns. While I would like the feature spec to verify the positive assumption that there is indeed a working feed feature, the controller level would be the place to test that the feeds really only contain the entries of the given account - something that is not checked by any spec right now and - in my opinion - should not be checked in a feature spec.

PR #906, opened 2017-11-14
add documentation about the maintenance job(s)

tilsammans:
See #861 for the actual change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,25 @@ The built-in Resque web server is mounted at `/background_jobs`. Use it to
 inspect the state of background jobs, and restart failed jobs. This functionality
 is only available for admins.
 
+## Maintenance
+
+As editors work with entries, the database will amass snapshots. These are
+automatically saved periodically to preventn someone losing a catastrophic
+amount of work.
+
+After a while you can safely remove these snapshots. We provide a Rake task:
+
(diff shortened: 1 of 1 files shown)


aviav:
I see the point of adding documentation for the task. I'd rather add it to the Troubleshooting section [1] of the Wiki, because needing the task is contingent on both a) an active user base and b) storage space running out, and this advice shoots trouble (see b)). 

README.md up until now seems to me like a place for less situational info. People who manage to grow an active userbase for Pageflow will at some point notice the Troubleshooting section. I'd be surprised if this happens only after they hit a storage limit for their database.

In terms of content, I approve this. If you move it to the Wiki, maybe the reference to Whenever can be left out to leave this deployment choice completely open. This would resemble our practice of not mentioning specific optional deployment choices in the documentation to decrease the future amount of stale docs. Good job in any case!

[1] https://github.com/codevise/pageflow/wiki/Troubleshooting

tilsammans:
Totally fine with the Wiki; not on board with the Troubleshooting section. To me that reads more like, "I have this error message now what do I do?". So I added a new page: https://github.com/codevise/pageflow/wiki/Maintenance

aviav:
Even better!

PR #910, opened 2017-12-02
[WIP] include the Pageflow version in all the cache keys

tilsammans:
It's pretty safe to say we want all caches purged after a version change.

When we do it like this, it's all taken care of.

### DIFF
--- a/app/models/pageflow/application_record.rb
+++ b/app/models/pageflow/application_record.rb
@@ -7,9 +7,7 @@ module Pageflow
     # A Pageflow version change warrants expiry of all caches.
     # TODO on Rails 5.2 move this code into the `version`.
     def cache_key
-      pageflow_version = Pageflow::VERSION.gsub('.', '')
-
-      super.sub(/\d+/, pageflow_version)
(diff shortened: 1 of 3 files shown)


app/models/pageflow/application_record.rb
```diff
@@ -1,5 +1,13 @@
...
   class ApplicationRecord < ActiveRecord::Base
     self.abstract_class = true
+
+    # We turn "people/5-20071224150000" into "people/1220-20071224150000"
+    # Where 1220 is the Pageflow::VERSION with periods removed.
+    # A Pageflow version change warrants expiry of all caches.
+    # TODO on Rails 5.2 move this code into the `version`.
+    def cache_key
+      super.sub(/\d+/, Pageflow::VERSION.delete('.'))
+    end
```
tf: Won't this cause two entries that were updated at the same time to have the same cache key? I think the initial number that is replaced here is the record id. Also I suggest not deleting the dots in the version string. If they are not allowed for some reason, I'd prefer replacing them with some other character. While it's surely not a problem in practice the theoretical name collision when upgrading from 12.1.11 to 12.11.1 bothers me on some level :wink:

PR #911, opened 2017-12-02
[WIP] upgrade domain/hostname to a legitimate model

tilsammans:
no more shenanigans with comma parsing using SQL.

I do need help with the admin, I guess there are various approaches how to do this, but what is the best one?

### DIFF
--- a/admins/pageflow/accounts.rb
+++ b/admins/pageflow/accounts.rb
@@ -133,8 +133,6 @@ module Pageflow
 
       def permitted_theming_attributes
         [
-          :cname,
-          :additional_cnames,
           :theme_name,
           :imprint_link_url,
           :imprint_link_label,
--- a/app/helpers/pageflow/themings_helper.rb
+++ b/app/helpers/pageflow/themings_helper.rb
@@ -1,7 +1,8 @@
 module Pageflow
   module ThemingsHelper
     DEFAULT_PUBLIC_ENTRY_OPTIONS = lambda do |theming|
-      theming.cname.present? ? {host: theming.cname} : nil
+      domain = theming.primary_domain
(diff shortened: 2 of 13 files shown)


tf:
Interesting idea. Is there a special feature that you are aiming to add here? I'm always a bit wary of large refactorings of working code that mainly intend to make things cleaner.

tilsammans:
I am hoping to reduce dependency on MySQL, which frankly, isn't a good database.

So everything with SQL code in it is a candidate for refactoring. Also, it's quite clear that hostnames are a (business) domain consideration, and should be modeled accordingly.

tf:
I'm on board with that goal. So then maybe the easiest first step might be to not change the admin UI at all? Basically moving the comma parsing from SQL to Ruby?

Going one step further could help address another long standing problem, though: At the moment, if two accounts use the same CNAME (something that still needs to be considered in the way the `primary` flag and the uniqueness validation work above) but specify different redirect URLs, there is no way to tell which one will win. This could probably be resolved by associating the redirect URL with the domain instead.

tilsammans:
> So then maybe the easiest first step might be to not change the admin UI at all? Basically moving the comma parsing from SQL to Ruby?

I don't see that path forward to be honest. I would rather bite the bullet and flesh out this new model thoroughly. We can solve your second pain point right away too. It's nothing more than a has-many relationship; pretty sure ActiveAdmin can model that nicely. I just don't have much experience with ActiveAdmin. Would someone on your side be able to do that part?

tf:
A bit hard to say when exactly we will have time to work on this, but we'll surely be able to help.

The most important first step, from my point of view, is defining how the new admin UI should work anyway. I see quite a few different options here.

PR #912, opened 2017-12-03
add missing return statement

tilsammans:
I found these changes in your uncommitted work @studiorabota

Does this need to go into Pageflow?

### DIFF
--- a/node_package/src/boot.js
+++ b/node_package/src/boot.js
@@ -76,9 +76,9 @@ export default function(pageflow) {
       }))
     );
 
-    widgetTypeRegistry.forEach(({name, component}) =>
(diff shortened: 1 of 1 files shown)


tf:
No. Neither does [`pageflow.widgetTypes.register` have a return value](https://github.com/codevise/pageflow/blob/1cac83ddd34be110dee4f13e7bef5ed32628cad6/app/assets/javascripts/pageflow/widget_types.js#L9) nor does [`forEach` regard the return value of its callback](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).

PR #914, opened 2017-12-05
[WIP] Initial spike for text embeds feature

tf:
Allow embedding React components in page text via explicit,
parameterized placeholders.

See `doc/creating_text_embeds.md` inside this PR for a rough outline of the idea.

Includes support for:
* Passing parameters to embeds
* Server side rendering
* Client side rendering
* Editor preview

To make this more user friendly, mainly additions to the editor would be needed:
* Menu in text area input to list available embeds
* Popup form to ease discovery and editing of parameters

### DIFF
--- a/app/assets/javascripts/pageflow/base.js
+++ b/app/assets/javascripts/pageflow/base.js
@@ -48,6 +48,8 @@
 //= require_tree ./widgets
 //= require ./react
 
+//= require ./test_embed
+
 pageflow = {
   log: function(text, options) {
     if (window.console && (pageflow.debugMode() || (options && options.force))) {
--- a/app/assets/javascripts/pageflow/components.js
+++ b/app/assets/javascripts/pageflow/components.js
@@ -1,5 +1,6 @@
 //= require_self
 //= require pageflow/react
+//= require ./test_embed
 
 var PAGEFLOW_EDITOR = false;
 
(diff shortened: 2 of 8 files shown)


tilsammans:
I would prefer a solution where the burden of integrating is on the developer. The {shortcode} has a little too much Wordpress-smell for my taste. ;)

Can we not create a `addFilter()` function inside [PageText.jsx](https://github.com/codevise/pageflow/blob/99143f6a81ad743a6dd1102db8a6d7918305660b/node_package/src/components/PageText.jsx) that we use as entry point for html-replacement functionality? That way, I can author a filter function in JavaScript, wire it into the component using a new configuration setting, and PageText itself would from then on use it. All it knows is, it's sending dangerous html into it, and dangerous html comes out.

Pretty much like a middleware stack I guess, but for html (text) filters.

tf:
From Pageflow's point of view this would of course be a very simple solution. For everything but trivial cases, I think, it's highly desirable to implement these text embeds as React components, though. That would not be possible with the suggested API. But even if we extended the API in some way to make this possible, I'd prefer if the hard parts of the problem could be solved by Pageflow itself:

* Parsing the DOM server side and inserting server side rendered components
* Mounting components inside the rendered DOM
* Re-mounting components on page configuration changes in the editor

Our current way of storing page contents makes these things quite tricky. Thus, I'd prefer isolating them somewhere.

Still, I also agree that the suggested syntax feels a bit too technical. At the moment I see two possibilities:

* We could think of the tag syntax as an intermediate format to delegate the hard part of the problem to Pageflow: There could still be an `addFilter` method for registering functions that insert tags of the suggested kind automatically.

* Improve editor integration such that the tags are replaced with some interactive, user friendly representation. That's the only way I see at the moment to have configurable text embeds - something that is not possible with the automagical approach.

PR #917, opened 2017-12-06
Update page type and widget type creation guides

tf:
* Describe the recommended React based approach to create page types

* Describe both the ERB and React based approach to create widget
  types

* Give an overview of the Redux related concepts in Pageflow

* Incorporate steps to create Rails engine written by @tilsammans 

This still needs a bit of proof reading, but I guess it's an okay starting point.


tf:
@tilsammans Sounds like a good idea. Maybe we can improve things like that in a separate iteration.

PR #919, opened 2017-12-20
Fix "hide text" swipe gesture for React page types

tf:
Reaching inside the page's DOM from the page widget to attach event
listeners does not work when React re-renders the page. Use event
delegation on the page's element which never changes during the page
life cycle.


tf:
Need to check if this causes a bug that we've been seeing on a staging system.

tf:
Seems to work fine. Bug was unrelated.

PR #922, opened 2017-12-24
[WIP] upgrade paperclip and aws dependency

tilsammans:
It's time to move to paperclip version 5 and aws-sdk-s3 version 1.

Lots of specs are breaking, but it seems fixable.

I propose for pageflow 12.1 branch we bump the dependency on paperclip to 4.3.7 and release 12.1.1. Paperclip will show deprecation warnings then, which is going to be very helpful in tracking all the changes.

Paperclip 5 is not backward compatible, as it requires the `s3_region` being set. So we can only do that on pageflow 12.2.


tf:
See #983

PR #923, opened 2017-12-24
[WIP] Upgrade paperclip to 4.3.7

tilsammans:
lots of AWS warnings in the console now, but that's actually good.
Something seems to have been broken with regards to interpolations
though.

paperclip 4.2, video file 4k path:
=> "/v1/main/pageflow/video_files/000/002/290/dash/4k/rendition.mpd"

paperclip 4.3, video file 4k path:
=> "/v1/main/pageflow/video_files/000/002/290/rendition.mpd"

In other words, the `dash/4k' bit is missing.
Still need to figure out why and how. ZencoderAttachment maybe?


tilsammans:
This problem is definitely in ZencodeAttachment @tf and to tell you the truth, I am not sure why it works right now! The `file_name_pattern` ends up in the `base_name_pattern` but not in the path method as far as I can see. Mysterious.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+[Compare changes](https://github.com/codevise/pageflow/compare/12-1-0...v12.1.1)
+
+##### Internals
+
+- Upgraded `paperclip` to version 4.3.7. This is currently the latest in the
+4.3 series and it contains deprecation warnings for the needed aws-sdk upgrade
+in paperclip 5. #923
+
 ### Version 12.1.0
```
tf: The changelog is a semi-generated file (items via [pr_log](https://github.com/tf/pr_log), sections/heading via [semmy](https://github.com/tf/semmy). Making changes in PRs makes this a bit harder and risks having merge conflicts.

tf:
One part of the problem is that Paperclip now [calls `instance.class.name`](https://github.com/thoughtbot/paperclip/blob/da56c39cee6444f5b44b16e345233e910f306bdb/lib/paperclip/interpolations.rb#L88) where it used to call `instance.class.to_s`. Since we only use a pretty [lame double](https://github.com/codevise/pageflow/blob/25b166eab8461415e03ebbe1affc42e2c7e1b461/spec/models/pageflow/zencoder_attachment_spec.rb#L22) for `instance`, this blows up. Not sure, maybe switching to a real model might be worth it here.

tf:
The other is, Paperclip used to allow path segments in `basename` since it only [did some Regex magic](https://github.com/thoughtbot/paperclip/blob/da56c39cee6444f5b44b16e345233e910f306bdb/lib/paperclip/interpolations.rb#L93) to get rid of the extension. Now [it uses `File.basename`](https://github.com/thoughtbot/paperclip/blob/15af98dfe5155ec27c4c7e9c5951c1a2b2d3cadb/lib/paperclip/interpolations.rb#L98) which throws away path segments.

CHANGELOG.md
tilsammans: I did not know! I will revert these changes and trust that you will work your magic to generate everything?

tilsammans:
> paperclip calls instance.class.name

Not true for this patch, which is paperclip 4.3.7. [They did drop the .to_s](https://github.com/thoughtbot/paperclip/blob/v4.3.7/lib/paperclip/interpolations.rb#L93) but left the rest intact.

tilsammans:
> Now it uses File.basename which throws away path segments.

This is indeed the case in 4.3.7.

I'll see what I can do to fix.

tilsammans:
@tf I changed the spec for the url template, but I think for the better. WDYT?

tilsammans:
Hi @tf 

This is as far as I can take this, I'm afraid. I just don't know where to look for the url, path and basename and everything that relates to them! Sorry. I did chuck the test doubles and replaced them with real instances. Indeed, this is required otherwise paperclip will just insert the default_url.

You will have to take this to the finish line, sorry! The good news is, we need to do this anyway for paperclip 5.

CHANGELOG.md
tf: Maybe we can just include these snippets for the changelog in the PR description itself? Especially for things like manual update steps and breaking changes, well written instructions are nothing a tool can generate, of course.

Maybe we should have a wiki page briefly outlining our PR process? I also sometimes add [WIP] prefixes to PRs just to stay on top of what has to be reviewed etc. Without clearly having stated a process, I sometimes fear that changing the title like this might come across a bit passive aggressive. Like saying "Shame! This not done yet!" :smile: 

It would be nice if Github had some more first class support for this. The wip label would be a nice alternative, but, unless I'm mistaken, only users with write access to the repo can edit labels. That would mean (theoretical :wink:) outside contributors would not have a chance to remove the wip label once one of us set it.

tf:
No problem. Thanks for the work you've already put into this. Using a real model will make things more robust for the future. Still, I'm a bit concerned about calling `has_attached_file` on `VideoFile` inside tests. Doesn't this alter the model globally - even beyond the scope of a single test?

tf:
See #983.

PR #924, opened 2017-12-24
fypos fixed

tilsammans:
* storytelling is one word
* pageflow's website is at https://pageflow.io


tilsammans:
I was actually prepared to push this to codevise/pageflow master directly, but muscle memory made me push it to scrollytelling/pageflow instead. These are just typos @tf was I correct in understanding that this one would have been 🆗 directly to master?

tf:
Yes. That would have been totally fine.

PR #926, opened 2017-12-25
[WIP] Ruby 2.5.0

tilsammans:
I am mostly curious if it will just work without changes. Let Travis find that out.

This one accidentally also contains my typo fixes.

I can pull those loose if they don't get merged earlier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
     - node_modules
 
 env:
-  - PAGEFLOW_RAILS_VERSION=4.2.6 PUBLISH_THEME_DOC=true COVERALLS_PARALLEL=true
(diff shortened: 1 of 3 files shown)


tf:
Interesting, indeed. No idea, though, what those errors mean or where they could come from.

tilsammans:
Travis only started supporting 2.5.0 proper since yesterday.

Need to add a `gem update --system` first, then it should work.

tilsammans:
Currently stuck because of the Devise syntax error. See https://mensfeld.pl/2017/12/ruby-2-5-0-upgrade-remarks/

It's fixed in devise 4.4.0 but that opens a huge can of worms, since we're currently on 3.5. :'(

tf:
see #1031

PR #927, opened 2017-12-27
Use Teaspoon fork from pageflow branch

tf:
Includes two fixes:

* Client options for Chrome driver support
  https://github.com/codevise/teaspoon/pull/2

* Backtrace printing
  https://github.com/codevise/teaspoon/pull/1


tf:
To be honest I do not understand why, but using `teaspoon-mocha` from the forked repository appears to also take `teaspoon` from that repository (according to `bundle show`)

PR #930, opened 2017-12-29
use audio5js master

tilsammans:
Upgrade to audio5.js current master, which contains a fix
for the promise returned by play().

On mobile Chrome and Safari autoplay is no longer possible. A user
interaction for playing audio is now mandatory. To check if autoplay
works or if we should show a button (to click play), we need the check
the promise returned by .play().

More details:
https://github.com/zohararad/audio5js/pull/93

<3 @studiorabota


tf:
see #949

PR #932, opened 2017-12-31
Devise 4.4.0

tilsammans:
* Upgraded devise to 4.4.0 (we need this for Ruby 2.5.0 compat)
* Drop devise-async, use ActiveJob for background mails now
* Use current syntax for sign_in method

### DIFF
--- a/spec/support/config/devise.rb
+++ b/spec/support/config/devise.rb
@@ -1,13 +1,4 @@
 RSpec.configure do |config|
-  config.include Devise::TestHelpers, type: :controller
-  config.include Devise::TestHelpers, type: :view_component
-
-  module DeviseTestHelpersWithScope
-    def sign_in(user)
-      super(:user, user)
-    end
-  end
-
-  config.include DeviseTestHelpersWithScope, type: :controller
-  config.include DeviseTestHelpersWithScope, type: :view_component
+  config.include Devise::Test::ControllerHelpers, type: :controller
+  config.include Devise::Test::ControllerHelpers, type: :view_component
 end
--- a/admins/pageflow/user.rb
+++ b/admins/pageflow/user.rb
@@ -132,7 +132,7 @@ module Pageflow
 
       if request.patch?
         if @user.update_with_password(user_profile_params)
-          sign_in @user, bypass: true
+          bypass_sign_in @user, scope: :user
           redirect_to admin_root_path, notice: I18n.t('pageflow.admin.users.me.updated')
         end
       end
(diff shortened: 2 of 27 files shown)


tilsammans:
Needs to be documented:

- [ ] the devise-async initialiser needs to be removed from the host app
- [ ] surface that ideally the devise generator needs to be run again, for 4.4.0 initialiser options.

tilsammans:
I would like to schedule this one for 12.2. The breaking change is that the devise-async initialiser has to be removed from the host application. And ideally the devise generator is run again, although there is no breakage.

Devise 4.4.0 is required to support ruby 2.5.0 in the future.
I saw no evidence of scheduled emails, just async emails, which is what ActiveJob can do.

spec/support/config/devise.rb
```diff
@@ -1,13 +1,4 @@
 RSpec.configure do |config|
-  config.include Devise::TestHelpers, type: :controller
-  config.include Devise::TestHelpers, type: :view_component
-
-  module DeviseTestHelpersWithScope
-    def sign_in(user)
-      super(:user, user)
-    end
```
tf: Let's update this one method to pass the scope option instead of having to change all current and future usages of `sign_in`.

tilsammans: I disagree strongly, actually.

By using the same name as the built-in method, we create confusion because the arguments are different. I personally don't mind all the use of `sign_in` in the specs, because this is formal API surface of devise, and [it is the recommended approach when testing](https://github.com/plataformatec/devise/wiki/How-To:-Sign-in-from-a-controller).

Arguably it could be more DRY, but I think the added layer of indirection hurts more than it helps.

Also: yes, surprisingly this version Just Worked, which is pretty cool!

tilsammans:
> Also I was somehow not aware that ActiveJob is already usable with our current Rails version.

Absolutely, [4.2 includes ActiveJob](http://guides.rubyonrails.org/v4.2/active_job_basics.html).

> Does it use the Resque adapter automatically?

No, it will have to be set in `config/application.rb`.

spec/support/config/devise.rb
tf: Alright then. I see your point. Let's go with the explicit/recommended approach then.

tf:
Since this one will need a manual update step in the changelog anyway, maybe we can mention the required change to have ActiveJob use Resque as well? Then we can have a separate PR to adapt the generator once we decided if we still want to ship Resque inside the default setup (see #940). 

Can you update the PR description with the required changelog steps? Otherwise it is ready for merge, right?

PR #949, opened 2018-02-21
[WIP] Replace Audio5js with Howler to improve mobile background media playback

tf:
Replace audio player which is used to

* play atmo audios
* play audio hotspots of linkmap page

The audio page already uses VideoJS to allow text track display.

Howler uses WebAudio API on modern browsers which allows partially working around autoplay limitations on Safari and mobile browsers:

* Add a button which is displayed initially to enable background media playback. This first user interaction is needed to unlock the audio context.
* Use unlocked audio context to play sound of muted background videos.
* No longer disable atmo on mobile devices.


tf:
@tilsammans We are currently testing this on a staging system. Could you also take it for a spin?

tf:
Have been fighting with some [subtle bugs inside Howler](https://github.com/goldfire/howler.js/issues/912). Optimistic that these can be solved.

PR #952, opened 2018-02-22
Replace PhantomJS with headless Chrome

### LINKED ISSUE #887: Drop PhantomJS; use Chrome headless?
I realise that the introduction of phantomjs to our test suite is relatively recent; but alas the project is not being maintained any more. Chrome headless works fine; [Rails switched too](https://github.com/rails/rails/pull/30930). We do the same?

tf:
Use headless Chrome for:

* Capybara js specs
* Teaspoon specs
* Karma specs inside `node_package`

Fixes #887


tilsammans:
Hurray!

PR #982, opened 2018-08-01
Allow using sassc-rails

tf:
Improve scss compilation times.


tf:
It will probably be best to remove the remove any dependency to `sass-rails` and `sassc-rails` to let the host application decide.

PR #1063, opened 2018-11-06
Add maxlength attribute to text input fields

### LINKED ISSUE #782: Set maxlength on editor text inputs
People often end up typing more text into input fields than can be stored in the respective database column. The save request fails without clear indication what needs to be fixed. To improve the situation we would need to

* add support for a `maxLength` option to the relevant [input views](https://github.com/codevise/pageflow/tree/master/app/assets/javascripts/pageflow/ui/views/inputs) 

* set sensible defaults or configure certain inputs specifically

DanielKuhn:
This will prevent errors where the input exeeds the
maximum character count for mysql varchar types.

fixes #782 

REDMINE-16097

### DIFF
--- a/app/assets/javascripts/pageflow/ui/views/inputs/text_input_view.js
+++ b/app/assets/javascripts/pageflow/ui/views/inputs/text_input_view.js
@@ -4,6 +4,11 @@
  * @param {boolean} [options.required=false]
  *   Display an error if the input is blank.
  *
+ * @param {number} [options.maxLength=255]
+ *   Maximum length of characters for this input.
+ *   To support legacy data which consists of more characters than the specified maxLength,
+ *   the option will only take effect for data which is shorter than the specified maxLength.
+ *
  * @see
  * {@link module:pageflow/ui.pageflow.inputWithPlaceholderText pageflow.inputWithPlaceholderText}
  * for placeholder related further options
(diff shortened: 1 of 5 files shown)


DanielKuhn:
Although text inputs for attributes stored in the `configuration` column could handle longer strings I think it's a good idea to restrict text inputs for text fields generally to a maximum length of 255 characters via the simple html attribute suppressing longer inputs.

I quickly investigated setting limits on textarea inputs to 65535 characters as well (the maximum character count for mysql `text` types), but that would be little out of scope since we're using the wysihtml5 editor for those. If we want to limit those it would probably be a good idea to add a character counter to the textareas displaying how many characters are left.

tf:
> I think it's a good idea to restrict text inputs for text fields generally to a maximum length of 255 characters

Agreed.

> I quickly investigated setting limits on textarea inputs to 65535 characters as well

The inputs for `configuration` settings all contribute to the total length of the stored value together. So restricting the length of one textarea might not even be enough since there still might be multiple text areas storing their data inside the same configuration column. Let's skip this for now and leave text area inputs for another day.

config/locales/de.yml
```diff
@@ -1812,6 +1812,7 @@ de:
             placeholder: Standard (%{text})
           text_input_view:
             required_field: Muss ausgefüllt werden
+            max_characters_exceeded: Maximale Zeichenanzahl überschritten (%{max_length})
```
tf: Please see https://github.com/codevise/pageflow/wiki/Adding-and-Deleting-Translation-Keys for how to change locales in PRs. `locales/{de,en}.yml` are basically generated files which cannot easily be changed directly.

DanielKuhn: Done.

Issue #621, opened 2016-09-08
Redirect back to versions tab after saving current revision


tf:
When clicking the "Save current version" button on the "Versions" tab of an entry admin page (`/admin/entries/n?tab=revisions`), the user is redirected to the top of the page with the "Members" tab active.

The `update` action inside the entry admin controller already [supports redirecting to a specified tab](https://github.com/codevise/pageflow/blob/master/admins/pageflow/entry.rb#L153), but the [button on the revisions tab](https://github.com/codevise/pageflow/blob/master/app/views/components/pageflow/admin/revisions_tab.rb#L49) does not yet pass the `tab` param.

As a further optimization, the [`tabs_view`](https://github.com/codevise/pageflow/blob/master/app/views/components/pageflow/admin/tabs_view.rb) could have an `id` attribute that could be used as an anchor whenever a `tab` param is present in [the `update` action](https://github.com/codevise/pageflow/blob/master/admins/pageflow/entry.rb#L153). That way the user would still see the revisions list after the reload even on smaller screens.

tf:
Closed by #1066

PR #1066, opened 2018-11-07
Redirect to active tab when modifying revisions

DanielKuhn:
Fixes issue https://github.com/codevise/pageflow/issues/621

REDMINE-16102


DanielKuhn:
This works for creating or restoring snapshots.
We might want to respect the `scope` attribute as well, so that when a user restores a snapshot from the "Publications And User Snapshots" or "Frozen" sub-tabs he is also redirected to that same tab, but I don't think it would increase the user experience

PR #1076, opened 2018-11-19
Direct Uploads

DanielKuhn:
Change file upload form to direct upload to S3 via presigned post (as described here: https://docs.aws.amazon.com/de_de/AmazonS3/latest/dev/PresignedUrlUploadObject.html).

Prerequisites / manual update steps:
- Update CORS-config for S3 bucket as described in `doc/setting_up_external_services.md`
- Remove nginx-upload-module (https://github.com/fdintino/nginx-upload-module) in host application

REDMINE-16119


app/assets/javascripts/pageflow/editor/models/file_uploader.js
```diff
@@ -33,7 +33,7 @@ pageflow.FileUploader = pageflow.Object.extend({
                                                   fileTypes: this.fileTypes}));
     }
 
-    this.entry.getFileCollection(fileType).add(file);
+    this.entry.getFileCollection(fileType).create(file);
```
tf: The promise returned by this method should only resolve after `create` invoked its `success` callback. Otherwise there is a chance of trying to access `s3_direct_upload_config` in `UploaderView` before it is available.

app/assets/javascripts/pageflow/editor/views/uploader_view.js
```diff
@@ -14,51 +10,42 @@ pageflow.UploaderView = Backbone.Marionette.View.extend({
...
-              xhr.abort();
-            });
+        pageflow.fileUploader.add(data.files[0]).then(function (record) {
+          data.record = record;
+
+          var s3UplopadConfig = record.attributes.s3_direct_upload_config;
+          form.data({host: s3UplopadConfig.host});
+          fileInput.fileupload({
+            url: s3UplopadConfig.url,
+            formData: s3UplopadConfig.fields
```
tf: Calling `fileupload` again here feels a bit confusing to me. Before, `formData` and `url`/`action` were set in the `submit` hook without needing to invoke the function again.

`record.get('s3_direct_upload_config')` would be bit more idiomatic than `record.attributes.s3_direct_upload_config`, I think.

app/assets/javascripts/pageflow/editor/views/uploader_view.js
```diff
@@ -14,51 +10,42 @@ pageflow.UploaderView = Backbone.Marionette.View.extend({
...
           });
-        }
-        catch(e) {
-          if (e instanceof pageflow.UploadError) {
-            pageflow.app.trigger('error', e);
-          }
-          else {
-            throw(e);
-          }
-        }
```
tf: The `try`/`catch` block is still needed to turn [`FileUploader` exceptons](https://github.com/codevise/pageflow/blob/1f7c77d3da2978a9b67e60182f401cebcc83e199/app/assets/javascripts/pageflow/editor/models/file_uploader.js#L25) into [alerts](https://github.com/codevise/pageflow/blob/1f7c77d3da2978a9b67e60182f401cebcc83e199/app/assets/javascripts/pageflow/editor/initializers/error_listener.js#L2).

app/assets/javascripts/pageflow/editor/views/uploader_view.js
```diff
@@ -14,51 +10,42 @@ pageflow.UploaderView = Backbone.Marionette.View.extend({
...
     this.bindUIElements();
 
-    this.$el.fileupload({
-      acceptFileTypes: new RegExp ('(\\.|\\/)(bmp|gif|jpe?g|png|ti?f|wmv|mp4|mpg|mov|asf|asx|avi|' +
-                                   'm?v|mpeg|qt|3g2|3gp|3ivx|divx|3vx|vob|flv|dvx|xvid|mkv|vtt)$',
-                                   'i'),
-      dataType: 'json',
-
+    var form = this.$el;
+    var fileInput = this.$('input:file');
```
tf: Naming the selector in the `ui` section and referencing it as `this.ui.fileInput` would be the more Backbone idiomatic approach here. Why do we need to to pass the input explicitly now?

app/controllers/pageflow/editor/files_controller.rb
```diff
@@ -74,6 +93,7 @@ def destroy
         authorize!(:edit, entry.to_model)
         verify_edit_lock!(entry)
         entry.remove_file(file)
+        file.destroy if file.uploadable?
```
tf: There already is some [logic around removing files in `DraftEntry`](https://github.com/codevise/pageflow/blob/1f7c77d3da2978a9b67e60182f401cebcc83e199/app/models/pageflow/draft_entry.rb#L60). Maybe this can go (and be tested) there as well?

app/models/concerns/pageflow/hosted_file.rb
```diff
@@ -4,67 +4,50 @@ module HostedFile
     include UploadedFile
 
     included do
-      has_attached_file(:attachment_on_filesystem, Pageflow.config.paperclip_filesystem_default_options)
-      has_attached_file(:attachment_on_s3, Pageflow.config.paperclip_s3_default_options)
+      alias_attribute :file_name, :attachment_on_s3_file_name
```
tf: Let's put the `has_attached_file` call here instead of having each descendant define their own attachment. `HostedFile` can have a method that can be overridden to provide additional attachment options. Also, since there are only attachments on S3 now, I suggest renaming to simply `attachment`.

app/models/concerns/pageflow/hosted_file.rb
```diff
@@ -4,67 +4,50 @@ module HostedFile
...
-      attachment_on_s3.present? ? attachment_on_s3 : attachment_on_filesystem
-    end
-
-    def attachment=(value)
-      self.attachment_on_filesystem = value
-    end
-
-    def keep_on_filesystem_after_upload_to_s3?
-      false
-    end
```
tf: This is something we will have to come back to. `Panorama::Package` [used to](https://github.com/codevise/pageflow-panorama/commit/23c5e6e7a931002f55c62dbfd25b20d82c155f15) rely on this functionality to unpack zip files. It no longer does, so this method is probably unused. Still removing it (which we pretty much have to since nothing is ever stored on the file system anymore now) will be sort of a breaking change.

app/assets/javascripts/pageflow/editor/views/uploader_view.js
DanielKuhn: That's just my old jQuery way of thinking here. Still getting used to Backbone paradigms.
I figured out the `el` from the backbone docs but not the `ui` part. Caught up now by looking at other js files within the project

app/assets/javascripts/pageflow/editor/views/uploader_view.js
DanielKuhn: 👍

app/controllers/pageflow/editor/files_controller.rb
DanielKuhn: True, the logic is there and tested so no need to do it here.
Problem is: [the code] (https://github.com/codevise/pageflow/blob/1f7c77d3da2978a9b67e60182f401cebcc83e199/app/models/pageflow/draft_entry.rb#L60) didn't work as expected: The usages are destroyed via the draft, but the file doesn't know about it at this point, so it never got deleted. A simple `reload` on the usages does the trick, though

app/models/concerns/pageflow/hosted_file.rb
DanielKuhn: resolved with https://github.com/codevise/pageflow/pull/1082

DanielKuhn:
@tf I removed the [WIP] from the title - this is ready to go from my point of view. The only issues left on my list are:
- I still have the behavior of the uploader view not opening a second time for a file that was aborted once - please recheck
- We could remove `doc/using_the_nginx_upload_module.md` I guess
- `vtt` files have not been tested, though I don't see any reason they wouldn't work

app/models/concerns/pageflow/hosted_file.rb
```diff
@@ -82,10 +75,10 @@ def self.columns(t)
...
       t.datetime(:attachment_on_filesystem_updated_at)
 
-      t.string(:attachment_on_s3_file_name)
-      t.string(:attachment_on_s3_content_type)
-      t.integer(:attachment_on_s3_file_size, limit: 8)
-      t.datetime(:attachment_on_s3_updated_at)
+      t.string(:attachment_file_name)
+      t.string(:attachment_content_type)
+      t.integer(:attachment_file_size, limit: 8)
+      t.datetime(:attachment_updated_at)
```
tf: Since the method is already deprecated and renaming the column here would be the sort of inconvenient breaking change the method was originally deprecated for, I suggest we remove it in 13.0 just like `keep_on_filesystem`.

app/models/concerns/pageflow/uploaded_file.rb
```diff
@@ -55,5 +55,13 @@ def cache_key
...
       "#{super}-#{state}"
     end
+
+    def direct_upload_config
+      presigned_post_config = Pageflow.config.paperclip_direct_upload_options.call(attachment)
+      {
+        url: presigned_post_config.url,
+        fields: presigned_post_config.fields
+      }
+    end
```
tf: Let's change `Pageflow.config.paperclip_direct_upload_options` to return this hash directly. This way we can avoid the `OpenStruct` in the test stub.

app/models/pageflow/image_file.rb
```diff
@@ -1,84 +1,63 @@
...
-    def attachment
-      processed_attachment.present? ? processed_attachment : unprocessed_attachment
+    def attachment_default_url
+      ':pageflow_placeholder'
     end
 
-    def attachment=(value)
-      self.unprocessed_attachment = value
-    end
+    do_not_validate_attachment_file_type(:attachment)
```
tf: Should no longer be needed, right?

app/views/pageflow/editor/files/_file.json.jbuilder
```diff
@@ -5,9 +5,10 @@ json.partial!(partial: 'pageflow/files/file',
 json.call(file,
           :state,
           :rights,
-          :usage_id)
+          :usage_id,
+          :direct_upload_config)
 
-json.retryable(file.retryable?)
+json.retryable(file.respond_to?(:retryable?) ? file.retryable? : false)
```
tf: If `HostedFile` had a default implementation of `retryable` returning `false` this would not be necessary.

db/migrate/20181123103416_rename_attachment_on_s3_to_attachment_for_all_hosted_files.rb
```diff
@@ -0,0 +1,18 @@
...
+                    :attachment_file_name
+      rename_column "pageflow_#{hosted_file_type}_files",
+                    :attachment_on_s3_content_type,
+                    :attachment_content_type
+      rename_column "pageflow_#{hosted_file_type}_files",
+                    :attachment_on_s3_file_size,
+                    :attachment_file_size
+      rename_column "pageflow_#{hosted_file_type}_files",
+                    :attachment_on_s3_updated_at,
+                    :attachment_updated_at
```
tf: Is `file_size`/`content_type` still set? 

We still need think about migrating `pageflow-panorama`.

spec/factories/hosted_files.rb
```diff
@@ -12,19 +12,25 @@ module Pageflow
...
+        attachment { nil }
+        attachment_file_name { 'image.jpg' }
+        state { 'uploading' }
+
+        after :create do |hosted_file|
+          FileUtils.mkdir_p(File.dirname(hosted_file.attachment.path))
+          attachment_path = Engine.root.join('spec', 'fixtures', hosted_file.attachment_file_name)
+          unless File.identical?(attachment_path, hosted_file.attachment.path)
+            FileUtils.cp(attachment_path, hosted_file.attachment.path)
+          end
```
tf: Can this logic be extracted into some helper/shared factory trait?

db/migrate/20181123103416_rename_attachment_on_s3_to_attachment_for_all_hosted_files.rb
DanielKuhn: `file_size` yes, `content_type` no, apparently. But as discussed, I'll remove this migration for compatibility reasons (S3 bucket name)

app/models/concerns/pageflow/hosted_file.rb
DanielKuhn: It's still referenced in the docs, though (`creating_file_types.md`). I'll remove it and update the doc accordingly

app/jobs/pageflow/process_file_job.rb
```diff
@@ -5,13 +5,16 @@ class ProcessFileJob < ApplicationJob
...
+        :ok
+      else
+        raise ActiveRecord::RecordInvalid
+      end
     rescue ActiveRecord::RecordInvalid, Errno::ENAMETOOLONG
-      file.processed_attachment = nil
-      file.save!
+      file.attachment = nil
+      file.attachment_file_name = 'error'
       :error
```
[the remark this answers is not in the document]
DanielKuhn: We actually need to remove the attachment if it's invalid. Otherwise the state transition fails due to the attachment validation. But I'll change it so that we keep the `file_name` in order to be able to retry the processing

app/models/concerns/pageflow/hosted_file.rb
```diff
@@ -4,63 +4,56 @@ module HostedFile
...
+      has_attached_file(:attachment,
+                        Pageflow.config.paperclip_s3_default_options
+                          .merge(
+                            default_url: lambda do |attachment|
+                              attachment.instance.attachment_default_url
+                            end,
+                            styles: lambda do |attachment|
+                              attachment.instance.attachment_styles(attachment)
+                            end
+                          ))
```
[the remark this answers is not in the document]
DanielKuhn: True. I changed the attachment name back to `attachment_on_s3` to avoid legacy problems. `ImageFile` (which is now a `HostedFile` as well) provides a new method `attachments_path_name` which is called in the Paperclip initializer for path interpolation and returns `processed_attachments`. For all other attachments this defaults to `attachment_on_s3s`

app/controllers/pageflow/editor/files_controller.rb
```diff
@@ -81,7 +90,7 @@ def destroy
       private
 
       def create_params
-        file_attachment_params
+        file_params.permit(:file_name)
```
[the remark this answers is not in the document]
DanielKuhn: Paperclip actually doesn't sanitize the file name. I added a validation on `HostedFile` and a spec for the `FilesController` to prevent file_name-spoofing

app/models/pageflow/image_file.rb
DanielKuhn: Removed now and replaced by filename validation (disallowing forward and backward slashes in the filename)

tf: I mean, not required _here_ since `HostedFile` already defines it.

app/models/concerns/pageflow/hosted_file.rb
```diff
@@ -62,27 +70,6 @@ def original_url
...
-      t.datetime(:attachment_on_filesystem_updated_at)
-
-      t.string(:attachment_file_name)
-      t.string(:attachment_content_type)
-      t.integer(:attachment_file_size, limit: 8)
-      t.datetime(:attachment_updated_at)
-
-      t.timestamps
-    end
-
```
tf: Extracted this into #1085 since it's a breaking change that needs to go into 13.0 and this PR will probably be part of 13.1. Can you rebase onto master?

DanielKuhn: Done

config/initializers/paperclip.rb
```diff
@@ -10,6 +10,11 @@
   Pageflow::PaperclipInterpolations::Support.class_basename(attachment, style)
 end
 
+Paperclip.interpolates(:attachments_path_name) do |attachment, _style|
+  record = attachment.instance
+  record.respond_to?(:attachments_path_name) ? record.attachments_path_name : 'attachment_on_s3s'
```
tf: Hard coding the name does not work. There are other attachments (e.g. video posters) that need to use their proper name for this interpolation. Since [`interpolates` merely defines a method](https://github.com/thoughtbot/paperclip/blob/17050e41aa31613da23c7c1b56381f431ee9d3ff/lib/paperclip/interpolations.rb#L13) on `Paperclip::Interpolations`, we should be able to call the original `attachment` method when we rename the `attachment` param to prevent shadowing.

DanielKuhn: Fixed in 039b0208615dfa81c26b592860db9b541438ee9b

app/models/pageflow/positioned_file.rb
```diff
@@ -34,7 +34,7 @@ def initialize
       end
 
       def thumbnail_url(*args)
-        ImageFile.new.attachment.url(*args)
+        ImageFile.new.thumbnail_url(*args)
```
tf: This broke a spec. Positioned file depends on getting the placeholder if the thumbnail is missing not the empty string. Maybe let's try to keep new changes as minimal as possible at this point.

app/models/concerns/pageflow/hosted_file.rb
```diff
@@ -67,7 +67,7 @@ def basename
     end
 
     def url
-      if attachment.present?
+      if ready?
         attachment.url
       end
```
tf: While this looks like a good idea, I'm not really sure of the implications of the change. Just like in the `thumbnail_url` case below, I'd suggest not make further changes unless they are really necessary.

app/models/concerns/pageflow/uploaded_file.rb
```diff
@@ -55,5 +55,12 @@ def cache_key
       # prevent caching outdated information.
       "#{super}-#{state}"
     end
+
+    def direct_upload_config
+      # the direct uploaad config is only needed for the initial upload.
+      # It can be omitted once the instance leaves the initial "uploading" state
+      return {} unless uploading?
+      Pageflow.config.paperclip_direct_upload_options.call(attachment)
+    end
```
tf: Thinking some more about this, I'm not convinced this is the best place to handle this case distinction. The method name suggests that this always returns an upload config. The fact that the comment is needed is another indicator.

The important questions is not about how we create an upload config, but about when we actually present it to the client. So I think the `pageflow/editor/files/_file.json.jbuilder` template would be a much better place for that. This would also give us a decent place to test this logic in a [controller spec](https://github.com/codevise/pageflow/pull/1076/files#diff-547d0353f9388654f30f0ba308e0d3f3R199)

spec/controllers/pageflow/editor/files_controller_spec.rb
```diff
@@ -192,14 +192,31 @@ module Pageflow
              params: {
                entry_id: entry,
                collection_name: 'image_files',
-               image_file: {attachment: image_fixture_upload}
+               image_file: {file_name: 'image.jpg'}
              },
              format: 'json')
 
-        expect(entry.image_files.first.unprocessed_attachment_file_name).to be_present
+        expect(entry.image_files.first.direct_upload_config).to be_present
```
tf: This should look at the response body instead of just calling the method.

config/initializers/paperclip.rb
```diff
@@ -10,6 +10,12 @@
   Pageflow::PaperclipInterpolations::Support.class_basename(attachment, style)
 end
 
+Paperclip.interpolates(:attachments_path_name) do |attachment, _style|
+  record = attachment.instance
+  return record.attachments_path_name if record.respond_to?(:attachments_path_name)
+  attachment.name.to_s.pluralize
+end
```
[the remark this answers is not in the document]
DanielKuhn: changed in 83fe6c148e9a54ec6303b6f995001a7e7168ab01

app/models/pageflow/positioned_file.rb
DanielKuhn: reverted in 61e91ea06b1c901a535a06ef9a66c78b7752eb20

app/models/concerns/pageflow/hosted_file.rb
DanielKuhn: You're right, this looked like a quick fix because `ready?` is defined as `attachment.present?` a couple of lines above, but since `ready` is overwritten by processing state machines we should leave it as is for now.
Reverted in 6232dad4ea664772d1591aa1ee3937c005dd7179

app/models/concerns/pageflow/uploaded_file.rb
DanielKuhn: changed and covered with spec in f0770050193e889ec9c9a5adb61e8094cad926f4

spec/controllers/pageflow/editor/files_controller_spec.rb
DanielKuhn: changed in f0770050193e889ec9c9a5adb61e8094cad926f4

PR #1100, opened 2019-01-07
Set locale to editors locale for new entries

DanielKuhn:
Customers who have set their language to english have reported that the default language for new stories is still German. This behavior was perceived as a bug.

This PR introduces a new column `default_locale` on account/theming level where a default language for new stories can be specified (defaults to German), which is then used for new stories.

For new accounts the `default_locale` will initially get set to the locale the user specified during signup.

REDMINE-16323

### DIFF
--- a/app/views/admin/accounts/_form.html.erb
+++ b/app/views/admin/accounts/_form.html.erb
@@ -21,7 +21,9 @@
       <%= theming.input :default_locale,
                         as: :select,
                         include_blank: false,
-                        collection: available_locales_collection,
+                        collection: Pageflow.config.available_public_locales.map{ |locale|
+                          [I18n.t("pageflow.languages.#{locale}"), locale.to_s]
+                        },
                         hint: t('pageflow.admin.themings.default_locale_hint') %>
       <%= theming.input :default_author,
         hint: t('pageflow.admin.themings.default_author_hint'),
--- a/app/views/admin/accounts/_form.html.erb
+++ b/app/views/admin/accounts/_form.html.erb
@@ -22,7 +22,7 @@
                         as: :select,
                         include_blank: false,
                         collection: Pageflow.config.available_public_locales.map{ |locale|
-                          [I18n.t("pageflow.languages.#{locale}"), locale.to_s]
+                          [t('pageflow.public._language', locale: locale), locale.to_s]
                         },
                         hint: t('pageflow.admin.themings.default_locale_hint') %>
       <%= theming.input :default_author,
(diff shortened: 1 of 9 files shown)


app/views/admin/accounts/_form.html.erb
```diff
@@ -18,6 +18,13 @@
     <% end %>
 
     <%= f.inputs do %>
+      <%= theming.input :default_locale,
+                        as: :select,
+                        include_blank: false,
+                        collection: Pageflow.config.available_public_locales.map{ |locale|
+                          [I18n.t("pageflow.languages.#{locale}"), locale.to_s]
```
tf:
```suggestion
                          [I18n.t('pageflow.public._language', locale: locale), locale.to_s]
```

see https://github.com/codevise/pageflow/blob/f92c4352a375fcbfba50739aac6d1e52ddb496c3/app/assets/javascripts/pageflow/editor/views/edit_meta_data_view.js#L29-L34

DanielKuhn: I saw that, but that would give us the local name of the locale. Since the user of the admin panel is either german or english, I thought it would be better to translate the names of the locales to the language specified for the current user

tf: On the other hand that'S also how it works when selecting a locale inside the editor interface (which is also only either german or english). Also this would mean we would have to release a new version of the Pageflow gem whenever a locale is added to the `pageflow-public-i18n`, which would defeat one of the main reasons to decouple translations into their own gem.

DanielKuhn: True. Ok, I'll change it to use the local names like in the editor

config/locales/new/create_new_entry_in_editors_locale.de.yml
```diff
@@ -0,0 +1,20 @@
...
+      bg: Bulgarisch
+      cs-CZ: Tschechisch
+      fa-IR: Persisch
+      gn: Guarani
+      id: Indonesisch
+      ms: Malaysisch
+      nl-BE: Holländisch (Belgien)
+      pl-PL: Polnisch
+      pt-BR: Portugiesisch (Brasilien)
+      ru-RU: Russisch
```
DanielKuhn: I'll leave these in. Who knows when they might be needed.

tf: If they are unused, I'd much prefer if we removed them and maybe stored the list somewhere else for later use.

PR #1123, opened 2019-02-21
Fix mask image in Safari 12

tf:
Safari 12 requires CORS headers for cross origin mask images. Do not
use asset host in mask image url to support CDN setups which do not
set CORS headers for static assets.

REDMINE-16141

x# Please enter the commit message for your changes. Lines starting


tf:
Does not work since image url is relative to css url.

See #1124

PR #1133, opened 2019-03-07
Add shared specs to pageflow-support to lint page types

tf:
This can be used by plugin engines to

* check that rendering page template does not raise errors

* check that rendering json seed does not raise errors

* check that all required translations are defined

Make `pageflow-support` depend on `factory_bot_rails` instead of
`factory_bot` to make sure factories are loaded automatically.

REDMINE-16580


tf:
Please don't merge yet. Still want to make sure that the shared tests pass for all page type plugins.

PR #1138, opened 2019-03-12
Structured data for videos, audios and background images

tf:
Branch still contains commits that could be turned into separate PRs.

REDMINE-16512


tf:
see #1145

PR #1147, opened 2019-04-04
Add migration to fix "processing_failed"-image files

DanielKuhn:
REDMINE-16706


tf:
In the other migration we copied all attachment attributes. Is there are reason why we only copy the file name here?

DanielKuhn:
@tf no reason, good catch

DanielKuhn:
I updated the migration so that it also fixes the records which were manipulated via production console. Ready to merge now

PR #1150, opened 2019-04-08
Remove papclip_filesystem_root config

DanielKuhn:
It's not needed anymore since we're not storing anything on the filesystem any longer

REDMINE-16119
REDMINE-16715


DanielKuhn:
@tf I found this branch in my local git and thought I'd push it out now that direct uploads are in place.

tf:
So this would be another breaking change that would need to ship with Pageflow 14, right? Have you checked that none of the other plugins still depend on any of this? Nothing comes to mind, but still...

DanielKuhn:
@tf The only other usages I could find are in the initializers of the host applications. But I added a deprecation warning for safety now.

PR #1156, opened 2019-05-07
Add lists to WYSIHTML editor

DanielKuhn:
REDMINE-16557

### DIFF
--- a/app/assets/javascripts/pageflow/ui/views/inputs/text_area_input_view.js
+++ b/app/assets/javascripts/pageflow/ui/views/inputs/text_area_input_view.js
@@ -73,6 +73,9 @@ pageflow.TextAreaInputView = Backbone.Marionette.ItemView.extend({
           u: {unwrap: this.options.disableRichtext ? 1 : 0},
           b: {unwrap: this.options.disableRichtext ? 1 : 0},
           i: {unwrap: this.options.disableRichtext ? 1 : 0},
+          ol: {unwrap: this.options.enableLists ? 0 : 1},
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/pageflow/ui/views/inputs/text_area_input_view.js
```diff
@@ -68,6 +68,9 @@ pageflow.TextAreaInputView = Backbone.Marionette.ItemView.extend({
           u: {unwrap: this.options.disableRichtext ? 1 : 0},
           b: {unwrap: this.options.disableRichtext ? 1 : 0},
           i: {unwrap: this.options.disableRichtext ? 1 : 0},
+          ol: {unwrap: this.options.disableRichtext ? 1 : 0},
+          ul: {unwrap: this.options.disableRichtext ? 1 : 0},
+          li: {unwrap: this.options.disableRichtext ? 1 : 0},
```
tf: Shouldn't this depend on `enableLists` instead?

DanielKuhn: `enableLists` controls whether or not the list-buttons are shown. `disbaleRichtext` is an independent option which causes the tags to get unwrapped/converted to plain text

tf: Yes. But `u`/`b`/`i` tags are unwrapped when `disableRichtext` is set to ensure those get stripped when pasting text from the clipboard. Wouldn't we want to strip out lists in copied texts as well if `enableLists` is false even if rich text is allowed?

PR #1163, opened 2019-05-23
WIP: Introduce perma_ids for ID-agnostic referencing

DanielKuhn:
In order to be able to move entries from dedicated to hosted pageflow,
this introduces a perma_id for UploadedFiles which can be used to
reference files within the scope of their entries even when their
ID changes during import into the new database.

REDMINE-16809

### DIFF
--- a/app/helpers/pageflow/video_files_helper.rb
+++ b/app/helpers/pageflow/video_files_helper.rb
@@ -25,9 +25,9 @@ module Pageflow
                   style: "background-position: #{position[:x]}% #{position[:y]}%;")
     end
 
-    def poster_image_tag(video_id, poster_image_id, options = {})
-      video_file = VideoFile.find_by_id(video_id)
-      poster = ImageFile.find_by_id(poster_image_id)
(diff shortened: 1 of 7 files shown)


app/helpers/pageflow/video_files_helper.rb
```diff
@@ -103,12 +98,11 @@ def video_file_script_tag(video_id, options = {})
...
     end
 
-    def video_file_non_js_link(entry, video_file_id)
-      if (video_file = VideoFile.find_by_id(video_file_id))
-        link_to(t('pageflow.public.play_video'),
-                short_video_file_path(entry, video_file),
-                class: 'hint')
-      end
+    def video_file_non_js_link(entry, video_file_perma_id)
+      return unless video_file = VideoFile.find_in_entry(@entry, video_file_perma_id)
```
tf: find_file_in_entry(:video_file, video_file_id)

pageflow.config.file_types.find(:video_file).model.find_

PR #1178, opened 2019-06-17
Always lookup files by revisions usage scope

DanielKuhn:
Introduce RevisionFileHelper and migrate all existing find_by_id calls
to find_file_in_entry(file_type, file_id) calls.
Stub finding files in usage scope for helper-specs by adding RevisionFileTestHelper.
Simplify lookup by usage scope by introducing UsedFile-factory.

REDMINE-16809

REDMINE-16877

### DIFF
--- a/app/models/pageflow/revision.rb
+++ b/app/models/pageflow/revision.rb
@@ -93,6 +93,12 @@ module Pageflow
       UsedFile.new(file)
     end
 
+    def find_file_by_id(model, id)
+      file = files(model).find_by(id: id)
+      return unless file
+      UsedFile.new(file)
+    end
+
     def share_providers
       self[:share_providers] || entry.theming.default_share_providers
     end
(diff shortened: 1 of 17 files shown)


app/models/pageflow/revision.rb
```diff
@@ -89,7 +89,8 @@ def find_files(model)
     end
 
     def find_file(model, id)
-      file = files(model).find(id)
+      file = files(model).find_by(id: id)
+      return unless file
       UsedFile.new(file)
     end
```
tf: `find_file` is also used in controllers ([example](https://github.com/codevise/pageflow/blob/6f240d5b36fb0d705f6413966804bbc450ac90e1/app/controllers/pageflow/editor/files_controller.rb#L82)) which rely on `RecordNotFound` exceptions to create 404 responses. Maybe we should add `find_file_by_id` and rename it to `find_file_by_perma_id` in the next PR?

DanielKuhn: Good idea. Done!

PR #1179, opened 2019-06-17
Change file lookup to usage.perma_id

DanielKuhn:
When moving entries from one environment to another, the ids of associated files change. Therefore we introduce a `perma_id` on FileUsage which stays unchanged during import/export and is then used for the file lookup in the importing environment

REDMINE-16809


app/models/pageflow/thumbnail_file_resolver.rb
```diff
@@ -35,7 +37,8 @@ def find_positioned_file_by_candiate(candidate)
     end
 
     def find_file_by_candidate(candidate)
-      file_model(candidate).find_by_id(record_id(candidate))
+      @entry = entry # must be set for finding file in usage scope
+      find_file_in_entry(file_model(candidate), record_id(candidate))
```
tf: Let's call `@entry.find_file_by_perma_id` directly instead.

spec/factories/used_file.rb
```diff
@@ -0,0 +1,17 @@
...
+
+      initialize_with do
+        file = build(model)
+        usage = file.usages.build(revision: revision)
+
+        UsedFile.new(file, usage)
+      end
+    end
+  end
+end
```
tf: How do the cases in which the `:used_file` factory is used differ from the `create_used_file` cases?

PR #1180, opened 2019-06-24
Refactor file modules

DanielKuhn:
### File modules:
`HostedFile` => `UploadableFile` (concerning uploading of files)
`UploadedFile` => `ReusableFile` (concerning usage of files)

### StateMachines:
`ProcessedFileStateMachine` => `ImageAndTextTrackProcessingStateMachine`
`EncodedFileStateMachine ` => `MediaEncodingStateMachine`

### StateMachineJobs:
`ProcessFileJob` => `ProcessImageOrTextTrackJob `

REDMINE-16877


app/state_machines/pageflow/media_encoding_state_machine.rb
```diff
@@ -19,7 +19,7 @@ module EncodedFileStateMachine
           transition any => 'waiting_for_encoding'
         end
 
-        event :retry do
+        event :can_retry_encoding do
```
tf: Since the state machine already defines `can_retry_encoding?`, this difference in method names is a little to subtle for my taste.  Maybe just `retry_encoding`?

app/models/concerns/pageflow/reusable_file.rb
```diff
@@ -56,14 +56,66 @@ def cache_key
...
+      'unused'
+    end
+
+    # Overwritten with the actual attachment of the file type.
+    # Unifies the handling of attachments throughout the codebase by relying on
+    # `attachment` always being present regardless of its actual name.
+    def attachment
+      require 'ostruct'
+      OpenStruct.new(original_file: 'unused')
+    end
```
tf: Where is this use? File types like `Chart::ScrapredSite` do not have a single attachment.

PR #1186, opened 2019-07-10
Configurable loading spinner (14.5 Backport)

awaismbajwa:
REDMINE-16560
Backport of #1215

### DIFF
--- a/app/models/pageflow/widget.rb
+++ b/app/models/pageflow/widget.rb
@@ -61,10 +61,15 @@ module Pageflow
       private
 
       def all
-        placeholders_by_role
-          .merge(defaults_by_role)
-          .merge(from_db_by_role)
-          .values
+        initial_widgets = placeholders_by_role.merge(defaults_by_role)
+        initial_widgets.merge(from_db_by_role) { |_role_key, old_val, new_val|
+          if old_val.configuration.present?
+            old_val.configuration.each do |key, value|
+              new_val.configuration[key] = value unless new_val.configuration.key?(key)
+            end
+          end
+          new_val
+        }.values
       end
 
       def from_db_by_role
@@ -73,13 +78,15 @@ module Pageflow
 
       def defaults_by_role
         config.widget_types.defaults_by_role.each_with_object({}) do |(role, widget_type), result|
-          result[role] = Widget.new(role: role, type_name: widget_type.name, subject: nil)
+          result[role] = Widget.new(role: role, type_name: widget_type.name,
+                                    subject: nil,
+                                    configuration:
+                                      config.widget_types.default_configuration(widget_type.name))
         end
       end
 
       def placeholders_by_role
         return {} unless options[:include_placeholders]
-
         config.widget_types.roles.each_with_object({}) do |role, result|
           result[role] = Widget.new(role: role, type_name: nil, subject: nil)
         end
--- a/app/models/pageflow/widget.rb
+++ b/app/models/pageflow/widget.rb
@@ -64,9 +64,8 @@ module Pageflow
         initial_widgets = placeholders_by_role.merge(defaults_by_role)
         initial_widgets.merge(from_db_by_role) { |_role_key, old_val, new_val|
           if old_val.configuration.present?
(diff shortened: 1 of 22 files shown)


awaismbajwa:
@tf Could you please review it?

node_package/src/loadingSpinner/components/MediaLoadingSpinner.jsx
```diff
@@ -0,0 +1,56 @@
...
+          <div className="title_loading_spinner-image"
+               style={backgroundImageInlineStyles(this.props)} />
+        </div>
+      );
+    }
+    else {
+      return <noscript />;
+    }
+  }
+}
```
tf: Inheritance is rather unidiomatic in React. The overridden `render` method requires a lot of formerly private functions of `TitleLoadingSpinner` to be exported which creates further dependencies and weakens encapsulation. 

Instead, I'd suggest to go with composition. Comparing two render methods I only see a few differences:

* `MediaLoadingSpinner` allows removing the logo. Any reason not to allow this for `TitleLoadingSpinner`?
* `TitleLoadingSpinner` supports invert option. If the logo is not removed, wouldn't this also be required for `MediaLoadingSpinner`?
* `TitleLoadingSpinner` contains additional elements for the title elements.

I'd therefore suggest to make `MediaLoadingSpinner` render `props.children` right after the background image div. Then `TitleLoadingSpinner` can use composition to render a `MediaLoadingSpinner` component passing the additional title elements as children.

`TitleLoadingSpinner` can still be connected separately to ask for additional widget attributes like `title` and `subtitle`, but all the now duplicated "map state to props" entries (editing, firstPageBackgroundImage. backgroundImage, ..) only need to live in `TitleLoadingSpinner`.

Since now media loading spinner is the more basic concept, we should consider renaming the CSS classes that have a `title_loading_spinner-` prefix to `media_loading_spinner-*`.

node_package/src/loadingSpinner/components/TitleLoadingSpinner.jsx
```diff
@@ -121,7 +123,9 @@ export function register() {
       }),
       entryTitle: entryAttribute('title'),
       title: widgetAttribute('title', {role: 'loading_spinner'}),
-      subtitle: widgetAttribute('subtitle', {role: 'loading_spinner'})
+      subtitle: widgetAttribute('subtitle', {role: 'loading_spinner'}),
+      invert: widgetAttribute('invert', {role: 'loading_spinner'}),
```
tf: If no custom background image has been selected, the background image of the first page is used. It would be nice if we could also use the `invert` attribute of the first page in this case. Otherwise the text on the loading spinner will be hard to read by default if the first page is inverted and uses a light image. 

We could even consider hiding the invert chech box in the loading spinner configuration editor when no custom background image is selected (see [`visibleBinding` and `visbile` options of input views](https://github.com/codevise/pageflow/blob/0c99e4f60703d8e60870ebb3dec6995205e7fee1/app/assets/javascripts/pageflow/ui/views/mixins/input_view.js#L97)).

app/models/pageflow/widget.rb
```diff
@@ -73,13 +78,15 @@ def from_db_by_role
 
       def defaults_by_role
         config.widget_types.defaults_by_role.each_with_object({}) do |(role, widget_type), result|
-          result[role] = Widget.new(role: role, type_name: widget_type.name, subject: nil)
+          result[role] = Widget.new(role: role, type_name: widget_type.name,
+                                    subject: nil,
+                                    configuration:
+                                      config.widget_types.default_configuration(widget_type.name))
```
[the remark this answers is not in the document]
awaismbajwa: two specs are added to widget_spec.rb

node_package/src/loadingSpinner/components/MediaLoadingSpinner.jsx
awaismbajwa: Refactored MediaLoadingSpinner and TitleLoadingSpinner.
MediaLoadingSpinner is now composed by TitleLoadingSpinner using its connect component

PR #1193, opened 2019-07-31
Facebook open graph image share changes

awaismbajwa:
REDMINE-16972
+added og:image:height and width meta properties.
-removed placeholder images for social share

### DIFF
--- a/app/models/pageflow/page.rb
+++ b/app/models/pageflow/page.rb
@@ -16,6 +16,18 @@ module Pageflow
       configuration['title'].presence || configuration['additional_title']
     end
 
+    def social_share_url(*args)
+      thumbnail_file.thumbnail_url(*args) unless thumbnail_file.blank?
+    end
+
+    def thumbnail_url(*args)
+      thumbnail_file.thumbnail_url(*args)
+    end
+
+    def thumbnail_file
+      ThumbnailFileResolver.new(page_type.thumbnail_candidates, configuration).find
+    end
+
     def page_type
       Pageflow.config.page_types.find_by_name!(template)
     end
--- a/app/models/pageflow/page.rb
+++ b/app/models/pageflow/page.rb
@@ -16,18 +16,6 @@ module Pageflow
       configuration['title'].presence || configuration['additional_title']
     end
 
-    def social_share_url(*args)
-      thumbnail_file.thumbnail_url(*args) unless thumbnail_file.blank?
-    end
-
-    def thumbnail_url(*args)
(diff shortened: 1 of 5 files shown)


app/views/pageflow/social_share/_image_tags.html.erb
```diff
@@ -1,5 +1,7 @@
 <% image_urls.each do |image_url| %>
   <%= tag :meta, property: "og:image", content: social_share_normalize_protocol(image_url) %>
+  <%= tag :meta, property: "og:image:width", content: 1200 %>
+  <%= tag :meta, property: "og:image:height", content: 630 %>
```
tf: Where do these values come from? The `medium` style is defined as `1024x1024>`. I think we will need a dynamic method to calculate the correct size. The dimensions of the original image are also stored on `ImageFile`. Maybe Facebook only cares about the ratio - not the actual size?

app/models/pageflow/page.rb
```diff
@@ -16,6 +16,18 @@ def title
...
+    end
+
+    def thumbnail_url(*args)
+      thumbnail_file.thumbnail_url(*args)
+    end
+
+    def thumbnail_file
+      ThumbnailFileResolver.new(page_type.thumbnail_candidates, configuration).find
+    end
+
```
tf: `ThumbnailFileResolver` now requires an entry. The `thumbnail_file` method has been moved. You can talk to Daniel about the changes he made in that area.

app/views/pageflow/social_share/_image_tags.html.erb
awaismbajwa: I am not sure how facebook uses these width height parameters. they have mentioned this 
> 
Use og:image:width and og:image:height Open Graph tags
Using these tags will specify the image dimensions to the crawler so that it can render the image immediately without having to asynchronously download and process it.
> 
from my observation I could not see any difference in post rendering when specifying different width and height
I found the above values 1200x600 by just searching the most optimum values for facebook image share.
We can  already have mechanism to get image dimensions in image_file.rb save_image_dimensions we can apply that?

awaismbajwa: applied actual image width and height from ImageFile

DanielKuhn (approved):

tf (commented):
Can you add specs for the newly added logic?

awaismbajwa:
> Can you add specs for the newly added logic?

specs are added

PR #1195, opened 2019-08-02
Entry add member changes

awaismbajwa:
* add user button is not shown if there are no users in the account
* current user is not considered in the list of users to be added
* If there are no users in the account, quota is checked and
message is displayed if exhausted, otherwise invite users message is displayed

REDMINE-16606

### DIFF
--- a/app/views/components/pageflow/admin/add_membership_button.rb
+++ b/app/views/components/pageflow/admin/add_membership_button.rb
@@ -16,6 +16,7 @@ module Pageflow
           path = new_admin_entry_membership_path(
             parent, entity_type: to_class_name(entity_type)
           )
+          quota = Pageflow.config.quotas.get(:users, parent.account)
         else
           path = new_admin_account_membership_path(
             parent, entity_type: to_class_name(entity_type)
@@ -28,7 +29,13 @@ module Pageflow
           rel = 'add_member'
         end
 
-        if membership_possible_for(user, parent, entity_type)
+        if parent.is_a?(Entry) && parent.account.users.length < 2
+          render 'pageflow/admin/entries/cannot_add_user',
+                 user: user,
+                 parent: parent,
+                 entity_type: entity_type,
+                 quota: quota
+        elsif membership_possible_for(user, parent, entity_type)
           para do
             link_to(button_label, path, class: 'button', data: {rel: rel})
           end
@@ -38,6 +45,11 @@ module Pageflow
                            class: 'button disabled',
                            data: {rel: rel}),
                'data-tooltip' => data_tooltip)
+          render 'pageflow/admin/users/cannot_add',
+                 user: user,
+                 parent: parent,
+                 entity_type: entity_type,
+                 quota: quota
         end
       end
 
--- a/spec/views/components/pageflow/admin/add_membership_button_spec.rb
+++ b/spec/views/components/pageflow/admin/add_membership_button_spec.rb
@@ -11,6 +11,54 @@ module Pageflow
         allow(helper).to receive(:url_for)
       end
 
+      it 'renders add user button when there is user to add' do
+        common_account = create(:account)
+        create(:user, :member, on: common_account)
+        account_manager = create(:user, :manager, on: common_account)
+        entry = create(:entry, account: common_account,
+                               with_editor: account_manager,
+                               title: 'Entry 1')
+
+        sign_in(account_manager, scope: :user)
+
+        render do
+          add_membership_button(account_manager, entry, 'entry')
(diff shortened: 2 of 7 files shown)


awaismbajwa:
REDMINE-16606

awaismbajwa:
> I think, this would be more flexible if we provided a way to render custom partials similar to the way [`pageflow/admin/users/_quota_exhausted.html.erb`](https://github.com/codevise/pageflow/blob/master/app/views/pageflow/admin/users/_quota_exhausted.html.erb) is used in [`Pageflow::Admin::UsersHelper`](https://github.com/codevise/pageflow/blob/master/app/helpers/pageflow/admin/users_helper.rb#L13).
> 
> You can also apply the [same pattern you see there](https://github.com/codevise/pageflow/blob/master/spec/helpers/pageflow/admin/users_helper_spec.rb#L25) to add specs for this new feature.
> 
> That way we can easily render more complex things (like links to the upgrade page) in Hosted Pageflow.

I agree, I had a look at it too but there are some counter arguments 
1. Quota is already used throughout pageflow and similar checks are applied therefore I thought it would be alright to use it for this small change.
2. add_membership_button is implemented as a component so not sure if this should be taken out of its implementation.

tf:
1. I also think that using quota is the right approach.

2. Stub template should also work for templates [rendered from components](https://www.rubydoc.info/gems/arbre2/2.1.0/Arbre/Rails/Rendering).

app/views/components/pageflow/admin/add_membership_button.rb
```diff
@@ -38,6 +39,10 @@ def build(user, parent, entity_type)
                            class: 'button disabled',
                            data: {rel: rel}),
                'data-tooltip' => data_tooltip)
+          para text_node I18n.t('pageflow.admin.user.entry_membership_hint')
+          if quota.present? && quota.exhausted?
```
DanielKuhn: quota is only defined in the `entry` case - in the user case this would crash

app/views/components/pageflow/admin/add_membership_button.rb
```diff
@@ -56,7 +62,7 @@ def entry_membership_possible_for_creator_and(user, parent)
         when User
           PotentialMemberships.creatable_by(user).entries_for_user(parent).any?
         when Entry
-          PotentialMemberships.creatable_by(user).users_for_entry(parent).any?
+          PotentialMemberships.creatable_by(user).users_for_entry(parent).without(user).any?
```
tf: This hides the button on the entry admin page if there is only one user in the account. But if the user visits the user admin page of their own user record, they can still add entries there. I think this is inconsistent.

On the other hand, if there are many users in the account and i want to assign users to entries to make clear who is responsible for what, why can't I add myself?

Also if all users of an account have been added to an entry, there is no need to display a message that you need to invite even more users.

I think, the important point is that "member management" does not make sense if there is only user in the account. So if I am alone in my account, I should not see the "Add user" button on the entry page. Instead there should be a text explaining what the membership concept is for.

In the context of Dedicated (where there are no quotas) or if my quota is not exhausted, this text can tell me to first invite users into my account before I can start managing memberships.

If my quota is exhausted, it can tell me to first go upgrade to use the feature. Since upgrading is only a concept in Hosted Pageflow, the partial inside of Pageflow can only include a generic message stating that the quota is exhausted. This can then be overridden in Hosted Pageflow to include a link to the upgrade page.

spec/views/components/pageflow/admin/add_membership_button_spec.rb
```diff
@@ -11,6 +11,55 @@ module Admin
...
+        entry = create(:entry, with_editor: user, title: 'Entry 1')
+        sign_in(user, scope: :user)
+
+        html = <<-HTML
+          <div class="custom_cannot_add_user_message">
+            Cannot add any more users.
+          </div>
+        HTML
+        template = {'pageflow/admin/entries/_cannot_add_user.html.erb' => html}
+        view.view_paths.unshift(ActionView::FixtureResolver.new(template))
```
tf: You  should be able to use `stub_template` instead.

app/views/components/pageflow/admin/add_membership_button.rb
awaismbajwa: So now it does not show add user button if users in account are less than 2.
In other cases when membership is not possible button gets disabled.

In both cases It checks if quota is exhausted and display quota exhausted message otherwise it displays invite more users message.

template pageflow/admin/entries/cannot_add_user can be overriden by host application

PR #1202, opened 2019-08-19
Update Twitter share URL

aviav:
Current URL only redirects to user's home page. It seems Twitter
currently offers a Web Intent API for our use case. See [1].

REDMINE-16999

[1] https://developer.twitter.com/en/docs/twitter-for-websites/tweet-button/guides/web-intent.html


aviav:
REDMINE-16999

PR #1211, opened 2019-09-12
added user member actions to impersonate and stop impersonate user

awaismbajwa:
REDMINE-16991

Added member actions which supports user impersonification.

remaining implementation should be done in the host application
based on the following gem.
https://github.com/ankane/pretender


awaismbajwa:
moved the changes to host application

PR #1232, opened 2019-10-31
Add active class color styling for mobile share panel

DanielKuhn:
REDMINE-17134

### DIFF
--- a/app/assets/stylesheets/pageflow/themes/default/mobile_navigation.scss
+++ b/app/assets/stylesheets/pageflow/themes/default/mobile_navigation.scss
@@ -100,5 +100,13 @@ $mobile-navigation-pictogram-visibility: $standard-page-pictogram-visibility !de
     svg {
       fill: $mobile-navigation-share-link-color;
     }
+
+    &.active {
+      color: $main-color;
+
+      svg {
+        fill: $main-color;
+      }
+    }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/pageflow/themes/default/mobile_navigation.scss
```diff
@@ -100,5 +100,13 @@ $mobile-navigation-pictogram-visibility: $standard-page-pictogram-visibility !de
     svg {
       fill: $mobile-navigation-share-link-color;
     }
+
+    &.active {
+      color: $main-color;
+
+      svg {
+        fill: $main-color;
```
tf: Shall we also introduce a new variable here to allow controlling this color individually like you did in the other PR?

DanielKuhn: Outline navigation bar only defines a handful of variables exactly for coloring, that's why I added the equivalent variables for sharing there.
In pageflow there's already so many SCSS-variables that I opted for using the main color instead of introducing yet more variables.

tf: I think SCSS-variables are a good thing actually. They are the public API of our theme.

PR #1248, opened 2019-12-02
Create Pageflow Scrolled Models

DanielKuhn:
REDMINE-17220


spec/pageflow/revision_component_spec.rb
```diff
@@ -12,23 +12,6 @@ module Pageflow
...
-
-      it 'differs for separate RevisonComponents' do
-        revision = create(:revision)
-        revision_component1 = TestRevisionComponent.create!(revision: revision)
-        revision_component2 = TestRevisionComponent.create!(revision: revision)
-
-        expect(revision_component1.perma_id).not_to eq(revision_component2.perma_id)
-      end
-    end
-
```
tf: I'd vouch for letting these specs stay since `RevisionComponent` is part of the public API and the perma id functionality is sort of part of it's contract. A little duplication in specs never hurts.

spec/models/concerns/pageflow/auto_generated_perma_id_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+        revision = create(:revision)
+        revision_component = TestRevisionComponent.create!(revision: revision)
+
+        expect {
+          revision_component.save!
+        }.not_to change { revision_component.perma_id }
+      end
+    end
+  end
+end
```
tf: I prefer the way `SerializedConfiguration` is tested below (with an anonymous class). From a super pessimistic point of view, these tests do not ensure that `AutoGeneratedPermaId` actually provides the functionality - only that `RevisionComponent` manages to provide it somehow.

spec/factories/sections.rb
```diff
@@ -0,0 +1,16 @@
...
+          transition: 'scroll',
+          fullHeight: true,
+          backdrop: {
+            image: 'some_image'
+          }
+        }
+      end
+    end
+  end
+end
```
tf: These factories should live in the scrolled entry type engine.

PR #1252, opened 2019-12-06
Decouple revision components from page types

aviav:
For entry types without the concept of pages, revision components are
needed as well. For this reason, revision components cannot remain
coupled to page types.

Page types can continue to define their own revision components,
however they are now registered globally.

REDMINE-17248 REDMINE-17243

### DIFF
--- a/config/initializers/revision_components.rb
+++ b/config/initializers/revision_components.rb
@@ -1,0 +1,7 @@
+Pageflow.after_configure do |config|
+  config.page_types.each do |page_type|
+    page_type.revision_components.each do |component|
+      config.revision_components.register(component)
+    end
+  end
+end
--- a/lib/pageflow/configuration.rb
+++ b/lib/pageflow/configuration.rb
@@ -104,6 +104,11 @@ module Pageflow
     # @return [FileTypes]
     attr_reader :file_types
 
+    # Used to register components whose current state must be
+    # persisted as part of a revision.
+    # @return [RevisionComponents]
+    attr_reader :revision_components
+
     # Used to register new types of widgets to be displayed in entries.
     # @return [WidgetTypes]
     attr_reader :widget_types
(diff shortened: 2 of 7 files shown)


lib/pageflow/configuration.rb
```diff
@@ -438,7 +439,7 @@ def paperclip_filesystem_root=(_val)
     end
 
     def revision_components
-      page_types.map(&:revision_components).flatten.uniq
+      @revision_components
     end
```
tf: Let's add `revision_components` to the `attr_reader`s at the top of the class. Let's also add a doc string since this will now be part of the public API.

lib/pageflow/revision_components.rb
```diff
@@ -0,0 +1,7 @@
+module Pageflow
+  class RevisionComponents < Array
```
tf: Inheriting from `Array` makes the collection mutable. Removing a revision component after the app has booted is not supported, though. Some of the other collection classes in this directory include `Enumerable` instead.

config/initializers/revision_components.rb
```diff
@@ -0,0 +1,7 @@
+Pageflow.after_configure do |config|
+  config.page_types.each do |page_type|
+    page_type.revision_components.each do |component|
+      config.revision_components.register(component)
+    end
+  end
+end
```
tf: [Page type help entry registration](https://github.com/codevise/pageflow/blob/master/config/initializers/help_entries.rb) had the same problem before, but now that we replicate the pattern it becomes more important: there is no easy way to test this code. We could consider moving all of this initializer code that applies configuration changes according to registered page types to a method in [`Pageflow::PageTypes`](https://github.com/codevise/pageflow/blob/master/lib/pageflow/page_types.rb) and call that from a single `after_configure` block in a single initializer. This would make testing easy and keep all of this logic in one place. This will come in handy, when we extract page types to the paged entry type engine at some point.

aviav:
See https://github.com/codevise/pageflow/pull/1256

PR #1253, opened 2019-12-06
Decouple file types from page types

aviav:
For entry types without the concept of pages, file types are needed as
well. For this reason, file types cannot remain coupled to page types.

Page types can continue to define their own file types, however they
are now registered globally.

REDMINE-17249 REDMINE-17243


aviav:
See https://github.com/codevise/pageflow/pull/1256

PR #1255, opened 2019-12-06
Let entry types provide frontend apps to render entries

tf:
Let entry types provide a Rack app that is then used to render previews and published entries. Authentication, authorization and redirecting can still be handled by the core controller.

See doc file in PR for details.

REDMINE-17208


spec/support/pageflow/entries_controller_test_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+    #   published revision or the draft.
+    def get_with_entry_env(action, entry:, mode: :published, params: {})
+      revision =
+        if mode == :published
+          entry.published_revision
+        else
+          entry.draft
+        end
+
+      published_entry = PublishedEntry.new(entry, revision)
```
DanielKuhn: How come we're always using `PublishedEntry` now, even when it's just a `draft` (i.e. `mode: :preview`)?

tf: That's also [the way it used to work before](https://github.com/codevise/pageflow/blob/master/app/controllers/pageflow/revisions_controller.rb#L13). `DraftEntry` has additional methods for editing entries (adding files, updating meta data). Whenever we are viewing an entry outside of the editor (whether it's a preview or the real thing), `PublishedEntry` is used.

PR #1258, opened 2019-12-09
File import feature flag

awaismbajwa:
REDMINE-17246

### DIFF
--- a/app/helpers/pageflow/config_helper.rb
+++ b/app/helpers/pageflow/config_helper.rb
@@ -2,8 +2,8 @@ module Pageflow
   module ConfigHelper
     include RenderJsonHelper
 
-    def editor_config_seeds
-      render_json_partial('pageflow/config/editor_seeds')
+    def editor_config_seeds(entry)
+      render_json_partial('pageflow/config/editor_seeds', entry: entry)
     end
   end
 end
--- a/app/views/pageflow/config/_editor_seeds.json.jbuilder
+++ b/app/views/pageflow/config/_editor_seeds.json.jbuilder
@@ -15,7 +15,8 @@ json.file_types Pageflow.config.file_types do |file_type|
     json.collection_name nested_file_type.collection_name
   end
 end
-json.file_importers Pageflow.config.file_importers do |file_importer|
+
+json.file_importers Pageflow.config_for(entry).file_importers do |file_importer|
   json.importer_name file_importer.name
   json.logo_source file_importer.logo_source
   json.authentication_provider file_importer.authentication_provider
--- a/spec/controllers/pageflow/editor/file_import_controller_spec.rb
+++ b/spec/controllers/pageflow/editor/file_import_controller_spec.rb
@@ -13,13 +13,36 @@ module Pageflow
         expect(response.status).to eq(302)
       end
 
-      it 'makes the call to the specified file importer' do
+      it 'does not invoke the disabled importer' do
         user = create(:user)
+        file_importer = TestFileImporter.new
         entry = create(:entry, with_editor: user)
         sign_in(user, scope: :user)
-        file_importer = FileImporterDummy.new
         pageflow_configure do |config|
-          config.file_importers.register(file_importer.name, file_importer)
+          config.features.register('test_file_importer') do |feature_config|
+            feature_config.file_importers.register(file_importer)
+          end
+        end
+
+        expect(file_importer).not_to receive(:search)
+        expect {
+          get(:search,
+              params: {query: 'test', file_import_name: file_importer.name, entry_id: entry})
+        }.to raise_error(RuntimeError, "Unknown file importer with name '#{file_importer.name}'.")
+      end
+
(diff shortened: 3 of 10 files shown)


app/views/pageflow/config/_editor_seeds.json.jbuilder
```diff
@@ -15,11 +15,14 @@ json.file_types Pageflow.config.file_types do |file_type|
...
-  json.importer_name file_importer.name
-  json.logo_source file_importer.logo_source
-  json.authentication_provider file_importer.authentication_provider
-  json.authentication_required file_importer.authentication_required current_user
+  if file_importer.feature_enabled?(@entry)
+    json.importer_name file_importer.name
+    json.logo_source file_importer.logo_source
+    json.authentication_provider file_importer.authentication_provider
+    json.authentication_required file_importer.authentication_required current_user
+  end
```
tf: File import does not need to know about the feature concept. If you use `Pageflow.config_for(entry)`, the importers registered inside a disabled feature will not be present. All of the methods like `feature_name`, `feature_enabled?`, `find_enabled_by_name` can then be removed.

app/helpers/pageflow/config_helper.rb
```diff
@@ -2,7 +2,7 @@ module Pageflow
   module ConfigHelper
     include RenderJsonHelper
 
-    def editor_config_seeds
+    def editor_config_seeds(_entry)
       render_json_partial('pageflow/config/editor_seeds')
     end
```
tf: If we add  `entry` as a parameter here, we should pass it the partial as a local instead of referencing the `@entry` variable there.

spec/controllers/pageflow/editor/file_import_controller_spec.rb
```diff
@@ -13,13 +13,36 @@ module Pageflow
...
+              params: {query: 'test', file_import_name: file_importer.name, entry_id: entry})
+        }.to raise_error
+      end
+
+      it 'makes the call to the specified file importer' do
+        user = create(:user)
+        file_importer = FileImporterHelper.new
+        entry = create(:entry,
+                       with_editor: user,
+                       features_configuration: {file_importer.feature_name => true})
```
tf: There is also a `with_feature` trait to make this easier.

spec/controllers/pageflow/editor/file_import_controller_spec.rb
```diff
@@ -13,13 +13,36 @@ module Pageflow
...
+          config.features.register(file_importer.feature_name) do |feature_config|
+            feature_config.file_importers.register(file_importer)
+          end
+        end
+
+        expect(file_importer).not_to receive(:search)
+        expect {
+          get(:search,
+              params: {query: 'test', file_import_name: file_importer.name, entry_id: entry})
+        }.to raise_error
```
tf: Always expect an explicit error when `raise_error`. This could be a syntax error that makes the spec pass...

spec/support/helpers/file_importer_helper.rb
```diff
@@ -1,9 +1,13 @@
 module Pageflow
-  class FileImporterDummy < Pageflow::FileImporter
+  class FileImporterHelper < Pageflow::FileImporter
```
tf: To fit the naming scheme of all other dummy implementations, I'd call this `TestFileImporter`.

spec/controllers/pageflow/editor/file_import_controller_spec.rb
```diff
@@ -152,12 +210,18 @@ module Pageflow
...
         pageflow_configure do |config|
-          config.file_importers.register(file_importer.name, file_importer)
+          config.features.register(file_importer.feature_name) do |feature_config|
+            feature_config.file_importers.register(file_importer)
+          end
         end
+
         selected_files = file_importer.search(nil, nil)['photos']
         meta_data = file_importer.files_meta_data(nil, selected_files)
         post(:start_import_job, params: {files: meta_data['files'],
```
tf: Especially for the action that starts the job and performs the actual import, I'd say it's important to also test that it is not possible if the feature is disabled. This only tests the happy path.

app/views/pageflow/config/_editor_seeds.json.jbuilder
tf: The spec testing that disabled importers are not included is still missing.

doc/creating_file_importers.md
```diff
@@ -99,6 +99,11 @@ implementation of the following methods:
 6. **logo_source** Return the source of the logo displayed in choose
    importer view.
 
+7. **feature_name** return the name of the feature against which 
+   file importer will be registered in the host application.
+   feature_name can be like `<file_importer_name>_file_importer`
+
+
```
tf: The import does not need to know about features. The Pexel's plugin uses this method when registering the feature, but also there is could just as well be hard coded in the plugin class.

awaismbajwa: sorry forgot to remove this

PR #1259, opened 2019-12-09
Defer creating filetypes

aviav:
Previously, FileType objects could be initialized using a model reference. This is now deprecated and will be removed in a future version. Please use a string that contains the model name instead

REDMINE-17243

### DIFF
--- a/lib/pageflow/file_types.rb
+++ b/lib/pageflow/file_types.rb
@@ -15,7 +15,7 @@ module Pageflow
     end
 
     def each(&block)
-      first_level_file_types = @file_types
+      first_level_file_types = trigger_lazy_loading_of_file_types
       lower_level_file_types = search_for_nested_file_types(first_level_file_types)
       (first_level_file_types + lower_level_file_types).uniq(&:model).each(&block)
     end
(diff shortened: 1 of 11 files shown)


lib/pageflow/file_types.rb
```diff
@@ -15,7 +15,16 @@ def clear
...
+      # To avoid dependency cycles due to autoloading for model
+      # references that are passed into FileType.new in some plugins,
+      # we allow registering as file types lambdas which resolve to a
+      # page type's file types. After some deprecation period, we
+      # could get rid of this and just set first_level_file_types to
+      # @file_types here.
+      file_types = @file_types.map do |file_type|
+        file_type.respond_to?(:call) ? file_type.call : file_type
+      end
+      first_level_file_types = file_types.flatten
```
tf: The lambda variant is tested indirectly [via `page_types_spec`](https://github.com/codevise/pageflow/blob/3e2b1ea7c743c8bf8734168b5cfaf193557d03ea/spec/pageflow/page_types_spec.rb#L92). I think the feature is special enough to warrant an explicit test case in `file_types_spec`. Let's also move this code block including the comment into a private method to name the concept and make this method easier to read.

tf: I find calling the method `file_types` a bit confusing since the whole class already is a collection of file types. I was rather thinking of a name that conveys additional information about what is happening. Something like `normalize_lazy_loaded_file_types` or `trigger_lazy_loading_of_file_types`. Also I'd move the `flatten` into the method since it is an implementation detail.

PR #1267, opened 2019-12-11
Extend config DSL for entry types

aviav:
REDMINE-17257

### DIFF
--- a/entry_types/paged/lib/pageflow_paged/plugin.rb
+++ b/entry_types/paged/lib/pageflow_paged/plugin.rb
@@ -4,5 +4,13 @@ module PageflowPaged
     def configure(config)
       config.entry_types.register(PageflowPaged.entry_type)
     end
+
+    class PagedConfiguration
+      include Pageflow::Configuration::EntryTypeConfiguration
+
+      def initialize(config)
+        super(config)
+      end
+    end
   end
 end
--- a/entry_types/scrolled/lib/pageflow_scrolled.rb
+++ b/entry_types/scrolled/lib/pageflow_scrolled.rb
@@ -10,7 +10,8 @@ module PageflowScrolled
     def entry_type
       Pageflow::EntryType.new(name: 'scrolled',
                               frontend_app: PageflowScrolled::EntriesController.action(:show),
-                              editor_fragment_renderer: editor_fragment_renderer)
+                              editor_fragment_renderer: editor_fragment_renderer,
+                              configuration: Plugin::ScrolledConfiguration)
     end
 
     private
--- a/lib/pageflow/configuration.rb
+++ b/lib/pageflow/configuration.rb
@@ -95,6 +95,9 @@ module Pageflow
     # @since 15.1
     attr_reader :entry_types
 
+    # @api private
+    attr_reader :entry_type_configs
+
     # Register new types of pages.
     # @return [PageTypes]
     # @since 0.9
@@ -373,6 +376,7 @@ module Pageflow
       @quotas = Quotas.new
       @themes = Themes.new
       @entry_types = EntryTypes.new
+      @entry_type_configs = {}
       @page_types = PageTypes.new
       @file_types = FileTypes.new
       @widget_types = WidgetTypes.new
@@ -444,6 +448,14 @@ module Pageflow
       ActiveSupport::Deprecation.warn('Pageflow::Configuration#paperclip_filesystem_root is deprecated.', caller)
     end
 
+    def for_entry_type(type, &_block)
+      type_name = type.name
+      entry_type_configs[type_name] ||= entry_types.find_by_name!(type_name).configuration.new(self)
+      config = entry_type_configs[type_name]
+
+      yield(config)
+    end
+
     # @api private
     def lint!
       @features.lint!
@@ -475,6 +487,12 @@ module Pageflow
       delegate :admin_attributes_table_rows, to: :config
       delegate :themes, to: :config
       delegate :file_importers, to: :config
+    end
+
+    module EntryTypeConfiguration
+      def initialize(config)
+        @config = config
+      end
     end
   end
 end
--- a/lib/pageflow/entry_type.rb
+++ b/lib/pageflow/entry_type.rb
@@ -4,7 +4,7 @@ module Pageflow
   # @since 15.1
   class EntryType
     # @api private
-    attr_reader :name, :frontend_app, :editor_fragment_renderer
+    attr_reader :name, :frontend_app, :editor_fragment_renderer, :configuration
 
     # @param name [String] A unique name.
     #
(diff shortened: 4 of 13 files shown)


entry_types/scrolled/lib/pageflow_scrolled.rb
```diff
@@ -7,10 +7,11 @@ def plugin
       PageflowScrolled::Plugin.new
     end
 
-    def entry_type
+    def entry_type(config)
       Pageflow::EntryType.new(name: 'scrolled',
                               frontend_app: PageflowScrolled::EntriesController.action(:show),
-                              editor_fragment_renderer: editor_fragment_renderer)
+                              editor_fragment_renderer: editor_fragment_renderer,
+                              configuration: Plugin::ScrolledConfiguration.new(config))
```
tf: I think a lot less changes like this would be necessary if we decided to pass a configuration class instead of a configuration object to `EntryType.new`. We could require these classes to have an initializer which takes a `config` as its only parameter. `for_entry_type` below could then instantiate that class and keep it in local hash which maps entry type names to entry type configuration objects the way it was in a previous iteration of this PR.

lib/pageflow/configuration.rb
```diff
@@ -476,5 +488,27 @@ class FeatureLevelConfiguration < Struct.new(:config)
       delegate :themes, to: :config
       delegate :file_importers, to: :config
     end
+
+    module EntryTypeConfiguration
+      def assign_config(value)
+        @config = value
+      end
```
tf: Renaming this to `initialize` and having configuration classes call `super(config)` would be more idiomatic, I think.

lib/pageflow/configuration.rb
```diff
@@ -476,5 +488,27 @@ class FeatureLevelConfiguration < Struct.new(:config)
...
+        if @config.respond_to?(method)
+          config.send(method, *args)
+        else
+          super
+        end
+      end
+
+      def respond_to_missing?(method_name, include_private = false)
+        @config.respond_to?(method_name) || super
+      end
```
tf: I'd suggest to directly start with a restricted interface here, which only delegates selected methods to the main configuration. `config_for` then needs to return the delegator we talked about, which delegates missing methods to the entry type specific config.

entry_types/paged/lib/pageflow_paged/plugin.rb
```diff
@@ -2,7 +2,18 @@ module PageflowPaged
...
   class Plugin < Pageflow::Plugin
     def configure(config)
-      config.entry_types.register(PageflowPaged.entry_type)
+      config.entry_types.register(PageflowPaged.entry_type(config))
+    end
+
+    class PagedConfiguration
+      include Pageflow::Configuration::EntryTypeConfiguration
+
+      attr_accessor :page_types
```
tf: Doesn't this mean that there are now two page type collections - one in the main config and one here? Calling `config.page_types.register` outside of an `for_entry` block would add items to a different collection than calling it inside of the block. I think, the specs currently only pass since `Pageflow.config_for` is often called with a `PublishedEntry`/`DraftEntry` instead of an `Entry` which both do not delegate `type_name` at the moment. This causes `Pageflow.config_for` below to not return an entry type specific config.

I'd suggest to postpone this change to a later step when we also add the compat layer.

lib/pageflow/entry_type.rb
```diff
@@ -4,18 +4,19 @@ module Pageflow
...
+    attr_reader :name, :frontend_app, :editor_fragment_renderer, :configuration
 
     # @param name [String] A unique name.
     #
     # @param frontend_app [#call] A rack app that renders the entry
     #   when not in the editor (i.e. preview and published entries).
     #
     # @param editor_fragment_renderer
-    def initialize(name:, frontend_app:, editor_fragment_renderer:)
+    def initialize(name:, frontend_app:, editor_fragment_renderer:, configuration:)
```
tf: Parameter not yet documented in inline doc. Let's also add a section to the ["Creating entry types" guide](https://github.com/codevise/pageflow/blob/master/doc/creating_entry_types.md) describing this feature.

lib/pageflow/global_config_api.rb
```diff
@@ -115,5 +120,29 @@ def build_config(&block)
...
+          config.send(method, *args)
+        elsif type_config.respond_to?(method)
+          type_config.send(method, *args)
+        else
+          super
+        end
+      end
+
+      def respond_to_missing?(method_name, include_private = false)
+        @config.respond_to?(method_name) || super
```
tf: Doesn't this need to consider `type_config` as well?

lib/pageflow/global_config_api.rb
```diff
@@ -44,8 +44,13 @@ def configure(&block)
...
     def config_for(target)
-      build_config do |config|
-        config.enable_features(target.enabled_feature_names)
+      config = build_config do |c|
+        c.enable_features(target.enabled_feature_names)
+      end
+      if target.respond_to?(:type_name) && config.entry_type_configs.key?(target.type_name)
+        ConfigDelegator.new(config, target.type_name)
+      else
+        config
```
tf: The type of the return value of `config_for` should not depend on whether `for_entry_type` was called. Otherwise it's not safe to use entry type specific methods. This could also be spec.

lib/pageflow/configuration.rb
```diff
@@ -476,5 +490,19 @@ class FeatureLevelConfiguration < Struct.new(:config)
...
     end
+
+    module EntryTypeConfiguration
+      def initialize(config)
+        @config = config
+      end
+
+      def new(*args)
+        initialize(*args)
+      end
```
tf: Why is this needed?

lib/pageflow/configuration.rb
```diff
@@ -476,5 +490,19 @@ class FeatureLevelConfiguration < Struct.new(:config)
...
+
+      def config
+        @config ||= Configuration.new
+      end
```
tf: Why is this needed?

lib/pageflow/configuration.rb
```diff
@@ -444,6 +450,14 @@ def paperclip_filesystem_root=(_val)
       ActiveSupport::Deprecation.warn('Pageflow::Configuration#paperclip_filesystem_root is deprecated.', caller)
     end
 
+    def for_entry_type(type, &_block)
+      type_name = type.name
+      config = entry_types.find_by_name!(type_name).configuration.new(self)
+      entry_type_configs[type_name] = config
+
+      yield(config)
+    end
```
tf: Calling `for_entry_type` a second time overwrites a previously created config. This case should be covered by a spec.

lib/pageflow/configuration.rb
```diff
@@ -95,6 +95,11 @@ class Configuration
     # @since 15.1
     attr_reader :entry_types
 
+    # Holds configurations for entry types.
+    # @return Hash
+    # @since 15.1
+    attr_accessor :entry_type_configs
+
```
tf: This can be a `attr_reader`. Let's also make it it `@api private` since it should not be used by outside code.

spec/support/helpers/test_entry_type.rb
```diff
@@ -5,10 +5,20 @@ def self.register(config, options = {})
     end
 
     def self.new(options = {})
-      EntryType.new(name: 'test',
+      EntryType.new(name: options[:name] || 'test',
```
tf: The `**options` below already allows overriding the name.

PR #1269, opened 2019-12-11
minor file import lint change

awaismbajwa:
REDMINE-17246


awaismbajwa:
Spec for TestFileImport is added

PR #1270, opened 2019-12-13
Add scrolled model serialization

DanielKuhn:
REDMINE-17259


entry_types/scrolled/spec/support/helpers/public_json_helper.rb
```diff
@@ -0,0 +1,10 @@
+require 'pageflow/json_test_helper'
+
+module JsonHelper
+  include Pageflow::JsonTestHelper
+end
+
+RSpec.configure do |config|
+  config.include(JsonHelper, type: :controller)
+  config.include(JsonHelper, type: :helper)
+end
```
tf: No need to define a new module. `config.include(Pageflow::JsonTestHelper, type: :controller)` would do the trick.

PR #1272, opened 2019-12-16
Remove obsolete member_label usage

aviav:
Formtastic 3.1 deprecated the member label option. This commit removes
the deprecation notice from the test log by addressing it.

See [1].

REDMINE-17257, REDMINE-17277

[1] https://github.com/justinfrench/formtastic/blob/master/DEPRECATIONS#L14

### DIFF
--- a/admins/pageflow/accounts.rb
+++ b/admins/pageflow/accounts.rb
@@ -71,6 +71,7 @@ module Pageflow
     end
 
     controller do
+      helper Pageflow::Admin::AccountsHelper
       helper Pageflow::Admin::FeaturesHelper
       helper Pageflow::Admin::FormHelper
       helper Pageflow::Admin::LocalesHelper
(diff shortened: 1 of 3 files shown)


lib/pageflow/configuration.rb
```diff
@@ -465,6 +465,13 @@ def enable_all_features
       features.enable_all(FeatureLevelConfiguration.new(self))
     end
 
+    # @api private
+    def share_providers_input_collection
+      available_share_providers.map do |provider|
+        [provider.to_s.camelize, provider]
+      end
+    end
+
```
tf: Let's move this to an admin helper. There are already other helpers that provide collections like this (e.g. [`Pageflow::Adimin::WidgetsHelper#admin_widget_types_collection_for_role`](https://github.com/codevise/pageflow/blob/a56c86ba90b987fe3a83a495b658540ba52ac433/app/helpers/pageflow/admin/widgets_helper.rb#L11)).

tf (changes requested):

tf (approved):

PR #1276, opened 2019-12-17
Improve support for Webpack setup in host appplication

tf:
* Update `bin/build-edge` script to include all build outputs

* Turn pageflow into peer dependency of scrolled package

* Provide Webpack config fragment in pageflow package

* Improve handling of svg files in Rollup build

REDMINE-17274

### DIFF
--- a/packages/pageflow/config/webpack.js
+++ b/packages/pageflow/config/webpack.js
@@ -1,0 +1,15 @@
+module.exports = {
+  externals: {
+    'backbone': 'Backbone',
+    'backbone.babysitter': 'Backbone.ChildViewContainer',
+    'cocktail': 'Cocktail',
+    'jquery': 'jQuery',
+    'jquery-ui': 'jQuery',
+    'jquery.minicolors': 'jQuery',
+    'underscore': '_',
+    'backbone.marionette': 'Backbone.Marionette',
+    'i18n-js': 'I18n',
+    'iscroll': 'IScroll',
+    'wysihtml5': 'wysihtml5'
+  }
+};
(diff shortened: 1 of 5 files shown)


packages/pageflow/config/webpack.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  externals: {
+    'backbone': 'Backbone',
+    'backbone.babysitter': 'Backbone.ChildViewContainer',
+    'cocktail': 'Cocktail',
+    'jquery': 'jQuery',
+    'jquery-ui': 'jQuery',
```
DanielKuhn: is this a typo or why is jQuery exported under three different keys?
Wouldn't this have to be `jquery-ui` and `jquery.minicolors` respectively?

tf: The thing is that neither `jquery-ui` nor `jquery.minicolors` define something like an export. Importing them only has the side effect of adding more stuff to `$`/`jQuery`. In development/production there is no need to import them at all since these legacy dependencies still come from the [asset pipeline](https://github.com/codevise/pageflow/blob/master/app/assets/javascripts/pageflow/editor/vendor.js). In Jest test there are no asset pipeline files, though. There these dependencies need to be imported to extend the jQuery global. Jest [maps these imports](https://github.com/codevise/pageflow/blob/3089735f5cf26b05aeee997b0a7b3d4bcf546f35/packages/pageflow/config/jest/index.js#L7) to its own copies of these files in the `packages/pageflow/src/vendor` directory.

So the above mapping is mainly a way to make these imports no-op for development/production. Somewhere down the line it would be nice to use a single source for these libraries. But for now this is the compromise to make Jest tests run without having to migrate everything away from the asset pipeline in one step.

PR #1278, opened 2019-12-18
Use seed data

DanielKuhn:
REDMINE-17260


entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
```diff
@@ -15,6 +15,10 @@
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
 
+    <script>
+      pageflowScrolledSeed = <%= scrolled_entry_json_seed(@entry) %>;
+    </script>
+
```
tf: This currently makes some editor specs for scrolled fail, since [the template is also rendered as an HTML seed](https://github.com/codevise/pageflow/blob/d3c62f8fddce1c0f2d919c5d4a2a87e40e2facf9/entry_types/scrolled/app/helpers/pageflow_scrolled/editor/seed_html_helper.rb#L6) for the editor iframe. `@entry` is not set in this context. Maybe we can default to setting `pageflowScrolledSeed` to an empty array in that case. Then we can tackle displaying real data in the editor as a separate step.

entry_types/scrolled/lib/pageflow_scrolled/seeds.rb
```diff
@@ -25,6 +25,11 @@ def sample_scrolled_entry(attributes)
                                         **attributes.except(:sections)) do |created_entry|
           created_entry.theming = attributes.fetch(:account).default_theming
 
+          features_configuration = created_entry.features_configuration
+                                                .merge('page_types_with_file_types' => false,
+                                                       'page_type.linkmap_page' => false)
+          created_entry.features_configuration = features_configuration
+
```
tf: Let's put this in the host application since it relies on a special temporary setup we have created there.

PR #1280, opened 2019-12-19
Updated active admin to 2.5.0

awaismbajwa:
REDMINE-17241

- All coffee scripts are removed and should not be added again. if it is required then coffee-rails gem should be included.
- Active admin load_paths are added using the setter `+=` instead of `unshift` method as a workaround to activeadmin initialisation order [issue](https://github.com/activeadmin/activeadmin/issues/5995)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,7 @@ group :development do
 end
 
 # Required to make imports in Active Admin stylesheet work
-gem 'sassc-rails'
+gem 'sassc-rails', '~> 2.1'
 
 # Required for XML serialization in Active Admin
 gem 'activemodel-serializers-xml'
@@ -34,3 +34,7 @@ gem 'webpacker'
 
 # Make tests fail on JS errors
 gem 'capybara-chromedriver-logger', git: 'https://github.com/codevise/capybara-chromedriver-logger', branch: 'do-not-raise-on-filtered-errors', require: false
+
+# temporary fix, so that we dont have to upgrade ruby version
+gem 'inherited_resources', '~> 1.9.0'
+gem 'responders', '~> 2.4.1'
--- a/Gemfile
+++ b/Gemfile
@@ -20,9 +20,6 @@ group :development do
   gem 'listen'
 end
 
-# Required to make imports in Active Admin stylesheet work
-gem 'sassc-rails', '~> 2.1'
-
 # Required for XML serialization in Active Admin
 gem 'activemodel-serializers-xml'
 
@@ -34,7 +31,3 @@ gem 'webpacker'
 
 # Make tests fail on JS errors
 gem 'capybara-chromedriver-logger', git: 'https://github.com/codevise/capybara-chromedriver-logger', branch: 'do-not-raise-on-filtered-errors', require: false
-
-# temporary fix, so that we dont have to upgrade ruby version
-gem 'inherited_resources', '~> 1.9.0'
-gem 'responders', '~> 2.4.1'
--- a/app/assets/stylesheets/pageflow/admin/embedded_index_table.scss
+++ b/app/assets/stylesheets/pageflow/admin/embedded_index_table.scss
@@ -1,6 +1,6 @@
 .embedded_index_table {
   .sortable a {
-    background: url("/assets/active_admin/orderable.png") no-repeat 100% 2px;
+    background: url($orderable-icon-url) no-repeat 100% 2px;
     padding-right: 13px;
   }
 
--- a/app/assets/stylesheets/pageflow/admin/embedded_index_table.scss
+++ b/app/assets/stylesheets/pageflow/admin/embedded_index_table.scss
@@ -1,3 +1,7 @@
+//since activeadmin 2.5.0 depends on ruby 2.4
+//this is a temporary fix to use activeadmin < 2.5 for supporting ruby 2.3
+$orderable-icon-url: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAcAAABGCAYAAAAAVo4aAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAH5JREFUeNpi3LhlOwMU1AExGxDXwARYoHQLEFdD2cxAXAliMKFJgEAFEHfBJEHGMKLhMpgkTsAEdch/NNwCk2xCdiEQtML4LEgCf6EubUX3Cgh0oNvJ+P//f7wOGpUclRwYSZb41CyidNbB8giNM+9oXhmVHHm5bJjUSAABBgDKKiwMMUxPwgAAAABJRU5ErkJggg==" !default;
+
 .embedded_index_table {
   .sortable a {
     background: url($orderable-icon-url) no-repeat 100% 2px;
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -23,10 +23,16 @@ group :development do
 end
 
 # Required to make imports in Active Admin stylesheet work
-gem 'sassc-rails', '~> 1.0'
+gem 'sassc-rails', '~> 2.1'
```
tf: I removed the version here in a recent change since it is already specified in the gemspec. Bundler outputs a warning if you specify the version in both places.

Gemfile
```diff
@@ -23,10 +23,16 @@ group :development do
...
 
 # Required for XML serialization in Active Admin
 gem 'activemodel-serializers-xml'
 
+# Make webpacker available in specs. Host applications that want to
+# use webpacker need to add it to their Gemfile themselves. Requiring
+# webpacker in an engine file (like we normally do) would force all
+# host application to install webpacker.
+gem 'webpacker'
+
```
tf: This is now on master. The commit on this branch should be gone after the next rebase.

app/assets/stylesheets/pageflow/admin/embedded_index_table.scss
```diff
@@ -1,6 +1,10 @@
+// since activeadmin 2.5.0 depends on ruby 2.4
+// this is a temporary fix to use both ruby 2.3 and ruby 2.4
```
tf: This is more about compatibility with ActiveAdmin < 2.5 (or whatever version introduced the icon urls) than about Ruby version compatibility.

pageflow.gemspec
```diff
@@ -21,7 +21,11 @@ Gem::Specification.new do |s|
   s.add_dependency 'rails', '~> 5.2.0'
 
   # Framework for admin interface
-  s.add_dependency 'activeadmin', '~> 1.3.0'
+  if RUBY_VERSION >= '2.4'
+    s.add_dependency 'activeadmin', '~> 2.5.0'
+  else
+    s.add_dependency 'activeadmin', '~> 2.0.0'
+  end
```
tf: The gemspec is evaluated at gem release time. So this has no effect when the gem is used in another project and makes the declared dependencies depend on the Ruby version on the machine of the developer who released the gem.

There is no reason to _require_ an Active Admin version above 2.5 in the Ruby 2.4 case if our gem works with Active Admin 2.0 anyway. Looking at the changes in the PR, I'd suspect that we could even go with `s.add_dependency 'activeadmin', ['>= 1.3.0', '< 3']`.

spec/support/pageflow/dummy/rails_template.rb
```diff
@@ -53,7 +53,14 @@ def source_paths
...
-  ActiveAdmin.application.load_paths.unshift(Dir[Rails.root.join('app/admin')].first)\n
+  ActiveAdmin.application.load_paths.unshift(Dir[Rails.root.join('admins')].first)\n
+END
+
+# add active admin patch to load pageflow activeadmin resources
+append_to_file('config/initializers/pageflow.rb', <<-END)
+  ActiveAdmin.after_load do |app|
+    Dir.glob("admins/**/*").each { |f| load(f) if File.file?(f) }
+  end\n
 END
```
tf: Why does [adding to `ActiveAdmin.application.load_paths`](https://github.com/codevise/pageflow/blob/0e9684524eaa46a757cbf35ec97efbdab14d7d35/lib/generators/pageflow/active_admin_initializer/active_admin_initializer_generator.rb#L10) no longer suffice?

Gemfile
awaismbajwa: I have removed this from Gemfile, why do we need to mention it in the Gemfile?

pageflow.gemspec
awaismbajwa: yes it works

spec/support/pageflow/dummy/rails_template.rb
awaismbajwa: activeadmin specifies new way to add load_paths like this
`config.load_paths = [File.join(Rails.root, "app", "ui")]`
I tried it before with config.load_paths.unshift but somehow it didn't work, now I tried again for debugging and it works fine now, I am changing the source according to this config load_paths now

awaismbajwa: for using unshift it works when i initialize load_paths with empty array like this
config.load_paths = []
If I dont initialize and print load_paths I get this
.rvm/gems/ruby-2.4.5/bundler/gems/cert_watch-a7cdd94ff724/admin
.rvm/gems/ruby-2.4.5/bundler/gems/pageflow-d2f784fe368b/admins
dedicated_pageflow/app/admin

If I initialize with empty array and then print, I get this
.rvm/gems/ruby-2.4.5/bundler/gems/cert_watch-a7cdd94ff724/admin
.rvm/gems/ruby-2.4.5/bundler/gems/pageflow-d2f784fe368b/admins

 it was not working because  dedicated_pageflow/app/admin does not exist

now that i initialize empty load_paths it works even with ActiveAdmin.application.load_paths
ActiveAdmin.application.load_paths = []

so now I can keep previous version and initialize load_paths with empty array in order to remove dedicated_pageflow/app/admin

awaismbajwa: this works in hosted application but this is not working in pageflow dummy app so for now I am keeping it like this in pageflow and trying to understand what is causing it in pageflow dummy app

Gemfile
tf: Mentioning it in the Gemfile causes the gem to be required when the dummy application boots. At one point this was needed to keep ActiveAdmin from raising an exception in Pageflow gem tests. Maybe this problem went away?

awaismbajwa: I tested bin/rspec after removing this and executed host application. I did not get any error

tf: Then let's remove it.

spec/support/pageflow/dummy/rails_template.rb
tf: To me it looks like [this commit](https://github.com/activeadmin/activeadmin/commit/2459474c6868aaab9b417fb14bd73345f6714f3d) added an initializer which runs after the `active_admin` initializer in the dummy app and resets the load path. If I change our code to 

```
Rails.application.config.after_initialize do
  ActiveAdmin.application.load_paths.unshift(Pageflow.active_admin_load_path)
end
```

the correct admin files are loaded again in pageflow specs and Dedicated. Not sure this is a stable solution, though. I think this is a bug in ActiveAdmin and we should probably file an issue.

I did not see a case yet where references to a non existing directory in the load paths caused an error. What kind of error message did you see?

awaismbajwa: There was no error about non existing directory, I was just suspecting it because in our host application merge request it loads the files when I initialize load_paths to empty array.

Should I make changes according to this?
`Rails.application.config.after_initialize do
  ActiveAdmin.application.load_paths.unshift(Pageflow.active_admin_load_path)
end`

tf: Can you open an issue in the ActiveAdmin repository first and see if we get some feedback? I think this should be fixed in ActiveAdmin and we are not in a hurry to update at the moment. So if possible I would like to avoid working around the bug like that.

awaismbajwa: sure

PR #1282, opened 2019-12-19
Move editor actions into editor controller

tf:
* Move pageflow/entries#partial to pageflow_paged/editor/entries

* Rename entries#edit action to editor/entries#show

REDMINE-17267


config/routes.rb
```diff
@@ -1,6 +1,6 @@
 Pageflow::Engine.routes.draw do
   constraints Pageflow.config(:ignore_not_configured => true).editor_route_constraint do
-    resources :entries, only: [:edit], shallow: true do
+    resources :entries, only: [], shallow: true do
```
DanielKuhn: don't we still need `:show` here?

PR #1283, opened 2019-12-19
Allow scoping file types and widget types by entry types

aviav:
REDMINE-17257

### DIFF
--- a/lib/pageflow/global_config_api.rb
+++ b/lib/pageflow/global_config_api.rb
@@ -46,12 +46,14 @@ module Pageflow
     def config_for(target)
       config = build_config do |c|
         c.enable_features(target.enabled_feature_names)
+        c.configure_entry_type(target.type_name) if target.respond_to?(:type_name)
       end
+
       if target.respond_to?(:type_name)
-        ConfigDelegator.new(config, target.type_name)
-      else
-        config
+        config = Configuration::ConfigView.new(config, target.type_name)
       end
+
+      config
     end
 
     # Register a block which shall be called after any configuration
@@ -94,6 +96,7 @@ module Pageflow
 
       @config = build_config do |config|
         config.enable_all_features
+        config.configure_all_entry_types
       end
 
       @after_global_configure_blocks ||= []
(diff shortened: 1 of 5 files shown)


lib/pageflow/global_config_api.rb
```diff
@@ -48,10 +48,11 @@ def config_for(target)
         c.enable_features(target.enabled_feature_names)
       end
       if target.respond_to?(:type_name)
-        ConfigDelegator.new(config, target.type_name)
-      else
-        config
+        config = ConfigDelegator.new(config, target.type_name)
+        config.configure_blocks.each { |block| block.call(config) }
```
tf: This causes entry type configure blocks to be executed after the `after_configure` blocks, which would break integrations like the ones we've built to register file types etc for page types. I think the entry type specific configure blocks need be evaluated during `build_config` at the same time when `enable_feature` is called above.

Let's cover this case with a spec.

lib/pageflow/global_config_api.rb
```diff
@@ -46,12 +46,12 @@ def configure(&block)
     def config_for(target)
       config = build_config do |c|
         c.enable_features(target.enabled_feature_names)
+        call_entry_type_config_blocks(c, target) if target.respond_to?(:type_name)
```
tf: We still need to handle the case, when we want to know all file types in the system ([example](https://github.com/codevise/pageflow/blob/master/lib/pageflow/ability_mixin.rb#L179)). To make it safe to enable configure blocks of multiple entry types, I guess, we will need to make [`FileTypes`](https://github.com/codevise/pageflow/blob/0fa72a514f620f169737a4eb908630fd8d0064da/lib/pageflow/file_types.rb#L10) resilient against duplicate registrations.

lib/pageflow/global_config_api.rb
```diff
@@ -121,6 +121,14 @@ def build_config(&block)
       end
     end
 
+    def call_entry_type_config_blocks(config, target)
+      type_name = target.type_name
+      delegator = ConfigDelegator.new(config, type_name)
+      delegator.entry_type_configure_blocks[type_name].each do |block|
+        block.call(delegator)
+      end
+    end
```
tf: This makes all config options available inside `for_entry_type` blocks, which we don't want. There could be a test which makes sure that things like `config.mailer_sender` are not available in these blocks.

lib/pageflow/global_config_api.rb
aviav: I'd argue that because `FileTypes`' `each` uses `uniq(&:model)` (see [1]), `FileTypes` doesn't need any more resilience against duplicate registrations

[1] https://github.com/codevise/pageflow/blob/0fa72a514f620f169737a4eb908630fd8d0064da/lib/pageflow/file_types.rb#L20

PR #1288, opened 2019-12-20
Scrolled model controllers

DanielKuhn:
REDMINE-17266


entry_types/scrolled/app/controllers/pageflow_scrolled/editor/chapters_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+        render json: chapter
+      end
+
+      def order
+        entry = Pageflow::DraftEntry.find(params[:entry_id])
+        storyline = Storyline.all_for_revision(entry.draft).first
+
+        params.require(:ids).each_with_index do |id, index|
+          storyline.chapters.update(id, position: index)
```
tf: The `order` actions of the other controllers already allow moving content elements/sections between arbitrary sections/chapters within the storyline. Let's also allow moving chapters between storylines here and extend the other `order` actions to also allow moving content elements/sections between storylines.

entry_types/scrolled/app/models/pageflow_scrolled/chapter.rb
```diff
@@ -3,6 +3,30 @@ class Chapter < Pageflow::ApplicationRecord
...
+    def ensure_storyline
+      unless Storyline.all_for_revision(revision).exists?
+        Storyline.create!(revision: revision, configuration: {main: true})
+      end
+      self.storyline = Storyline.all_for_revision(revision).first
+    end
+
+    def set_initial_position
+      self.position = storyline.chapters.count
+    end
```
tf: Instead of setting the position here, I'd suggest following the same pattern as in the paged REST controllers and add `position` to the list of permitted params. That way the client side code can [determine the initial position](https://github.com/codevise/pageflow/blob/d5fd99fa09503bc3aa3d9326b242e7723109cfa1/packages/pageflow/src/editor/models/Storyline.js#L92). In the current editor, the sitemap uses this [to insert chapters](https://github.com/codevise/pageflow-sitemap/blob/a5907142794564b21d3149c907d8b1a27cf847e7/app/assets/javascripts/pageflow/sitemap/editor/controllers/editor_mode_controller.js#L252) and pages at arbitrary positions.

entry_types/scrolled/config/routes.rb
```diff
@@ -1,5 +1,21 @@
...
+      collection do
+        put :order
+      end
+    end
+
+    resources :content_elements do
+      collection do
+        put :order
+      end
+    end
```
tf: Let's use shallow nested resources to get urls of the form `POST .../sections/5/content_elements` instead of `POST .../content_elements?section_id=5`.

entry_types/scrolled/app/controllers/pageflow_scrolled/editor/content_elements_controller.rb
```diff
@@ -0,0 +1,51 @@
+module PageflowScrolled
```
tf: See `ChaptersController`.

entry_types/scrolled/app/controllers/pageflow_scrolled/editor/sections_controller.rb
```diff
@@ -3,8 +3,51 @@ module Editor
     # @api private
```
tf: See `ChaptersController`.

entry_types/scrolled/app/models/pageflow_scrolled/chapter.rb
```diff
@@ -3,6 +3,30 @@ class Chapter < Pageflow::ApplicationRecord
...
 
-    has_many :sections, -> { order('position ASC') }, dependent: :destroy, inverse_of: :chapter
+    belongs_to :storyline, touch: true
+    has_many :sections,
+             -> { order('pageflow_scrolled_sections.position ASC') },
+             dependent: :destroy,
+             inverse_of: :chapter
+    has_many :content_elements, through: :sections
+
+    delegate :entry, to: :storyline
```
tf: If I'm not missing something, these delegates are currently only used in tests. In the paged `Chapter` and `Storyline` models they were added as sort of last resort when first the `Revision` model was inserted between `Chapter` and `Entry` and later the `Storyline` model was placed between `Chapter` and `Revsion` to keep calls of the form `chapter.entry` working. In general I'd consider them an anti-pattern, though, since they lead to a series of DB queries which are often hard to eager load to prevent N+1 situations.

In the tests we can use transient factory attributes ([example](https://github.com/codevise/pageflow/blob/d5fd99fa09503bc3aa3d9326b242e7723109cfa1/spec/factories/pages.rb#L13), [example](https://github.com/codevise/pageflow/blob/d5fd99fa09503bc3aa3d9326b242e7723109cfa1/spec/factories/chapters.rb#L11)) to accomplish the same thing.

entry_types/scrolled/spec/controllers/pageflow_scrolled/editor/content_elements_controller_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+        put(:order,
+            params: {
+              entry_id: entry.id,
+              section_id: section.id,
+              ids: [content_elements.first.id, content_elements.last.id]
+            }, format: 'json')
+
+        expect(content_elements.first.reload.position).to eq(0)
+        expect(content_elements.last.reload.position).to eq(1)
+      end
```
tf: Given that authorization is one of the main controller concerns, I find it a good rule of thumb to always test controllers along the line of what is permitted and what is forbidden: What are the minimal permissions required to perform an action successfully. What are the maximal permissions not yet sufficient to perform it. For simple actions that can be authenticated vs not authenticated or authorized vs not authorized. For actions with more complex params (like this one), I try to come up with cases that combine representative inputs from different classes, e.g. permitted content element ids but section id of other entry etc.

entry_types/scrolled/app/controllers/pageflow_scrolled/editor/content_elements_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  module Editor
+    # @api private
+    class ContentElementsController < ActionController::Base
+      include Pageflow::EditorController
+
+      def create
+        section = Section.find(params[:section_id])
```
tf: For content elements and sections there is no test ensuring that it's not possible to create models in a different entry. At least here it is thus still possible.

entry_types/scrolled/app/models/pageflow_scrolled/storyline.rb
```diff
@@ -0,0 +1,30 @@
...
+             dependent: :destroy,
+             inverse_of: :storyline
+    has_many :sections,
+             -> { reorder(SECTIONS_ORDER) },
+             through: :chapters
+    has_many :content_elements,
+             -> { reorder(CONTENT_ELEMENTS_ORDER) },
+             through: :sections
+
+    delegate :entry, to: :revision
```
tf: Unused?

entry_types/scrolled/config/routes.rb
tf: Since the `shallow` option is missing, this creates routes of the form `.../sections/:section_id/content_elements/:id` for update and delete actions. The section id is not used, though. Backbone routing will be easier if we only need to work with one level hierarchy. 

`sections` can then also be nested `chapters`.

entry_types/scrolled/spec/controllers/pageflow_scrolled/editor/chapters_controller_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+        expect {
+          patch(:update,
+                params: {
+                  entry_id: entry,
+                  id: chapter_in_other_entry,
+                  chapter: {
+                    configuration: {title: 'another title'}
+                  }
+                }, format: 'json')
+        }.to raise_error(ActiveRecord::RecordNotFound)
```
tf: This means a 500 response which is not what we exepct. `Pageflow::ApplicationController` had a `rescue_from` to turn all `ActiveRecord::RecordNotFound` errors into 404s. This sometimes lead to hard to track down bugs if an unexpected exception was silently converted into a 404. `Pageflow::EditorController` therefore rescues these exceptions explicitly at places where they are expected. I'd suggest to the same in the actions here.

entry_types/scrolled/spec/controllers/pageflow_scrolled/entries_controller_spec.rb
```diff
@@ -11,6 +11,7 @@ module PageflowScrolled
     describe '#show' do
       it 'renders script tag for frontend js' do
         entry = create(:entry, :published)
+        create(:scrolled_chapter, revision: entry.published_revision)
```
tf: This looks like it might either be not needed or hiding a code path that cannot handle entries without chapters.

entry_types/scrolled/spec/helpers/pageflow_scrolled/entry_json_seed_helper_spec.rb
```diff
@@ -27,8 +28,9 @@ def render(helper, entry)
 
       it 'orders sections according to position attribute' do
         entry = create(:published_entry)
-        section1 = create(:section, revision: entry.revision, position: 4)
-        section2 = create(:section, revision: entry.revision, position: 3)
+        chapter = create(:scrolled_chapter, revision: entry.revision)
+        section1 = create(:section, chapter: chapter, position: 4)
+        section2 = create(:section, chapter: chapter, position: 3)
```
tf: We also need to make sure now that sections respect the order of their parent chapters.

entry_types/scrolled/spec/controllers/pageflow_scrolled/entries_controller_spec.rb
DanielKuhn: This was added to fix the linting spec for new entries. I removed it in favor of building a placeholder storyline in https://github.com/codevise/pageflow/pull/1288/commits/3dbbf34a19651a4b21c8be2ee507b8d492410620#diff-4f20f4b7f92dc9e9012f90518a9b90edR20

entry_types/scrolled/spec/helpers/pageflow_scrolled/entry_json_seed_helper_spec.rb
DanielKuhn: Good catch. I adjusted the code and specs in https://github.com/codevise/pageflow/pull/1288/commits/3dbbf34a19651a4b21c8be2ee507b8d492410620 , following the same pattern as in Pageflow, where the serialization order is maintained by scoping through the entry down to pages. Here I do the same thing but scoping through the main storyline which defines the order of its child models.

PR #1292, opened 2019-12-23
Render scrolled seed data in editor

tf:
* Let entry types define entry Backbone model

* Render scrolled entry json seed in editor seed

* Set up post message communcation in editor

* Extend RenderJsonHelper

REDMINE-17260


packages/pageflow/src/editor/api/index.js
```diff
@@ -71,15 +71,38 @@ export const EditorApi = Object.extend(
...
      * @alias fileImporters
-     *  
      */
     this.fileImporters = new FileImporters();
   },
 
+  /**
+   * Configure editor for entry type.
+   *
+   * @param {string} name
```
DanielKuhn: currently the `name` doesn't seem to go anywhere?

tf: Yes. Since there is only ever one entry type defined in the editor js at the moment, we don't need the name currently. I left it in to align the API with the other register methods and also in case we need it down the line.

PR #1297, opened 2020-01-06
Render camel case in responses

DanielKuhn:
REDMINE-17304


DanielKuhn:
changes will be included in #1288

PR #1298, opened 2020-01-08
Use chapters from db

DanielKuhn:
REDMINE-17300


tf:
Can you extract the Scene/Section commit to a separate PR, so I can merge it and rebase my changes?

DanielKuhn:
> Can you extract the Scene/Section commit to a separate PR, so I can merge it and rebase my changes?

I did the refactoring after introducing chapters into the hierarchy, which would lead to merge conflicts. But since the specs are fixed: Maybe we could just merge this and I'll add the navigation bar changes on another branch?

PR #1300, opened 2020-01-09
Use chapter data in navigation

DanielKuhn:
REDMINE-17300

### DIFF
--- a/entry_types/scrolled/lib/pageflow_scrolled/seeds.rb
+++ b/entry_types/scrolled/lib/pageflow_scrolled/seeds.rb
@@ -9,10 +9,13 @@ module PageflowScrolled
     # @param [Hash] attributes  attributes to override defaults
     # @option attributes [Account] :account  required
     # @option attributes [String] :title  required
-    # @option attributes [Array] :sections  required
-    #   An array of section configurations, each containing a key "foreground"
+    # @option attributes [Array] :chapters  required
+    #   An array of chapter configurations, each containing a key "sections"
+    #   which lists the separate sections of each chapter.
+    #   Each section has a "foreground"-key
     #   under which an array of content_element configurations is stored.
-    #   Each content_element configuration must provide a "type" and "props".
+    #   Each content_element configuration must provide a "type"-attribute
+    #   to determine the React component used to render this content element.
     # @yield [entry] a block to be called before the entry is saved
     # @return [Entry] newly created entry
     def sample_scrolled_entry(attributes)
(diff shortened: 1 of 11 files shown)


entry_types/scrolled/lib/pageflow_scrolled/seeds.rb
```diff
@@ -53,21 +50,34 @@ def say_creating_scrolled_entry(entry)
...
       end
     end
 
-    def create_content_element(section, content_element_config, position)
+    def create_content_element(section, content_element_config)
       section.content_elements.create!(
         type_name: content_element_config['type'],
-        configuration: content_element_config['props'].except('anchor'),
-        position: position
+        configuration: content_element_config['props']
```
tf: Why is position no longer required?

entry_types/scrolled/package/src/frontend/Chapter.js
```diff
@@ -4,7 +4,7 @@ import Section from './Section';
 
 export default function Chapter(props) {
   return (
-    <div id={'chapter' + props.permaId}>
+    <div id={'chapter-'+props.permaId}>
```
tf: Just as a aside: A template string, i.e. `` `chapter-${props.permaId}` ``, would be more modern/idiomatic, I think

entry_types/scrolled/package/src/frontend/navigation/AppHeader.js
```diff
@@ -51,11 +58,27 @@ export default function AppHeader() {
...
+    setActiveChapterLink(chapterLinkId);
     setMobileNavHidden(true);
   };
 
+  function renderMenuLinks(chapters) {
+    return chapters.map((chapter, index) => {
+      const chapterIndex = index + 1;
+      const chapterLinkId = 'chapterLink' + chapterIndex;
+      return (
+        <li key={index} className={styles.chapterListItem}>
```
tf: Also just as aside: Whenever possible it's best to use keys that represent stable identity. React uses keys to determine whether an element has been inserted. If you used the chapter id as key, for example, and then a chapter gets inserted after the first one, React can actually reuse the DOM nodes of all chapters and just inserts one new DOM node. Using an index as a key causes React to to view the inserted chapter as the new way the second chapter now looks. This causes all following chapter nodes to be updated, basically updating the nth chapter with the contents of the (n-1)th.

DanielKuhn: 👍

PR #1301, opened 2020-01-10
Extract part of appearance options into paged entry type

aviav:
REDMINE-17297


tf:
Let's go with the version implemented in #1304.

PR #1305, opened 2020-01-13
Register scrolled storyline as revision component

DanielKuhn:
Deep copy all chapters, sections and content elements of the storyline
to new revisions.

REDMINE-17315


DanielKuhn:
Added storyline spec covering copying to new revision.

PR #1319, opened 2020-01-27
Sharing, imprint, privacy in navigation

DanielKuhn:
Add tooltips for info (imprint, privacy, file rights) and sharing (with configured share providers)

REDMINE-17313
REDMINE-17314


app/helpers/pageflow/config_helper.rb
```diff
@@ -27,5 +27,36 @@ def config_file_model_types_seed(json, config)
...
+      end
+      json.copyright do
+        json.copyright_link_label raw(theming.copyright_link_label)
+        json.copyright_link_url theming.copyright_link_url
+      end
+      json.privacy do
+        json.privacy_link_label I18n.t('pageflow.public.privacy_notice')
+        json.privacy_link_url "#{theming.privacy_link_url}?lang=#{entry.locale}"
+      end
+    end
```
tf: I'm on the fence if this really carries its weight as a separate helper. Since this will only be used by the scrolled entry seed helper, it might as well be part of the JSON template. In it's current state it also duplicates most of the logic in `Pageflow::EntriesHelper#entry_global_links`. So one could also consider instead only extracting a helper for the privacy link url to prevent building that in two places. Is therea reason why the copyright link label is passed `raw` while imprint link label isn't?

Also, helpers used across engine boundaries should have a spec.

entry_types/scrolled/package/src/entryState/legalInfo.js
```diff
@@ -0,0 +1,81 @@
...
+ * fileRights // => "author of image 1, author of image 2"
+ */
+export function useFileRights() {
+  const entryState = useEntryState();
+
+  const defaultFileRights = entryState.config.defaultFileRights;
+  const imageFiles = getItems(entryState.collections, 'imageFiles');
+  const imageFileRights = imageFiles.map(function (imageConfig) {
+    return imageConfig.rights ? imageConfig.rights.trim() : undefined;
+  }).filter(Boolean).join(', ');
```
tf: The original Ruby helper included file rights for other file types - even if it's called "Bildrechte" in the UI.

entry_types/scrolled/package/src/entryState/legalInfo.js
```diff
@@ -0,0 +1,81 @@
...
+  const fileRights = !!imageFileRights ? imageFileRights : defaultFileRights.trim();
+  const fileRightsString = !!fileRights ? ('Bildrechte: ' + fileRights) : '';
```
tf: Eventually, we will have to use I18n here. This already might be a problem in different places of the scrolled frontend already?

entry_types/scrolled/package/src/entryState/legalInfo.js
```diff
@@ -0,0 +1,81 @@
...
+ * all images used in the entry.
+ * If none of the images has a rights attribute configured,
+ * it falls back to the default file rights of the entry's account,
+ * otherwise returns an empty string
+ *
+ * @example
+ *
+ * const fileRights = useFileRights();
+ * fileRights // => "author of image 1, author of image 2"
+ */
```
tf: To make these docs show up in the correct category, you need to add items to the [scrolled `documentation.yml` file](https://github.com/codevise/pageflow/blob/30b53db563d2fc1743e932a33b5e6ece56e58903/entry_types/scrolled/package/documentation.yml#L11)

entry_types/scrolled/package/src/entryState/legalInfo.js
```diff
@@ -0,0 +1,81 @@
...
+  const imageFiles = getItems(entryState.collections, 'imageFiles');
+  const imageFileRights = imageFiles.map(function (imageConfig) {
+    return imageConfig.rights ? imageConfig.rights.trim() : undefined;
+  }).filter(Boolean).join(', ');
+  const fileRights = !!imageFileRights ? imageFileRights : defaultFileRights.trim();
+  const fileRightsString = !!fileRights ? ('Bildrechte: ' + fileRights) : '';
+
+  return useMemo(() => {
+    return fileRightsString;
+  }, [imageFileRights]);
```
tf: Just as an aside, this `useMemo` does nothing since the dependency, `imageFileRights`, is a new object every time (constructed by the `map` call a few lines up). It also doesn't hurt, though.

entry_types/scrolled/package/src/frontend/navigation/ChapterLinkTooltip.js
```diff
@@ -0,0 +1,23 @@
...
+                  type='light'
+                  place='bottom'
+                  effect='solid'
+                  className={classNames(headerStyles.navigationTooltip,
+                                        styles.chapterLinkTooltip)}>
+      <div>
+        <h3 className={styles.tooltipHeadline}>Kapitel {props.chapterIndex}</h3>
+        <p>
+          {props.summary}
+        </p>
```
tf: Summary can include HTML. I had [this commit](https://github.com/tf/pageflow/commit/c0c724c2b475c836f864d541658833f59537466a) on my outline branch, which I removed now since it was the only conflict between our branches and was somewhat unrelated.

entry_types/scrolled/package/src/frontend/navigation/LegalInfoLink.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export function LegalInfoLink(props) {
+  return (
+    <div>
+      {props.label && props.url &&
+      <a target="_blank" href={props.url}>
+        {props.label}
+      </a>
```
tf: Label can include HTML entities (i.e. Copyright sign).

entry_types/scrolled/package/src/entryState/legalInfo.js
DanielKuhn: Ok, I added a note to add those later.

entry_types/scrolled/package/src/entryState/legalInfo.js
DanielKuhn: True, at least in the nav bar there's already "Kapitel" hardcoded also.
I'll add a ticket for I18n in Redmine.

PR #1321, opened 2020-01-30
Re-add help entries for appearance options

aviav:
REDMINE-17297

### DIFF
--- a/doc/creating_entry_types.md
+++ b/doc/creating_entry_types.md
@@ -175,8 +175,20 @@ export const RainbowEntry = Entry.extend({
 });
 ```
 
-The preview and outline views need to be Backbone views, that will
-reveive the entry via the `model` option.
+The preview and outline views need to be Backbone views that will
+receive the entry via the `model` option.
+
+Pageflow expects translation keys that are scoped to entry type. For
+example, if we add an input to appearance inputs like this
+
+``` javascript
+const appearanceInputs = (tabView, _options) => {
+  tabView.input('manual_start', CheckBoxInputView);
+};
+```
+
+Pageflow is looking for translation keys of the form
+`pageflow.entry_types.paged.editor.entry_metadata_configuration_attributes.manual_start.{inline_help,label}`.
 
 ### REST Controllers
 
--- a/packages/pageflow/spec/editor/views/EditMetaDataView-spec.js
+++ b/packages/pageflow/spec/editor/views/EditMetaDataView-spec.js
@@ -5,17 +5,20 @@ import {EditMetaDataView, editor} from 'pageflow/editor';
 import {CheckBoxInputView} from 'pageflow/ui';
 
 import {factories} from '$support';
+import * as support from '$support';
 import {ConfigurationEditor} from '$support/dominos/ui';
 
 describe('EditMetaDataView', () => {
   it('renders general tab', () => {
-    const entry = factories.entry();
+    editor.registerEntryType('test');
+    const entry = factories.entry({}, {type_name: 'test'});
     const view = new EditMetaDataView({
       model: entry,
       tab: 'general',
       state: {
         config: {}
-      }
+      },
+      editor: editor
     });
 
     view.render();
@@ -30,7 +33,7 @@ describe('EditMetaDataView', () => {
   it('renders entry type specific appearance inputs on widgets tab', () => {
     const entry = factories.entry();
     editor.registerEntryType('test', {
-      appearanceInputs: (the_view, the_entry, theming) => {
+      appearanceInputs: (the_view, _options) => {
         the_view.input('cheese', CheckBoxInputView);
       }
     });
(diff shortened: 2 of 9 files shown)


packages/pageflow/src/editor/views/EditMetaDataView.js
```diff
@@ -30,10 +30,12 @@ export const EditMetaDataView = Marionette.Layout.extend({
     var state = this.options.state || {};
     var features = this.options.features || {};
     var editor = this.options.editor || {};
+    var entryTypeName = (editor.entryType && editor.entryType.name) || 'none';
 
     var configurationEditor = new ConfigurationEditorView({
       model: entry.metadata.configuration,
-      tab: this.options.tab
+      tab: this.options.tab,
+      attributeTranslationKeyPrefixes: ['pageflow.entry_types.' + entryTypeName + '.pageflow/entry_metadata_configuration']
```
tf: Key segments of the form `pageflow/...` are used by rails when generating translation keys for namespaced classes like `Pageflow::Page`. Here we are free to choose whatever name we want and are therefore not forced to repeat the word `pageflow` later in the key.

Since this key format is part of Pageflow's public API, I'd try to make it as consistent as possible with other already existing keys that serve similar purposes. Examples from the code base:

1. `'pageflow.editor.files.attributes.' + this.options.fileType.collectionName` - in `EditFileFile.js`.
2. `'pageflow.editor.widgets.attributes.' + this.name` - in `WidgetType.js` used in `EditWidgetView.js`
3. `this.seed.translation_key_prefix + '.page_attributes'` in `PageType.js` used in `EditPageView.js`
4.  `pageflow.storyline_attributes` in `EditStorylineView.js`

Since all existing usages include the word `attributes` in the key, I'd suggest doing the same here.

As discussed, the third variant (`pageflow.before_after.page_attributes`) is problematic since it risks name collisions between page type names and other existing keys of the form `pageflow.*`.

The last two variants also do not make it clear that these are translations used in the editor.

Combining these ideas I'd suggest to either go with:

* `pageflow.editor.entry_metadata_configuration_attributes.<type_name>` - pro: most similar to widgets/files case
* `pageflow.entry_types.<type_name>.editor.entry_metadata_configuration_attributes` - pro: all keys for an entry type live under a common prefix

Whatever you choose, please extend the example in `creating_entry_types.md` to show an example function adding an input and the corresponding translation key that needs to be defined.

Earlier, we talked shortly about adding a spec. This would be nice, but I'd be ok with skipping it here in the interest of getting things merged.

packages/pageflow/src/editor/views/EditMetaDataView.js
```diff
@@ -30,10 +30,12 @@ export const EditMetaDataView = Marionette.Layout.extend({
     var state = this.options.state || {};
     var features = this.options.features || {};
     var editor = this.options.editor || {};
+    var entryTypeName = (editor.entryType && editor.entryType.name) || 'none';
```
tf: `editor.entryType` is always present in "real life". If this is here to make specs work, I'd suggest to instead register an entry type in the affected spec. We already do this in some places, when views depend on entry types.

packages/pageflow/src/editor/views/EditMetaDataView.js
aviav: I'll go with the second suggestion. As opposed to widgets, entry types are much more likely to later be extracted from the gem. This justifies the different treatment, I'd say

doc/creating_entry_types.md
```diff
@@ -175,8 +175,20 @@ export const RainbowEntry = Entry.extend({
...
-The preview and outline views need to be Backbone views, that will
-reveive the entry via the `model` option.
+The preview and outline views need to be Backbone views that will
+receive the entry via the `model` option.
+
+Pageflow expects translation keys that are scoped to entry type. For
+example, if we add an input to appearance inputs like this
+
+``` javascript
+const appearanceInputs = (tabView, _entry, _theming) => {
```
tf:
```suggestion
const appearanceInputs = (tabView, _options) => {
```

doc/creating_entry_types.md
```diff
@@ -175,8 +175,20 @@ export const RainbowEntry = Entry.extend({
...
+  tabView.input('manual_start', CheckBoxInputView);
+};
+```
+
+Pageflow is looking for translation keys of the form
+`pageflow.entry_types.paged.editor.entry_attributes.manual_start.{inline_help,label}`.
```
tf:
```suggestion
`pageflow.entry_types.paged.editor.entry_metadata_configuration_attributes.manual_start.{inline_help,label}`.
```

packages/pageflow/spec/editor/views/EditMetaDataView-spec.js
```diff
@@ -49,6 +52,35 @@ describe('EditMetaDataView', () => {
...
 
+  support.useFakeTranslations({
+    'pageflow.entry_types.strange.editor.entry_metadata_configuration_attributes.quark.label': 'Up',
+    'pageflow.entry_types.strange.editor.entry_metadata_configuration_attributes.quark.inline_help': 'Help yourself!'
+  });
+
+  it('uses entry type-specific translation keys if provided', () => {
+    const entry = factories.entry();
+    editor.registerEntryType('strange', {
+      appearanceInputs: (the_view, the_entry, theming) => {
```
tf:
```suggestion
      appearanceInputs: (the_view, _options) => {
```

aviav:
I think I covered them all now

PR #1328, opened 2020-02-11
Imported translations from rails to scrolled entry js environment

awaismbajwa:
REDMINE-17375

### DIFF
--- a/entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
+++ b/entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
@@ -16,7 +16,7 @@
     <div id="root"></div>
 
     <%= javascript_pack_tag 'pageflow-scrolled-frontend' %>
-
+    <script> I18n.locale = "<%= @entry.locale %>"; </script>
     <%= scrolled_entry_json_seed_script_tag(@entry, @seed_options || {}) %>
   </body>
 </html>
--- a/entry_types/scrolled/package/src/frontend/index.js
+++ b/entry_types/scrolled/package/src/frontend/index.js
@@ -20,7 +20,12 @@ export * from './SectionThumbnail';
 export {default as Entry} from './Entry';
 export {EntryStateProvider} from '../entryState/EntryStateProvider'
 
+import I18n from 'i18n-js';
+window.I18n = I18n;
+
 window.pageflowScrolledRender = function(seed) {
+  I18n.translations = I18n.translations || {};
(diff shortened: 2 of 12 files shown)


entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
```diff
@@ -18,5 +18,7 @@
     <%= scrolled_entry_json_seed_script_tag(@entry, @seed_options || {}) %>
 
     <%= javascript_pack_tag 'pageflow-scrolled-frontend' %>
+    <%= public_i18n_javascript_tag(@entry) %>
+
```
tf: This includes all paged public translations. Let's keep scrolled translations in a separate scope (e.g. `pageflow_scrolled.public.*`) and extend the helper make the prefix configurable.

entry_types/scrolled/package/src/editor/index.js
```diff
@@ -12,6 +12,9 @@ import * as globalInterop from 'pageflow/editor';
 
 Object.assign(pageflow, globalInterop);
 
+import i18n from 'i18n-js';
+window.I18n = i18n;
+
```
tf: This should not be needed since in the editor `I18n` is automatically setup to include all translations via the `i18n-js` gem.

entry_types/scrolled/package/src/frontend/SoundDisclaimer.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import i18n from 'i18n-js';
```
tf: All other places import i18n with a capital first letter. Let's stick with that.

entry_types/scrolled/package/src/frontend/index.js
```diff
@@ -7,6 +7,9 @@ import Entry from './Entry';
 import './global.module.css';
 import {EntryStateProvider} from '../entryState';
 
+import i18n from 'i18n-js';
+window.I18n = i18n;
+
```
tf: The [latest changes on master](https://github.com/codevise/pageflow/blob/108c2269343792cb630e199d07ab8ff29de113c9/entry_types/scrolled/app/helpers/pageflow_scrolled/entry_json_seed_helper.rb#L18) introduce a single global function called   `pageflowScrolledRender`. If we use this function to pass the public translations, we don not need to export the `I18n` object on window.

tf:
We also need to update the `pageflow-public-i18n` version requirement once that gem is released.

tf:
see #1342

PR #1336, opened 2020-02-17
Add videoEmbed content element

DanielKuhn:
REDMINE-17386


entry_types/scrolled/package/src/frontend/ContentElement.module.css
```diff
@@ -79,5 +78,6 @@
 .selected:hover:after,
 .selected:after {
   border: solid 1px currentColor;
-  opacity: 0.8
+  opacity: 0.8;
+  pointer-events: none;
```
tf: This change makes the [specs fail](https://travis-ci.org/codevise/pageflow/jobs/651980712#L2795) since [the click](https://github.com/codevise/pageflow/blob/da1b7e220d4f71ddb2ca70d3f536e62be2450705/entry_types/scrolled/spec/features/entry_editor/editing_a_content_element_spec.rb#L19) on the content element is intercepted by the pseudo element.

entry_types/scrolled/package/src/contentElements/videoEmbed/VideoEmbed.js
```diff
@@ -0,0 +1,35 @@
...
+import styles from './VideoEmbed.module.css';
+import classNames from 'classnames';
+import {InlineCaption} from "../inlineImage/InlineCaption";
+
+export function VideoEmbed({configuration}) {
+  return (
+    <div className={styles.VideoEmbed}>
+      <div className={classNames(styles.embedWrapper, styles[configuration.aspectRatio])}>
+        <ReactPlayer className={styles.embedPlayer}
+                     key={btoa(JSON.stringify(configuration))}
```
tf: Why is this needed? If it is needed, I'd suggest constructing a less opaque string from the content element id etc.

entry_types/scrolled/package/src/contentElements/videoEmbed/VideoEmbed.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import ReactPlayer from 'react-player';
+import styles from './VideoEmbed.module.css';
+import classNames from 'classnames';
+import {InlineCaption} from "../inlineImage/InlineCaption";
```
tf: This is fine for now, but generally I'd be careful to introduce too many inter-dependencies between content elements. Maybe a variation of the rule of three: Copy to new content element on first reuse. Extract to `pageflow-scrolled/frontend` (and test and document) on second reuse.

entry_types/scrolled/package/src/contentElements/videoEmbed/VideoEmbed.js
DanielKuhn: Changing the url reloads the player component but changing options doesn't (by default), so `hideControls` doesn't get picked up.
I took the idea from [here](https://github.com/CookPete/react-player/issues/209).
Since the element only has two options currently I could construct the key from those, but when adding new options later one would certainly forget to add them to the key, that's why I constructed the key from the whole configuration object to make sure the key changes when options change.

entry_types/scrolled/package/src/frontend/ContentElement.module.css
DanielKuhn: I saw that too.
The behavior is as desired, but I need to come up with an idea for a test-compliant change.

PR #1345, opened 2020-02-26
Entry state integration for video files

DanielKuhn:
REDMINE-17371


entry_types/scrolled/package/spec/frontend/Video-spec.js
```diff
@@ -1,15 +1,14 @@
...
 import {render} from '@testing-library/react'
 
 import {Video} from 'frontend/Video';
 
-describe('Image', () => {
+describe('Video', () => {
   it('renders', () => {
     const {getByRole} =
-      render(<Video id="videoBoatDark" />)
+      render(<Video id="TODO" />)
```
tf: CI fails since `render` does not provide the needed React context. Use `renderInEntry` instead.

entry_types/scrolled/package/src/contentElements/inlineVideo/stories.js
```diff
@@ -1,10 +1,10 @@
 import '../frontend';
-import {storiesOfContentElement} from 'pageflow-scrolled/spec/support/stories';
+import {storiesOfContentElement, filePermaId} from 'pageflow-scrolled/spec/support/stories';
 
 storiesOfContentElement(module, {
   typeName: 'inlineVideo',
   baseConfiguration: {
-    id: 'videoGarzweilerDrohne',
+    id: filePermaId('videoFiles', 'interview_toni'),
```
tf: This does not work since the seed entry [created for storybook](https://github.com/codevise/pageflow/blob/84c6f7904065fbd5cc46036440a42aab542284f6/entry_types/scrolled/lib/tasks/pageflow_scrolled_tasks.rake#L39) does not include the video file. If it did, we would still have no way to transcode the video in CI, since the seed code now implicitly relies on development mode Resque workers which are not present in CI. Maybe it's ok to render inline videos with unset video files for now?

entry_types/scrolled/package/src/frontend/Video.js
```diff
@@ -6,95 +6,70 @@ import MutedContext from './MutedContext';
...
-  const poster_videoKaterchen  = awsBucket+'posterframes/poster_katerchen.jpeg';
-  const videoGarzweilerLoop1  = awsBucket+'braunkohle_loop1.mp4';
-  const poster_videoGarzweilerLoop1  = awsBucket+'posterframes/poster_braunkohle_loop1.jpeg';
-  const videoGarzweilerLoop2  = awsBucket+'braunkohle_loop2.mp4';
-  const poster_videoGarzweilerLoop2  = awsBucket+'posterframes/poster_braunkohle_loop2.jpeg';
-  const videoGarzweilerDrohne  = awsBucket+'braunkohle_drone.mp4';
-  const poster_videoGarzweilerDrohne  = awsBucket+'posterframes/poster_braunkohle_drone.jpeg';
-  const videoInselInterviewToni  = awsBucket+'pageflow_insel_interview_toni02.mp4';
-  const poster_videoInselInterviewToni  = awsBucket+'posterframes/poster_pageflow_insel_interview_toni02.jpg';
+import {useFile} from '../entryState';
```
tf: See https://redmine.codevise.de/issues/17301#note-19

entry_types/scrolled/package/src/frontend/Video.js
```diff
@@ -6,95 +6,70 @@ import MutedContext from './MutedContext';
...
+ * @param {Object} props
+ * @param {number} props.id - Perma id of the video file.
+ */
+export function Video(props) {
+  const videoFile = useFile({collectionName: 'videoFiles', permaId: props.id});
 
-  const videoRef = useRef();
-  const state = props.state;
+  if (videoFile) {
+    const videoRef = useRef();
```
tf: The lint warnings are right. Hooks can not be used conditionally. This also leads to errors in the browser consoles. You could extract a component that uses the hooks instead and render _it_ conditionally.

entry_types/scrolled/package/src/editor/models/SectionConfiguration.js
```diff
@@ -18,13 +18,20 @@ export const SectionConfiguration = Configuration.extend({
              (this.attributes.backdrop &&
               this.attributes.backdrop.image.toString().startsWith('#') ? 'color' : 'image');
```
tf: This line throws an exception if I try to open the section settings of the "Einsatz von Medien" section. `backdrop` is set, but, `backdrop.image` is undefined (only `backdrop.video` is present). Calling `toString` on `undefined` causes the error.

entry_types/scrolled/lib/pageflow_scrolled/seeds.rb
```diff
@@ -58,18 +68,27 @@ def say_creating_scrolled_entry(entry)
...
+    def create_files(draft_entry, file_type, file_data_by_name)
+      file_data_by_name.transform_values do |data|
+        say("     creating #{file_type} file from #{data['url']}")
+
+        file = draft_entry.create_file!(Pageflow::BuiltInFileType.send(file_type),
+                                        state: file_type == :image ? 'processed' : 'uploading',
+                                        attachment: URI.parse(data['url']),
+                                        configuration: data['configuration'])
+        if file_type == :video
+          if draft_entry.entry_title == 'Storybook seed'
```
tf: If this was an option like @skip_encoding@, it could be set by the rake task without needing to hard code a name. It would also be easier to test then.

entry_types/scrolled/package/src/frontend/Video.js
DanielKuhn: Refactored to component

entry_types/scrolled/package/src/editor/models/SectionConfiguration.js
DanielKuhn: Fixed

entry_types/scrolled/lib/pageflow_scrolled/seeds.rb
DanielKuhn: Option added. Refactored attributes to keyword argument and added options as separate argument

entry_types/scrolled/package/src/frontend/VideoPlayer.js
```diff
@@ -0,0 +1,72 @@
...
+
+import styles from './VideoPlayer.module.css';
+
+/**
+ * Render a player for a given video file.
+ *
+ * @param {Object} props
+ * @param {number} props.videoFile - The file to be played
+ */
+export function VideoPlayer(props) {
```
tf, line 15: This component is an implementation detail of the `Video` component. I would have probably left it as a private function inside the same file, but that might be a matter of taste.

Adding inline docs, though, makes it appear in the API doc even though it is not exported in `frontend/index.js` and can thus not be used outside of the `frontend` module.

entry_types/scrolled/package/src/frontend/Video.js
```diff
@@ -1,103 +1,19 @@
...
+import React from 'react';
+import {useFile} from '../entryState';
+import {VideoPlayer} from './VideoPlayer';
+
+/**
+ * Render a video file.
+ *
+ * @param {Object} props
+ * @param {number} props.id - Perma id of the video file.
+ */
```
tf, line 10: Needs to be added to `documentation.yml` to make it appear at the right position in the TOC of the API doc.

tf:
Also a small merge conflict now.

PR #1350, opened 2020-02-28
Browser compatability

awaismbajwa:
REDMINE-17369


tf:
* autoprefixer change has been extracted to #1391
* Intersection observer polyfill has been installed from npm package in #1389 
* Firefox bug could be reproduced and is now tracked in REDMINE-17596. The proposed fix changes the reveal effect to a scroll in effect, though.

PR #1351, opened 2020-03-02
Add before/after content element

aviav:
REDMINE-17372


entry_types/scrolled/package/src/contentElements/inlineBeforeAfter/BeforeAfter.js
```diff
@@ -1,99 +1,107 @@
...
+    <div ref={beforeAfterRef}
+      className={cx({[styles.selected]: isSelected, [styles.wiggle]: wiggle},
+      styles.container)}>
+      <InitialSliderPositionIndicator parentSelected={isSelected}
+        position={initial_slider_position}/>
+      {/* onSliderPositionChange: Prevent wiggle if user uses slider */}
+      <ReactCompareImage leftImage={beforeImageUrl} rightImage={afterImageUrl}
+                         leftImageLabel={before_label} rightImageLabel={after_label}
+                         leftImageAlt={beforeImageAlt} rightImageAlt={afterImageAlt}
+                         sliderPositionPercentage={initial_slider_position/100}
```
DanielKuhn: Since you're already calculating it up top, you could use `initialSliderPos` here.

entry_types/scrolled/package/src/frontend/index.js
```diff
@@ -25,6 +27,7 @@ export * from './i18n';
 export * from './SectionThumbnail';
 export {default as Entry} from './Entry';
 export {EntryStateProvider} from '../entryState/EntryStateProvider'
+export {useEditorSelection} from './EditorState';
```
DanielKuhn: Why is this needed here? Unused?

entry_types/scrolled/package/src/frontend/index.js
```diff
@@ -14,6 +14,8 @@ const editMode = window.location.pathname.indexOf('/editor/entries') === 0;
 export {api as frontend} from './api';
 
 export * from './Image';
+export * from './useImageUrl';
+export * from './useImageAlt';
```
DanielKuhn: These two need are only needed until `useFile` is fixed for usage in content elements. But for now they can stay here.

entry_types/scrolled/package/src/frontend/index.js
aviav: We need to export this here to be able to use it in BeforeAfter.js

PR #1356, opened 2020-03-11
Hide logo on page level

DanielKuhn:
REDMINE-17436

### DIFF
--- a/packages/pageflow/src/frontend/page_types/mixins/common_page_css_classes.js
+++ b/packages/pageflow/src/frontend/page_types/mixins/common_page_css_classes.js
@@ -5,6 +5,7 @@ pageflow.commonPageCssClasses = {
   updateCommonPageCssClasses: function(pageElement, configuration) {
     pageElement.toggleClass('invert', configuration.get('invert'));
     pageElement.toggleClass('hide_title', configuration.get('hide_title'));
+    pageElement.toggleClass('hide_logo', !!configuration.get('hide_logo'));
 
     toggleModeClass(pageflow.Page.textPositions, 'text_position');
     toggleModeClass(pageflow.Page.delayedTextFadeIn, 'delayed_text_fade_in');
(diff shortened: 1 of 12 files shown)


packages/pageflow-react/src/pages/createPageType.jsx
```diff
@@ -47,6 +48,11 @@ export default function({Component,
         id: pageId(pageElement),
         position: options.position
       }));
+      if(!!configuration.hide_logo) {
+        jQuery('.header_logo').stop( true, true ).fadeOut(1000);
+      } else {
+        jQuery('.header_logo').stop( true, true ).fadeIn(1000);
+      };
```
tf: Doing it here, will only make it work for React based pages - not the legacy ones. 

What we are trying to achieve is very similar to toggling between the inverted/normal logo variant based on the inverted setting of the page. You already found the `updateCommonPageCssClasses` function which does this for the background image logo. The watermark logo is part of the header, which has its own jQuery widget `packages/pageflow/src/frontend/jquery_ui_widgets/header.js` to [handle such logic](https://github.com/codevise/pageflow/blob/9872a63d8e7b527b4cfaaa6eaeb8019fc090b3d3/packages/pageflow/src/frontend/jquery_ui_widgets/header.js#L37). The corresponding CSS for the header lives [here](https://github.com/codevise/pageflow/blob/9872a63d8e7b527b4cfaaa6eaeb8019fc090b3d3/app/assets/stylesheets/pageflow/themes/default/logo/variant/watermark.scss#L100), where we can also place a new CSS class that hides the logo instead of using jQuery animations.

packages/pageflow/src/frontend/page_types/mixins/common_page_css_classes.js
```diff
@@ -5,6 +5,7 @@ pageflow.commonPageCssClasses = {
   updateCommonPageCssClasses: function(pageElement, configuration) {
     pageElement.toggleClass('invert', configuration.get('invert'));
     pageElement.toggleClass('hide_title', configuration.get('hide_title'));
+    pageElement.toggleClass('hide_logo', configuration.get('hide_logo'));
```
tf: `toggleClass` has the unpleasant behavior to toggle the class on every call if the second parameter is `undefined` instead of `false`. The two lines above, therefore depend on the fact that `invert` and `hide_title` both have [boolean defaults](https://github.com/codevise/pageflow/blob/9872a63d8e7b527b4cfaaa6eaeb8019fc090b3d3/packages/pageflow/src/editor/models/Configuration.js#L22). The safer thing would have been to pass `!!configuration.get('hide_title')`. The double bang turns `undefined` into `false`. This is done is most other places using `toggleClass` in the code base.

PR #1361, opened 2020-03-20
Create scrolled entries

aviav:
REDMINE-17508

### DIFF
--- a/admins/pageflow/entry.rb
+++ b/admins/pageflow/entry.rb
@@ -130,6 +130,14 @@ module Pageflow
                   input_html: {style: 'width: 200px'})
         end
 
+        if f.object.new_record?
+          f.input :type_name,
+                  as: :select,
+                  include_blank: false,
+                  collection: entry_type_collection,
+                  wrapper_html: {style: 'display: none'}
+        end
+
         if authorized?(:configure_folder_for, resource)
           folder_collection = collection_for_folders(resource.account, resource.folder)
           f.input(:folder,
@@ -165,6 +173,18 @@ module Pageflow
                     rel: 'duplicate',
                     confirm: I18n.t('pageflow.admin.entries.confirm_duplicate')
                   })
+      end
+    end
+
+    collection_action :entry_types do
+      account = Pageflow::Account.find(params[:account_id])
+
+      if authorized?(:see_entry_types, account)
+        @entry_types = helpers.entry_type_collection_for_account(account)
+
+        render(layout: false)
+      else
+        render(partial: 'not_allowed_to_see_entry_types', status: 403)
       end
     end
 
@@ -251,7 +271,7 @@ module Pageflow
       end
 
       def permitted_attributes
-        result = [:title]
+        result = [:title, :type_name]
         target = if !params[:id] && current_user.admin?
                    Account.first
                  elsif params[:id]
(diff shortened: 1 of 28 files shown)


admins/pageflow/entry.rb
```diff
@@ -168,6 +180,14 @@ module Pageflow
...
     end
 
+    collection_action :entry_types do
+      account = Pageflow::Account.find(params[:account_id])
+
+      @entry_types = helpers.entry_type_collection_for_account(account)
+
+      render(layout: false)
+    end
+
```
tf, line 190: This action could use a test. Do we want to prevent users from querying available entry types of accounts they are not authorized for otherwise?

admins/pageflow/entry.rb
```diff
@@ -130,6 +130,18 @@ module Pageflow
...
+          if col.length > 1
+            f.input :type_name,
+                    as: :select,
+                    include_blank: false,
+                    collection: col
+          else
+            f.hidden_field :type_name, value: col.first.first
+          end
+        end
+
```
tf, line 144: This causes the select to be shown soon as a second entry type is registered in the system. We could consider, hiding the select via JS when it only contains a single item. This would let us install the scrolled entry type behind a disabled feature flag without letting our customers know that entry types are about to be a thing. Also we could simplify the branching logic here. Toggling between accounts with different available entry types would then make the select show up and disappear as well, which might be a bit more confusing than showing a disabled select.

config/initializers/entry_types.rb
```diff
@@ -1,4 +0,0 @@
-Pageflow.configure do |config|
-  config.plugin(PageflowPaged.plugin)
-  config.plugin(PageflowScrolled.plugin)
-end
```
tf, line 4: Requiring the host app to register the entry type plugins is sort of a breaking change. Let's postpone that to a point where we continue decoupling the entry type engines from the main gem. For now, we could still let the entry type plugins define the feature flags and make `PageflowPaged` enable its feature flag by default. That way newly generated host apps would still only provide the paged type by default. The changes you did in Dedicated will make the transition smooth for our developers.

admins/pageflow/entry.rb
aviav: The 'nicest' solution that I'm currently thinking of from a UX perspective would be to only hide the select for users who never have more than one option for the entry type at a time. However, I'm assuming that in terms of complexity that would be excessive. In case you'd think it's okay, I'd go with that, though.

Without further input, I'll just go with hide instead of disable

tf: Agreed. The simple solution of hide instead disable sounds fine for now.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -150,15 +150,6 @@ def self.name
...
       end
-
-      it 'is forbidden for account publisher' do
-        account = create(:account)
-
-        sign_in(create(:user, :publisher, on: account))
-        get(:quota_state, params: {account_id: account})
-
-        expect(response.body).to have_selector('.user_quota_not_allowed')
-      end
```
tf, line 161: There _is_ a policy spec for this, but most of Pageflow's controller specs still test the border between allowed and forbidden stuff (e.g. to make sure the `authorize` call that invokes the tested policy is in there at all). Also this tests a different code path of the controller action than the other two.

PR #1377, opened 2020-04-06
Seed data integration for audio files

DanielKuhn:
REDMINE-17547

### DIFF
--- a/entry_types/scrolled/config/locales/new/de.yml
+++ b/entry_types/scrolled/config/locales/new/de.yml
@@ -89,6 +89,13 @@ de:
           attributes:
             id:
               label: "Video"
+        inlineAudio:
+          name: Inline-Audio
+          tabs:
+            general: Inline-Audio
+          attributes:
+            id:
+              label: "Audio"
         inlineBeforeAfter:
           name: Vorher/Nachher
           tabs:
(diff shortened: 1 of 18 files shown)


entry_types/scrolled/lib/tasks/pageflow_scrolled_tasks.rake
```diff
@@ -72,6 +72,14 @@ namespace :pageflow_scrolled do
...
                 }
               }.stringify_keys
+            },
+            audio_files: {
+              rocky_theme: {
+                url: 'https://s3-eu-west-1.amazonaws.com/de.codevise.pageflow.development/pageflow-next/seed-assets/audios/Rocky_Theme.m4a',
+                configuration: {
+                  testReferenceName: 'rocky_theme'
+                }
+              }.stringify_keys
```
tf, line 82: Do we need to think about the license of this file?

tf: Not yet documented here: https://github.com/codevise/pageflow/blob/0c84a00949f2c34934ebd71da5306dcc5e0beb65/entry_types/scrolled/package/spec/support/stories.js#L72-L78

tf:
Percy failure seems to be related to inline audio, though. Can you investigate this?

DanielKuhn:
I added #1382 as a prerequisite for this.

PR #1378, opened 2020-04-06
Loading spinner image fov fix

awaismbajwa:
REDMINE-17549

### DIFF
--- a/packages/pageflow-react/src/loadingSpinner/components/MediaLoadingSpinner.jsx
+++ b/packages/pageflow-react/src/loadingSpinner/components/MediaLoadingSpinner.jsx
@@ -82,13 +82,22 @@ function preventScrollBouncing(e) {
   e.preventDefault();
 }
 
-function backgroundImageInlineStyles({firstPageBackgroundImageUrl, backgroundImage, blurStrength}) {
+function backgroundImageInlineStyles({firstPageBackgroundImageUrl, backgroundImage, blurStrength, backgroundImageX, backgroundImageY}) {
+  
+  var backgroundPosition = {
+    x: backgroundImageX != undefined ? backgroundImageX : 50,
+    y: backgroundImageY != undefined ? backgroundImageY : 50
+  } 
   const url = backgroundImage ? backgroundImage.urls.medium : firstPageBackgroundImageUrl;
   if (url) {
-    return {
+    var style = {
       backgroundImage: `url("${url}")`,
-      filter: 'blur('+blurStrength+'px)'
+      filter: 'blur('+blurStrength+'px)',
     };
+    if (backgroundImage) {
+      style.backgroundPosition = `${backgroundPosition.x}% ${backgroundPosition.y}%`
+    }
+    return style;
   }
 }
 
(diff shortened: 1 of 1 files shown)


tf:
Branch needs to be based on `master`.

packages/pageflow-react/src/loadingSpinner/components/MediaLoadingSpinner.jsx
```diff
@@ -82,12 +86,13 @@ function preventScrollBouncing(e) {
...
 
-function backgroundImageInlineStyles({firstPageBackgroundImageUrl, backgroundImage, blurStrength}) {
+function backgroundImageInlineStyles({firstPageBackgroundImageUrl, backgroundImage, blurStrength, backgroundImageX, backgroundImageY}) {
   const url = backgroundImage ? backgroundImage.urls.medium : firstPageBackgroundImageUrl;
   if (url) {
     return {
       backgroundImage: `url("${url}")`,
-      filter: 'blur('+blurStrength+'px)'
+      filter: 'blur('+blurStrength+'px)',
+      backgroundPosition: backgroundImage ? `${backgroundImageX}% ${backgroundImageY}%`: ''
```
tf, line 95: What happens when background position has not been set?

awaismbajwa: actually I had some unexpected behaviour yesterday with pageflow-react js 
while working with local and remote copy and that made me scratch my head really hard because
of unexpected behaviour and some how these commits also reflect that. I need to fix this MR

packages/pageflow-react/src/loadingSpinner/components/MediaLoadingSpinner.jsx
```diff
@@ -82,13 +82,17 @@ function preventScrollBouncing(e) {
...
   const url = backgroundImage ? backgroundImage.urls.medium : firstPageBackgroundImageUrl;
   if (url) {
-    return {
+    var style = {
       backgroundImage: `url("${url}")`,
-      filter: 'blur('+blurStrength+'px)'
+      filter: 'blur('+blurStrength+'px)',
     };
+    if (backgroundImage) {
+      style.backgroundPosition = `${backgroundImageX}% ${backgroundImageY}%`
```
tf, line 93: This results in `undefined% undefined%` if no position has been set yet. For page backgrounds we [default to a centered position](https://github.com/codevise/pageflow/blob/943210b603ae94c83b110009ed63adbd7cc51d4a/packages/pageflow-react/src/components/BackgroundImage.jsx#L63).

awaismbajwa: I tested this by using the image without setting fov values and then I didnot get undefined. can you please mention how can i reproduce undefined? Or I will try again

tf: This happens for every entry where the position of the custom loading spinner image has not been set yet. Example: https://staging-hosted-pageflow.codevise.de/admin/entries/132

packages/pageflow-react/src/loadingSpinner/components/MediaLoadingSpinner.jsx
```diff
@@ -82,13 +82,22 @@ function preventScrollBouncing(e) {
   e.preventDefault();
 }
 
-function backgroundImageInlineStyles({firstPageBackgroundImageUrl, backgroundImage, blurStrength}) {
+function backgroundImageInlineStyles({firstPageBackgroundImageUrl, backgroundImage, blurStrength, backgroundImageX, backgroundImageY}) {
+  
+  var backgroundPosition = {
+    x: backgroundImageX ? backgroundImageX : 50,
+    y: backgroundImageY ? backgroundImageY : 50
```
tf, line 89: Since 0 is falsy, this makes it fall back to 50 when 0 is selected as position.

awaismbajwa: oops

PR #1386, opened 2020-04-17
[WIP] Remove all occurrences of jQuery from pageflow frontend

DanielKuhn:
Replace jQuery.Deferred with native Promises.
Adjust specs.

REDMINE-17583

### DIFF
--- a/entry_types/paged/packages/pageflow-paged/src/frontend/VideoPlayer/Lazy.js
+++ b/entry_types/paged/packages/pageflow-paged/src/frontend/VideoPlayer/Lazy.js
@@ -1,17 +1,13 @@
+import $ from 'jquery';
 import _ from 'underscore';
-import {VideoPlayer} from './VideoPlayer';
-import {browser} from '../browser';
-import {log} from '../base';
-import {assetUrls} from '../assetUrls';
+import {VideoPlayer, browser, log, assetUrls} from 'pageflow/frontend';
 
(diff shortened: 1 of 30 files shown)


packages/pageflow/src/frontend/VideoPlayer/Lazy.js
```diff
@@ -8,9 +7,11 @@ import {assetUrls} from '../assetUrls';
...
 export const Lazy = function(template, options) {
   var placeholder = $('<span class="video_placeholder" />'),
-      that = this,
-      readyCallbacks = new $.Callbacks(),
-      disposeTimeout, videoTag, videoPlayer, html;
+    that = this,
+
+    // TODO: Replace $.Callbacks
+    readyCallbacks = new $.Callbacks(),
+    disposeTimeout, videoTag, videoPlayer, html;
```
tf, line 14: `VideoPlayer.Lazy` is [only used by linkmap page](https://github.com/codevise/pageflow-linkmap-page/blob/c598a8b3775aab41b802f3f976ce1e48f63fe2a0/app/assets/javascript/pageflow/linkmap_page/page_type.js#L203). For video page types it was replaced by a React implementation.

I therefore suggest moving it to `pageflow-paged`. Then it can keep it's jQuery dependency and we do not have to think about `$.Callbacks`.

DanielKuhn:
closed in favor of #1396

PR #1394, opened 2020-04-20
Media-API

awaismbajwa:
REDMINE-17584

### DIFF
--- a/package/src/frontend/media/createMediaPlayer.js
+++ b/package/src/frontend/media/createMediaPlayer.js
@@ -1,0 +1,50 @@
+import {browser} from '../browser';
+import {VideoPlayer} from '../VideoPlayer';
+
+export const createMediaPlayer = function (options) {
+  options.tagName = options.tagName || 'video';
+  let isAudio = options.tagName == 'audio';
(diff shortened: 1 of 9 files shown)


package/src/frontend/media/createMediaPlayer.js
```diff
@@ -0,0 +1,49 @@
+import {browser} from '../browser';
+import {VideoPlayer} from '../VideoPlayer';
+
+export const createMediaPlayer = function (options) {
+  let isAudio = options.isAudio;
+  let playsInline = options.playsInline;
+
+  let mediaElementTemplate = document.createElement(options.tagName);
```
tf, line 8: tagName is redundant since there is already isAudio.

tf: A lot of the tests also do not pass tagName. What gets created then? It might be worth covering `isAudio` with a spec.

awaismbajwa: this problem is mainly occurred because of me having to work on two different branches going back and forth to make changes and then rebasing. let me look into it closely

PR #1396, opened 2020-04-21
Decouple frontend from jquery and backbone

DanielKuhn:
REDMINE-17583


package/spec/frontend/media_player/hooks_spec.js
```diff
@@ -64,27 +62,31 @@ describe('hooks', function() {
...
         hooks(player, {
           before: function() {
-            return deferred.promise();
+            return promise;
           }
         });
 
         player.play();
-        deferred.resolve();
+        await promiseResolve();
```
tf, line 80: Awaiting resolve does nothing since it does not return a promise. We instead have to await the play. We cannot await `play` directly, since it stays pending until the `before` promise is resolved. We could do

```
const playPromise = player.play();
promiseResolve()
await playPromise
```

package/spec/frontend/media_player/hooks_spec.js
```diff
@@ -64,27 +62,31 @@ describe('hooks', function() {
...
 
       it('does not call original play method until promise returned by before is resolved', function() {
         var player = fakePlayer();
-        var deferred = new jQuery.Deferred();
         hooks(player, {
           before: function() {
-            return deferred.promise();
+            return Promise.resolve();
           }
         });
```
tf, line 92: This test now looks a bit misleading. Before the `before` promise stayed pending. Now `originalPlay` _would_ be called, it is just not called synchronously, which makes the test pass,

We could consider rejecting the promise instead and expecting that play also rejects and orginal play is not called.

package/spec/frontend/media_player/hooks_spec.js
```diff
@@ -104,18 +106,23 @@ describe('hooks', function() {
...
+            return promise;
           }
         });
 
         player.play();
         player.pause();
-        deferred.resolve();
+        promiseResolve();
 
         expect(player.originalPlay).not.toHaveBeenCalled();
```
tf, line 127: I think this test also just passes since there is no await. Maybe you did this and I'm missing something, but generally it's often a good idea to try to change some code back to see a changed test fail and make sure it fails for the right reason.

package/spec/frontend/media_player/hooks_spec.js
```diff
@@ -104,18 +106,23 @@ describe('hooks', function() {
         expect(player.originalPlay).toHaveBeenCalled();
```
tf, line 106: This test also need to await play.

package/spec/frontend/media_player/volume_binding_spec.js
```diff
@@ -91,8 +90,7 @@ describe('volumeBinding', function() {
     it('does not fade in until promise returned by play is resolved', function() {
       var player = fakePlayer();
       var settings = new Backbone.Model({volume: 98});
-      var deferred = new jQuery.Deferred();
-      player.play = function() { return deferred.promise(); };
+      player.play = function() { return Promise.resolve() };
```
tf, line 93: Without looking at any details here, I just want to point out that this changes the test in an important war: Before the promise stayed pending. We should again be sure that this fails for the right reason.

package/spec/frontend/media_player/volume_binding_spec.js
```diff
@@ -103,12 +101,10 @@ describe('volumeBinding', function() {
...
       volumeBinding(player, settings);
 
       player.playAndFadeIn(500);
-      deferred.resolve();
 
       expect(player.fadingVolume).toBe(98);
```
tf, line 109: For all the tests that use things like `fadingVolume`/`currentVolume` we need a new strategy: Since `playAndFadeIn` is async, we can no longer observe effects directly after invocation. We also do not want to rewrite all tests to wait until the whole process is done, like we did in most of the other cases.

What we want to test is that, when we call methods like `playAndFadeIn`, `fadeVolume` is called eventually but before `playAndFadeIn` resolves. We can do this by rewriting properties like `fadingVolume` to promises as done in [this example commit](https://github.com/tf/pageflow/commit/6b7f6e62496d51fe5e60df2bc36efe1af5b5ff6c).

package/src/frontend/VideoPlayer/prebuffering.js
```diff
@@ -26,30 +25,30 @@ export const prebuffering = function(player) {
...
               return;
             }
 
             count++;
 
             if (player.isBufferedAhead(delta) || count > maxCount) {
               log('finished prebuffering video ' + player.src());
-              deferred.resolve();
-              player.prebufferDeferred = null;
+              promise.resolve();
```
tf, line 50: We should make sure that everything we rewrite at least has a very basic test. This for example does not exist.

entry_types/paged/packages/pageflow-paged/src/frontend/ChapterFilter.js
```diff
@@ -1,5 +1,5 @@
 import _ from 'underscore';
-import {Object} from 'pageflow/frontend'
+import {default as Object} from '../Object'
```
tf, line 2: This is the same as `import Object from`. The `as Object` part is only needed when re-exporting (see index.js) to give the export a name. I'd also place the file in the `frontend` directory since it is not shared with `editor`.

package/src/frontend/browser/browser.js
```diff
@@ -100,32 +104,35 @@ export const browser = (function(){
...
 
-      $.when.apply(null, _.map(_.keys(tests), function(name) {
+      Promise.all(_.map(_.keys(tests), function(name) {
         return asyncHas(name).then(function(result) {
           var cssClassName = name.replace(/ /g, '_');
 
-          $('body').toggleClass('has_' + cssClassName, !!result);
-          $('body').toggleClass('has_no_' + cssClassName, !result);
+          document.body.classList.toggle('has_' + cssClassName, !!result)
+          document.body.classList.toggle('has_no_' + cssClassName, !result)
```
tf, line 128: So far `pageflow-paged/frontend` did not require polyfills. Now we need at least one for `Promise` and `classList` for IE11 (https://caniuse.com/#search=classlist). For `Promise` we can use the `core-js` polyfill that we already use in scrolled.

entry_types/paged/packages/pageflow-paged/src/frontend/ChapterFilter.js
DanielKuhn: But `Object.js` does not export `Object`, only `BaseObject`. Therefore it's exported `{default as Object}` in `pagflow/ui` and `pageflow-paged` and imported `{default as Object}` here.

DanielKuhn: I'll move it to frontend, though

package/src/frontend/browser/browser.js
DanielKuhn: added `core-js` and `classlist-polyfill` to `pageflow-paged` package

entry_types/paged/packages/pageflow-paged/src/frontend/ChapterFilter.js
tf: Since it's a default export the name doesn't matter. The import alone determines the name.

DanielKuhn: I see, omitting the curly braces makes all the difference

package/src/frontend/VideoPlayer/prebuffering.js
```diff
@@ -58,19 +64,19 @@ export const prebuffering = function(player) {
         };
 
         timeout();
-        player.prebufferDeferred = deferred;
+        player.prebufferPromise = promise;
       }
     }
 
-    return player.prebufferDeferred ? player.prebufferDeferred.promise() : deferred.resolve().promise();
+    return player.prebufferPromise ? player.prebufferPromise : promise;
```
tf, line 71: Ok, so the spec for the "simple" success case would also have been worth it after all: In manual testing I noticed that autoplay videos sometimes do not start. This is caused by this line: If the video has already buffered enough right from the start, a pending promise instead of a resolved one is returned now.

package/spec/frontend/video/prebuffering_spec.js
```diff
@@ -0,0 +1,65 @@
...
+    it('aborts prebuffering on pause', function() {
+      const player = fakePlayer();
+      const callback = sinon.spy();
+
+      prebuffering(player);
+
+      player.prebuffer().then(callback);
+      player.pause();
+
+      expect(callback).not.toHaveBeenCalled();
```
tf, line 34: Have you seen this spec fail? I think it passes for the wrong reason. Since `callback` is invoked asynchronously, it will never have been called right after the `prebuffer` call.

package/src/frontend/VideoPlayer/filterSources.js
```diff
@@ -1,31 +1,35 @@
...
+  if (playerElement.tagName !== 'Video') {
     return playerElement;
   }
 
   var changed = false;
 
   if (browser.has('mp4 support only')) {
     // keep only mp4 source
-    $(playerElement).find('source').not('source[type="video/mp4"]').remove();
+    playerElement.querySelectorAll('source').forEach(source => {
```
tf, line 12: Just something to be aware of: `querySelectorAll` returns a `NodeList` - not an `Array` - which does not support `forEach` in older browsers. Apparently we already have a polyfill for this, through one of our dependencies.

package/src/frontend/VideoPlayer/filterSources.js
```diff
@@ -1,31 +1,35 @@
-import $ from 'jquery';
 import {browser} from '../browser';
 
 export const filterSources = function(playerElement) {
-  if (!$(playerElement).is('video')) {
+  if (playerElement.tagName !== 'Video') {
```
tf, line 4: Apparently always uppercase. https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName

PR #1398, opened 2020-04-22
React media api

awaismbajwa:
REDMINE-17585


entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
```diff
@@ -10,6 +10,8 @@
     />
     <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700&display=swap" rel="stylesheet">
     <title>Pageflow Next Showcase</title>
+    
+    <%= javascript_include_tag 'pageflow/vendor' %>
```
awaismbajwa, line 14: it is temporary. to stop backbone unavailable error

entry_types/scrolled/package/src/contentElements/inlineAudio/InlineAudio.js
```diff
@@ -1,17 +1,25 @@
...
+  if (configuration.background_type=='background_image') {
+    poster = useFile({collectionName: 'imageFiles', permaId: this.props.background_image_id});
+  }
+  else if (configuration.background_type=='background_video'){
+    poster = useFile({collectionName: 'videoFiles', permaId: this.props.background_video_id});
+  }
   return (
     <div ref={ref} className={classNames(styles.root)}>
       <div>
+        <MediaPlayer type={'audio'} poster={poster} sources={file} />
```
awaismbajwa, line 22: this whole file  changes are just for testing

package/spec/frontend/media/media_spec.js
```diff
@@ -0,0 +1,75 @@
...
+      m4a: 'http://example.com/example.m4a',
+      mp3: 'http://example.com/example.ogg'
+    },
+  }
+  describe('#getPlayer', function() {
+    let player = media.getPlayer(fileSource, {
+      type: 'audio',
+      playsInline: true,
+      loop: true
+    });
```
tf, line 21: Never put test code directly into `describe` blocks. If `getPlayer` causes an exception, the whole test suite will crash. If you want to share test code between tests, you can use a `beforeEach` block. As a default, I'd rather duplicate code like this in all test cases to make them more self contained. 

Tests in the Pageflow repository normally follow the so called triple A structure: Arrange, Act, Assert.

These three blocks should ideally be easy to spot when reading a text, for example by separating them with empty lines.

The Arrange block, sets everything up for the test. It creates dependencies and fixture data that are required to run the code we actually want to test. If there is repetitive setup code, we can consider extracting it to a `beforeEach` block, but should - if possible - create a test helper instead to prevent introducing [mystery guests](https://thoughtbot.com/blog/mystery-guest).

The act block runs the code we want to test. It should never be extracted to a `beforeEach` block since it's the core of the test.  Also extracting the act part often leads to mixing up test cases. Even in this simple example, the `media.getPlayer` call (which is the act part of these tests) contains details related to two different tests: the `loop` option and the `playsinline` option. Also extracting the act block mostly does not scale. What if want to add a test case for when the `loop` option is missing? We will no longer be able to use a common act block.

The assert block, contains the expectations about what the act block did.

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -0,0 +1,26 @@
...
+  render(){
+    let playerHTML = '';
+    if (this.state.player) {
+      playerHTML = this.state.player.el().innerHTML;
+    }
+    return (
+      <div dangerouslySetInnerHTML={{__html: playerHTML}}></div>
+    );
+  }
+}
```
tf, line 26: Let's use hooks for all new components.

Using `innerHTML` basically recreates the DOM elements. This will neither work with blessed elements. Also it causes the inserted media elements to no longer be controlled by the player instance created above. We can use a ref and `appendChild` instead.

package/src/frontend/media/media.js
```diff
@@ -0,0 +1,29 @@
...
+      player.muted(value);
+    });
+  },
+  processSource: function (fileSource, type) {
+    return [
+      {type: type+'/ogg', src: `${fileSource.urls.ogg}?u=1`},
+      {type: type+'/mp4', src: `${fileSource.urls.m4a}?u=1`},
+      {type: type+'/mp3', src: `${fileSource.urls.mp3}?u=1`}
+    ];
+  },
```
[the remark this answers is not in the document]
awaismbajwa: Basically this was done in my old media ticket that sources were prepared in react component and then passed to MediaTag. Same is done in existing pageflow-react setup.
Since we don't have MediaTag now so we can simply pass the sources to player.src(sources) method.
after this processSources method is not required anymore

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -0,0 +1,28 @@
...
+  const playerWrapper = useRef(null);
+  const [player, setPlayer] = useState(undefined);
+  
+  useEffect( () => {    
+    if (props.sources && player == undefined) {
+      setPlayer(media.getPlayer(props.sources, {
+        poster: props.poster
+      }));
+    }
+  } );
```
tf, line 14: Please install a linter plugin for your editor, so you see React hooks related errors.

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -0,0 +1,28 @@
...
+    </div>
+  );
+
+}
+
+function renderPlayer(props, player, parentElement) {
+  if (player && parentElement.current.childElementCount == 0) {
+    parentElement.current.appendChild(player.el());
+  }
+}
```
tf, line 28: Render functions should be side effect free. You can use a `useEffect` hook to obtain and append the player. Then you also do not need the state. Since this component manipulates the DOM, we should make sure that React  does not update the DOM which might erase our changes. You can take a look at the way `MediaTag` from `pageflow-paged-react` uses `shouldComponentUpdate` to this. For functional components, we can use `React.memo` to achieve a similar thing.

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -0,0 +1,28 @@
+import {useEffect, useState, useRef} from 'react';
+import {media} from 'pageflow/frontend';
+
+export function MediaPlayer(props){
```
tf, line 4: As an important shared component, this should have a test.

entry_types/scrolled/package/src/frontend/index.js
```diff
@@ -21,6 +21,8 @@ export * from './InlineCaption';
 export * from './Text';
 export * from './Video';
 
+export * from './MediaPlayer';
```
tf, line 24: Now that `MediaPlayer` only takes sources, I think it would make sense to introduce a `VideoPlayer` component (or rewrite the existing `Video` component) so that we still have a component that takes a video file as returned by `useFile`, comes up with the correct sources and renders a `MediaPlayer`.

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
```diff
@@ -0,0 +1,34 @@
...
+    expect(queryByRole('audio')).toBeNull();
+  });
+
+  it('renders audio tag for audio sources', () => {
+    const {queryByRole} =
+      renderInEntry(<MediaPlayer type={'audio'} sources={audioSources} />, {
+        seed: {}
+      });
+
+    expect(queryByRole('audio')).toBeDefined();
```
awaismbajwa, line 31: queryByRole('..') is somehow always returning null. What could be the reason of it?

entry_types/scrolled/package/src/frontend/index.js
awaismbajwa: it can be handled by passing type as a prop

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
tf: I don't know. Maybe you can inspect the rendered result by looking at `container.innerHTML` (https://testing-library.com/docs/react-testing-library/api#container)?

awaismbajwa: If I dont use renderInEntry then i do get something like this

`<body class="has_autoplay_support has_css_animations has_no_facebook_toolbar has_no_ie has_no_ios_platform has_no_iphone_platform has_no_mobile_platform has_no_phone_platform has_pushstate_support has_request_animation_frame_support has_no_touch_support has_rewrite_video_sources_support has_stop_buffering_support has_buffer_underrun_waiting_support has_prebuffering_support has_no_mp4_support_only has_no_mse_and_native_hls_support has_no_native_video_player has_volume_control_support has_audio_context_volume_fading_support"><div><div aria-label="Audio Player" class="vjs-paused vjs-controls-disabled vjs-audio vjs-workinghover vjs-v6 vjs-user-active video-js player" id="pageflow_media_element_0" lang="en-us" role="region" tabindex="-1"><audio class="vjs-tech" id="pageflow_media_element_0_html5_api" tabindex="-1" /><div aria-disabled="false" class="vjs-poster vjs-hidden" tabindex="-1" /><div aria-atomic="true" aria-live="off" class="vjs-text-track-display" /></div></div></body>`

tf: > If I dont use renderInEntry 

What does this mean? `container` should also be usable in combination with `renderInEntry`.

awaismbajwa: I dont know about this API, I can try

awaismbajwa: it is working now

entry_types/scrolled/package/src/frontend/index.js
tf: As [mentioned earlier](https://github.com/codevise/pageflow/pull/1398#discussion_r413558763) there will be more logic involved for assembling the list of sources. This logic needs to live somewhere and it should not be in the individual clients using the `MediaPlayer`. I therefore still think a video specific component would make sense to already add now. It will also make adding stories easier, since there we also have to deal with `VideoFile` objects from seed date instead of just source arrays.

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  it('renders audio tag for audio sources', () => {
+    const {container} = renderInEntry(<MediaPlayer type={'audio'} sources={audioSources} />, {
+      container: document.createElement('div')
+    });
+    
+    expect(container.querySelector('audio')).toBeDefined();
+  });
+
```
tf, line 30: The component has more logic then what is tested here. Especially the releasePlayer part seems interesting. I'd suggest using [react-testing-library's unmount helper](https://testing-library.com/docs/react-testing-library/api#unmount) and [spy on](https://jestjs.io/docs/en/jest-object#jestspyonobject-methodname) the releasePlayer method.

awaismbajwa: added one more spec 'releases media player on component unmount'

entry_types/scrolled/package/src/frontend/index.js
awaismbajwa: added VideoPlayer and AudioPlayer in pageflow-scrolled/frontend

entry_types/scrolled/package/src/frontend/VideoPlayer.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {Object} props
+ * @param {number} props.id - Perma id of the video file.
+ */
+export function VideoPlayer(props) {
+  const videoFile = useFile({collectionName: 'videoFiles', permaId: props.id});
+  const processSources = file => [
+    {type: 'video/ogg', src: `${file.urls.ogg}?u=1`},
+    {type: 'video/mp4', src: `${file.urls.m4a}?u=1`},
+    {type: 'video/mp3', src: `${file.urls.mp3}?u=1`}
+  ];
```
tf, line 18: These sources do not make sense. You can look at the old implementation as a reference ([1], [2]). Since we won't be doing bandwidth detection, you can ignore everything regarding that and always use high quality.

[1] https://github.com/codevise/pageflow/blob/master/entry_types/paged/packages/pageflow-paged-react/src/media/components/VideoFilePlayer/sources.js
[2] https://github.com/codevise/pageflow/blob/master/entry_types/paged/packages/pageflow-paged-react/src/media/components/VideoFilePlayer/__spec__/sources-spec.js

awaismbajwa: could you please write how sources should look like? 
it is difficult to decode from the mentioned references

tf: I'd say it's basically in the test:

* All files have mp4 and hls urls
* Some files have Dash urls (Dash was only added in a later version, so files encoded before do not have these outputs
* If the user requested a specific quality (medium, fullhd) include only that url in the sources
* Fullhd or 4K are only available of the originally uploaded video has at least that quality. So even if the user requested 4K, we have to fallback to a lower quality if the requests quality is not present.

awaismbajwa: If I understand correctly in current state it should just use high quality url file.urls['high'] ? 
or should I implement that sources function separately ?

tf: We need the sources logic eventually. We could call this PR done without it, but since it is blocked by the jQuery stuff anyway, we could also continue here.

entry_types/scrolled/package/src/frontend/AudioPlayer.js
```diff
@@ -0,0 +1,24 @@
...
+import {useFile} from '../entryState';
+import {MediaPlayer} from './MediaPlayer';
+
+
+/**
+ * Render audio file in MediaPlayer.
+ *
+ * @param {Object} props
+ * @param {number} props.id - Perma id of the audio file.
+ */
```
tf, line 11: It's a bit cumbersome, but things with inline docs need to be added to the TOC in `package/documentation.json`. Otherwise the show up at the very end of the sidebar in the [generated docs](https://codevise.github.io/pageflow-docs/scrolled/js/master/index.html).

awaismbajwa: is it `package/documentation.yml`?
I added AudioPlayer and VideoPlayer in there

tf: Ah yes, my bad.

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -0,0 +1,42 @@
...
+    if (props.sources) {
+      let player = media.getPlayer(props.sources, {
+        poster: props.poster,
+        playsInline: props.playsInline,
+        loop: props.loop,
+        controls: props.controls,
+        tagName: props.type,
+        ended: function () {
+          props.nextSectionOnEnd && scrollToSection('next');
+        }
```
tf, line 22: If you listen to this event via `player.on`, you can get around the special handling for the `ended` option in `createMediaPlayer`.

package/spec/frontend/media/media_spec.js
```diff
@@ -0,0 +1,96 @@
...
+      let player1 = media.getPlayer(fileSources, {});
+      let player2 = media.getPlayer(fileSources, {
+        playerId: 'test'
+      });
+      media.releasePlayer('test');
+      media.mute(true);
+
+      expect(player1.muted()).toEqual(true);
+      expect(player2.muted()).toEqual(false);
+    });
```
tf, line 93: This spec fails for some reason.

entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
```diff
@@ -10,6 +10,8 @@
...
```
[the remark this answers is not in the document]
awaismbajwa: yes, I have replaced this with videojs and added it to the list of precompile assets

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
awaismbajwa: I did try this initially but due to some reason event was not getting invoked at that time but now I have tried it again and it is working fine.

entry_types/scrolled/package/src/frontend/VideoPlayer.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+VideoPlayer.defaultProps = {
+  interactive: false,
+  controls: true
+};
+
+
+function processSources(file, quality='auto', {hasHighBandwidth} = {}) {
```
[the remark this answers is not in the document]
awaismbajwa: removed this parameter

entry_types/scrolled/package/.storybook/preview-head.html
```diff
@@ -1,4 +1,6 @@
 <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700&display=swap" rel="stylesheet">
+<script src='videojs.js'></script>
+<script src='videojs-dash.js'></script>
```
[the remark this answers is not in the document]
awaismbajwa: this was the temp setup. I have added videojs and videojs-dash in the node_module/@storybook/core/public just to make it work for my testing.

entry_types/scrolled/package/documentation.yml
```diff
@@ -15,10 +15,12 @@ toc:
       React components for building content elements.
     children:
       - Audio
+      - AudioPlayer
       - Image
       - InlineCaption
       - Text
       - Video
+      - VideoPlayer
```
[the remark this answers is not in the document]
awaismbajwa: yes

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+
+      return () => {
+        media.releasePlayer(player);
+        playerWrapper.innerHTML = '';
+        player.dispose();
+      }
+    }
+  }, [scrollToSection, props.autoplay, props.state, props.sources, props.poster, props.type,
+      props.loop, props.controls, props.nextSectionOnEnd, props.playsInline]);
```
tf, line 46: Looking at how complicated it is to construct a memoized array of sources in the two components above, makes me think we should reconsider our approach here. Also while testing, I still saw unneeded recreation of the player when I selected elements in the editor and in some cases on scroll in the storybook.

The solution implemented here is exactly what we discussed, but looking at it now and testing it, I think the main problem is that we depend on clients of `MediaPlayer` component to pass stable props to ensure the player is not recreated. The reason the player is not recreated should be a feature of the `MediaPlayer` itself, though. Especially since `MediaPlayer` receives more props than is actually needed for the player creation (e.g. `nextSectionOnEnd`), this problem will otherwise always come up again whenever we add something to `MediaPlayer` or use it in a new setting.

To achieve this, I think we can use a similar approach to how `MediaTag` works in `pageflow-paged-react`, which uses `shouldComponentUpdate` to re-render the component only if props changed in a substantial way. The counterpart of `shouldComponentUpdate` for function components is [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo). I suggest to extract a component (which could be called `PlayerContainer`) that takes only the props required to call `getPlayer`. It recreates the player in a `useEffect` that runs on every render and then uses `React.memo` with a custom comparison function to only rerender when the props changed. This comparison can work exactly like in `pageflow-paged-react`'s `MediaTag` with a deep comparison of sources. That way it does not matter to pass the same instance of the `sources` object, but just one with equal urls.

It can handle `onSetup`/`onDispose` props just like `MediaTag` to pass the created `player` object out. The rest of the player state/autoplay logic can then continue to live in `MediaPlayer`. That way autoplay code will also live in a different effect than player creation. The above code recreated the player each time `props.state` changes.

package/src/frontend/media/media.js
```diff
@@ -2,14 +2,18 @@ import {createMediaPlayer} from './createMediaPlayer';
 
 export const media = {
   players: {},
+  muteState: true,
+  mediaOff: true,
```
tf, line 6: I think, `mediaOff` should continue to live in `pageflow-scrolled` since it is used specifically to make scrolled section thumbnails work. Also React context is a good fit IMO since whether media should be off is really all about where a component is rendered (i.e. in the preview or in a thumbnail). It's not global in the way that `muteState` is. In the current state, `mediaOff` is never set anywhere causing autoplay to break in published entries.

rollup.config.js
```diff
@@ -120,6 +121,7 @@ const pageflow = [
     output: {
       file: pageflowPackageRoot + '/frontend.js',
       format: 'esm',
+      globals: {'videojs': 'videojs'}
```
tf, line 124: `globals` is not used for `esm` builds as far as I'm aware.

package/src/frontend/media/media.js
awaismbajwa: ok, I reverted back to using useMediaSettings hook for this. But still i did not find mediaOff being set anywhere?

tf: When a component that is rendered in a [`SectionThumbnail`](https://github.com/codevise/pageflow/blob/023fdb7c5b3b917bb94f4adafa18a31f8bd9c753/entry_types/scrolled/package/src/frontend/SectionThumbnail.js#L9), there is no provider of `MutedContext` above it. So `useMediaSettings` will fall back to the [default value of the `MutedContext` context](https://github.com/codevise/pageflow/blob/023fdb7c5b3b917bb94f4adafa18a31f8bd9c753/entry_types/scrolled/package/src/frontend/MutedContext.js#L3), which sets `mediaOff` to `true`. `Entry` [renders a provider for `MutedContext`](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/src/frontend/Entry.js#L63) without setting `mediaOff`, it will thus be `undefined`  for components reading it via `useMediaSettings`. It would have probably been better to explicitly set it to `false` in `Entry` to make both cases grepable.

entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -0,0 +1,44 @@
...
+        if (player.readyState() > 0) {
+          player.play();
+        } else {
+          player.on('loadedmetadata', player.play);
+        }
+      } else {
+        player.pause();
+      }
+    }
+  }, [mediaSettings.mediaOff, props.autoplay, props.state, player]);
```
tf, line 30: Using a local variable for `player` does not work: On first render, the `onSetup` function and the function passed to `useEffect` share the closure of the render function. So when the effect fires it might be able to read the `player` variable set by `onSetup`. On next render, `onSetup` will not be executed again, though. Still the effect function now sees the closure of this new call of the render function. `player` will thus be undefined. To fix this, `player` needs to be a ref. As a rule of thumb, whenever you would have [assigned something to `this`](https://github.com/codevise/pageflow/blob/023fdb7c5b3b917bb94f4adafa18a31f8bd9c753/entry_types/paged/packages/pageflow-paged-react/src/media/components/createFilePlayer/index.jsx#L46) on a class component, you will probably need a ref in a function component.

Also `mediaOff` will only ever be `true` or `undefined`. The strict equality check will therefore always fail. Checking for falsiness (i.e. `!mediaSettings.mediaOff`) makes autoplay work.

Normally I'd say let's add a test to cover this logic. But I'm hesitant since this functionality will change when we wire up `playerState` in the next step. It might be a good example case, though, since a lot of similar specs will be needed for the `playerState` integration.

PR #1399, opened 2020-04-23
Repair user invite form for admin users

aviav:
REDMINE-17539

### DIFF
--- a/admins/pageflow/user.rb
+++ b/admins/pageflow/user.rb
@@ -122,9 +122,11 @@ module Pageflow
 
     collection_action :quota_state do
       @account = Pageflow::Account.find(params[:account_id])
-      authorize!(:add_member_to, @account)
-
-      render(layout: false)
+      if authorized?(:see_user_quota, @account)
+        render(layout: false)
+      else
+        render(partial: 'not_allowed_to_see_user_quota')
+      end
     end
 
     collection_action 'me', title: I18n.t('pageflow.admin.users.account'), method: [:get, :patch] do
--- a/app/policies/pageflow/account_policy.rb
+++ b/app/policies/pageflow/account_policy.rb
@@ -88,6 +88,11 @@ module Pageflow
          query.has_at_least_role?(:manager))
     end
 
+    def see_user_quota?
+      user.admin? ||
+        query.has_at_least_role?(:manager)
+    end
+
     def edit_role_on?
       user.admin? || query.has_at_least_role?(:manager)
     end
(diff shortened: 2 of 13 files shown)


app/policies/pageflow/account_policy.rb
```diff
@@ -88,6 +88,11 @@ def add_member_to?
          query.has_at_least_role?(:manager))
     end
 
+    def see_user_quota?
+      user.admin? ||
+        query.has_at_least_role?(:manager)
+    end
+
```
tf, line 95: Since all the other methods of this class are covered by specs in quite some detail, I'd lean towards also adding specs for this new one.

admins/pageflow/user.rb
```diff
@@ -122,9 +122,11 @@ module Pageflow
 
     collection_action :quota_state do
       @account = Pageflow::Account.find(params[:account_id])
-      authorize!(:add_member_to, @account)
-
-      render(layout: false)
+      if can?(:see_user_quota, @account)
```
tf, line 125: ActiveAdmin supports multiple [authorization adapters](https://activeadmin.info/13-authorization-adapter.html#using-the-cancan-adapter) and provides a more general `authorized?` helper that will work with any of them. Using `can?` here has the same effect but depends on the fact that we use the CanCan adapter.

spec/features/admin/inviting_a_user_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+feature 'as system admin, inviting a user' do
+  scenario 'switching between accounts with and without free quota', js: true do
+    pageflow_configure do |config|
+      config.quotas.register(:users, QuotaDouble.account_exhaustible)
+    end
+
+    account_with_quota = create(:account)
+    account_without_quota = create(:account, name: 'Exhausted account')
```
tf, line 10: The above quota double is a bit too magic for my taste. By only looking at the test, I never would have guessed that it's the name of the account that causes the quota to be exhausted for the one account and available for the other.

Maybe `QuotaDouble.account_exhaustible` can receive a mapping from account to desired quota state etc. Alternatively I would at least give the method a really descriptive name like `QuotaDouble.determine_state_by_account_name` to point out what is going on here. Still, specifying a mapping feels simpler and more expressive.

spec/support/helpers/quota_double.rb
```diff
@@ -38,4 +38,22 @@ def state_description
...
+
+      def state
+        self.class.exhausted_accounts.include?(account) ? 'exhausted' : 'available'
+      end
+
+      def state_description
+        state == 'exhausted' ? EXHAUSTED_DESCRIPTION : AVAILABLE_DESCRIPTION
+      end
+    end
+  end
```
tf, line 58: Using `define_method` instead of `def` would have been the alternative since it does not make local variables go out of scope.

PR #1401, opened 2020-04-27
Use files in CI/storybook seed

DanielKuhn:
REDMINE-17585

### DIFF
--- a/entry_types/scrolled/lib/tasks/pageflow_scrolled_tasks.rake
+++ b/entry_types/scrolled/lib/tasks/pageflow_scrolled_tasks.rake
@@ -90,7 +90,7 @@ namespace :pageflow_scrolled do
             }
           },
           options: {
-            skip_encoding: ENV.fetch('PAGEFLOW_SKIP_ENCODING_STORYBOOK_FILES', true)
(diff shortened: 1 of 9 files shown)


entry_types/scrolled/lib/tasks/pageflow_scrolled_tasks.rake
```diff
@@ -90,7 +90,7 @@ namespace :pageflow_scrolled do
             }
           },
           options: {
-            skip_encoding: true
+            skip_encoding: ENV.fetch('PAGEFLOW_SKIP_ENCODING_STORYBOOK_FILES', true)
```
tf, line 93: The string "false" will be truthy.

tf: Maybe false should be the default? As a developer I will probably want a working storybook, right? Then CI can use the env var to override?

DanielKuhn: True, especially since the PR also changes the content elements to use the transcoded files.

PR #1402, opened 2020-04-27
Update edge script

aviav:
Current version throws error, because there is no frontend.js.

REDMINE-17606


tf:
Thanks. Yeah, another one I noticed last week and forgot to commit on master. Sorry you have to discover them one by one now.

PR #1411, opened 2020-05-06
Simplify some request specs

aviav:
Maybe at some intermediate stage of development it was necessary to
specify a type name on entry creation here. By now, we can do without
that, so these specs can be simplified.

REDMINE-17508

### DIFF
--- a/spec/requests/pageflow/entries_show_request_spec.rb
+++ b/spec/requests/pageflow/entries_show_request_spec.rb
@@ -34,7 +34,7 @@ module Pageflow
 
     context 'with other known format' do
       it 'responds with not found' do
-        entry = create(:entry, :published, type_name: 'test')
+        entry = create(:entry, :published)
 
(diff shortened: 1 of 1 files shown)


tf:
I think this change misses the point of the spec. The whole spec is about defining a `frontend_app` for a custom entry type called `test` and then testing how the published entry HTTP endpoint behaves for entries _of that type_. Some requests might not actually make it all the way to the custom app since they are blocked by authorization logic etc. early on, but still each spec should be about interacting with an entry of the defined entry type. Removing the entry type, will still make most of the specs work (as this PR shows), but risks to test something completely different, namely that the frontend app defined by the default paged entry type has the desired behavior. But we want to test that this behavior can be achieved with any custom frontend app.

aviav:
Interesting. I had never tried to understand what that frontend_app parameter is for, but reading the documentation, I now understand what that is. Indeed, this 'simplification' makes things a little too simple

PR #1412, opened 2020-05-06
Media player state hook

awaismbajwa:
REDMINE-17588

### DIFF
--- a/entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
@@ -2,7 +2,9 @@ import React from 'react';
 import '@testing-library/jest-dom/extend-expect'
 import 'support/mediaElementStub';
 import 'support/fakeBrowserFeatures';
+import {getInitialPlayerState, getPlayerActions} from 'support/fakePlayerState';
 
+import MutedContext from 'frontend/MutedContext';
 import {render} from '@testing-library/react'
 import {media} from 'pageflow/frontend';
 import {MediaPlayer} from 'frontend/MediaPlayer';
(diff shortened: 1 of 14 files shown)


entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
```diff
@@ -30,41 +32,103 @@ describe('MediaPlayer', () => {
...
+    let state = getInitialPlayerState();
+    let playerActions = getPlayerActions();
+    const {rerender} = render(<MediaPlayer type={'video'} sources={getVideoSources()} playerState={state} playerActions={playerActions} />);
+    
+    state = getInitialPlayerState();
+    state.shouldPlay = true;
+
+    let spyPlay = jest.spyOn(media.players['0'], 'play');
+    rerender(<MediaPlayer type={'video'} sources={getVideoSources()} playerState={state} playerActions={playerActions} />);
+    expect(spyPlay).toHaveBeenCalledTimes(1);
```
tf, line 80: It might have been simpler to to stub `media.getPlayer` and return a fake player with mock functions for things like `play`. These tests now depend on the fact that `media` stores its players in a `players` property, which is an implementation detail that might change when we implement the blessed pool.

This is not a blocker, though. We can merge like this and change it later.

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
```diff
@@ -30,41 +32,103 @@ describe('MediaPlayer', () => {
...
+    let spyPlay = jest.spyOn(media.players['0'], 'playOrPlayOnLoad');
+        
+    rerender(
+      <MutedContext.Provider value={{muted: true}}>
+        <MediaPlayer type={'video'} sources={getVideoSources()} state={'active'} autoplay={true} playerState={state} playerActions={playerActions} />
+      </MutedContext.Provider>
+    );
+    setTimeout(()=>{
+      expect(spyPlay).toHaveBeenCalled();
+    }, 500);
```
tf, line 131: To use `setTimeout` in a test, the test needs to be async. Like this Jest does not know which test the `expect` belongs to. You can change it to `not.toHaveBeenCalled` and still see the test pass (without an assertion).

The core problem here, though, is the fact that we wrap the `MediaPlayer` in another component during the second render. This causes the `MediaPlayer` from the first render to be unmounted. The new `MediaPlayer` component then requests its own `player`. So `playOrPlayOnLoad` never gets called on the original player we spy on.

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
```diff
@@ -30,41 +32,103 @@ describe('MediaPlayer', () => {
...
+  //   state = getInitialPlayerState();
+  //   state.fadeDuration = 100;
+  //   state.shouldPlay = true;
+  //   console.log(media.players['0'].playAndFadeIn);
+    
+  //   let spyPlayAndFadeIn = jest.spyOn(media.players['0'], 'playAndFadeIn');
+  //   rerender(<MediaPlayer type={'video'} sources={getVideoSources()} playerState={state} playerActions={playerActions} />);
+  //   expect(spyPlayAndFadeIn).toHaveBeenCalledTimes(1);
+  //   expect(spyPlayAndFadeIn).toHaveBeenCalledWith(100);
+  // });
```
tf, line 98: This spec passes if we enable volume fading [here](https://github.com/codevise/pageflow/pull/1412/files#diff-1763f9021578071122c83cf4dbaef00fL30).

awaismbajwa: sorry where? it is navigating to the changes tab I did not get what you mean

tf: Sorry for the broken link. I mean here:

https://github.com/codevise/pageflow/blob/43ca01f51d1dd2ab150f0ec2e692035da81ae7be/package/src/frontend/media/createMediaPlayer.js#L30

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
awaismbajwa: Thanks alot for this

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
awaismbajwa: let's make these changes in blessed player pool task.
So that this gets merged and Daniel can continue.

tf: Agreed. I'll merge soon as CI passes.

PR #1414, opened 2020-05-11
Update edge build script

aviav:
REDMINE-17508


tf:
Ah yes, I forgot to change this in #1404. But the frontend file is not simply gone, it turned into a [output directory](https://github.com/codevise/pageflow/blob/7c8d2268814490a0456dd3ae6da95794b72af3e5/rollup.config.js#L195) for multiple chunks. In my local workspace there are actually a lot of old versions of these chunk files. So we probably need to `git clean -xf` the directory during build. Also we need to update the `gemspec` to include this directory.

tf:
see #1415

PR #1415, opened 2020-05-11
Rewrite build scripts to use gemspec files

tf:
REDMINE-17603

### DIFF
--- a/bin/build-edge
+++ b/bin/build-edge
@@ -1,17 +1,24 @@
 #!/bin/bash
 
+find_ignored_files_from_gemspec() {
+  gemspec_files=$(ruby -e \
+                       'puts Gem::Specification.load("pageflow.gemspec").files.reject { |f| File.directory?(f) }.join(" ")')
+
+  # --no-index includes ignored files even if there already is a build
+  # commit which force added them on this branch.  The `|| true` makes
+  # sure the script does not abort just because there were no matching
+  # files.
+  git check-ignore --no-index $gemspec_files || true
+}
+
 add_and_commit() {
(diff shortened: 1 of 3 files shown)


bin/build-edge
```diff
@@ -1,17 +1,24 @@
 #!/bin/bash
 
+find_ignored_files_from_gemspec() {
+  gemspec_files=$(ruby -e \
+                       'puts Gem::Specification.load("pageflow.gemspec").files.reject { |f| File.directory?(f) }.join(" ")')
```
aviav, line 5: This line looks like it shouldn't pass a linter. Maybe the indentaton is okay, but the line is still very long. Well, this is just a shell script, so I won't insist

tf: You are right. I'm just really scared of whitespace, quoting and line breaks in shell scripts - mainly because I don't understand. Will try to be braver next time.

aviav: Maybe the basics are simple here. I'd say, close the string, escape the line break, reopen the string on the next line. Haven't tested it, though.

See https://unix.stackexchange.com/a/281310

tf: Posts like [this](https://unix.stackexchange.com/questions/272353/how-to-break-a-long-string-into-multiple-lines-in-the-prompt-of-read-p-within-t) make it look like it might be more involved. It's probably not that hard in this case since we do not care about extra white space in the Ruby script caused by indentation etc. But still...

aviav: Interesting pattern, pack a long string into a short variable and repeat until it fits!

PR #1417, opened 2020-05-12
[WIP] Remove dependency between label width and initial slider position.

UsmanMuhammad:
[REDMINE-17372](https://redmine.codevise.de/issues/17372)

- [X]  remove dependency between label width and slider initial position
- [X]  give height to element if there are no images
- [X]  add simple placeholder layout (differently colored rectangles)
- [X ]  delay label fadeout after mousedown/touchstart


tf:
This has been decomposed into multiple smaller PRs.

PR #1418, opened 2020-05-12
[TMP] React media api

aviav:


entry_types/scrolled/lib/tasks/pageflow_scrolled_tasks.rake
```diff
@@ -35,6 +35,8 @@ namespace :pageflow_scrolled do
           Paperclip::Storage::S3.class_eval { def flush_writes; end }
         end
 
+        Pageflow.config.features.enable('scrolled_entry_type')
+
```
tf, line 39: I'd suggest enabling the feature via the entry's `feature_configuration` (similar to [here](https://gitlab.codevise.de/pageflow/dedicated_pageflow/blob/master/db/seeds.rb#L112)). Otherwise the created entry will still be invalid next time we try to save it.

aviav: I just realized this should be the account's feature configuration. Will fix. Maybe then, also Travis'll pick this up again

tf: Great, yes, you are right. Can you do the same change for the seed data of Dedicated? There we also create a "Pageflow Scrolled" account.

aviav: 👍

aviav:
Closing TMP PR after stuff has been merged

PR #1420, opened 2020-05-14
Make entry templates editable

aviav:
REDMINE-17306

### DIFF
--- a/admins/pageflow/entry_templates.rb
+++ b/admins/pageflow/entry_templates.rb
@@ -1,0 +1,127 @@
+module Pageflow
+  ActiveAdmin.register EntryTemplate, as: 'EntryTemplate' do
+    menu false
+    config.batch_actions = false
+    actions :new, :create, :edit, :update
+    form partial: 'form'
+
+    belongs_to :account, parent_class: Pageflow::Account
+
+    breadcrumb do
+      breadcrumb_links.first 3
+    end
+
+    controller do
+      helper Pageflow::Admin::FormHelper
+      helper Pageflow::Admin::WidgetsHelper
+
+      def new
+        @page_title = page_title('new', params[:entry_type])
+        super
+      end
+
+      def create
+        @entry_template = EntryTemplate.new(
+          entry_template_params.merge(account_id: permitted_params[:account_id])
+        )
+        authorize!(:create, @entry_template)
+        create! { redirect_path }
+        update_widgets if @entry_template.errors.empty?
+      end
+
+      def edit
+        @page_title = page_title('edit', resource.entry_type)
+        super
+      end
+
+      def update
+        @entry_template = EntryTemplate.find(params[:id])
+        @entry_template.assign_attributes(entry_template_params)
+        params[:entry_template].delete('share_providers')
+        params[:entry_template].delete('configuration')
+        authorize!(:update, @entry_template)
+        update! { redirect_path }
+        update_widgets if @entry_template.errors.empty?
+      end
+
+      private
+
+      def page_title(rest_verb, entry_type_name)
+        "#{I18n.t('pageflow.admin.entry_templates.page_title.' + rest_verb)} "\
+        "#{I18n.t('activerecord.values.pageflow/entry.type_names.' + entry_type_name)}"
+      end
+
+      def permitted_params
+        params.permit(
+          :account_id,
+          entry_template: permitted_entry_template_attributes
+        )
+      end
+
+      def permitted_entry_template_attributes
+        [
+          :id,
+          :account_id,
+          :entry_type,
+          :theme_name,
+          :default_author,
+          :default_publisher,
+          :default_keywords,
+          :default_locale,
+          share_providers: {},
+          configuration: {}
+        ]
+      end
+
+      def redirect_path
+        admin_account_path(params[:account_id], tab: 'entry_templates')
+      end
+
+      def entry_template_params
+        current_params = permitted_params[:entry_template]&.to_hash
+
+        if current_params
+          config = true_false_strings_to_booleans_or_numbers(
+            current_params['configuration']
+          )
+          share_providers = true_false_strings_to_booleans_or_numbers(
+            current_params['share_providers']
+          )
+          current_params.merge('configuration' => config,
+                               'share_providers' => share_providers)
+        else
+          {}
+        end
+      end
+
+      def true_false_strings_to_booleans_or_numbers(hash)
+        hash&.map { |k, v| [k, to_boolean_or_number(v)] }&.to_h
+      end
+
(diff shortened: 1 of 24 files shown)


admins/pageflow/entry_templates.rb
```diff
@@ -0,0 +1,66 @@
+module Pageflow
+  ActiveAdmin.register EntryTemplate, as: 'EntryTemplate' do
```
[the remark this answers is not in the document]
tf: I think we should silence these for `register` and `controller` in the `admins` dir, just like we do for [`describe` in the `spec` dir](https://github.com/codevise/pageflow/blob/6329cf329b7ea3b32e739ec4c59629dd5ba9e92b/.rubocop.yml#L42)

admins/pageflow/entry_templates.rb
```diff
@@ -0,0 +1,66 @@
...
+    menu false
+    config.batch_actions = false
+    actions :new, :create, :edit, :update
```
tf, line 5: I still think only providing `edit` and overriding `find_resource` (as mention [here](https://redmine.codevise.de/issues/17306#note-56)) would have been a good idea. Is there an issue with that approach I'm not aware of? Now validations/constraints ensure db consistency, but I can still follow a stale "new" link to a form that I will never be able to save after I fill it in, right?

app/models/pageflow/account.rb
```diff
@@ -30,6 +30,26 @@ def first_paged_entry_template
...
+      potential_entry_templates = free_type_names.map do |type_name|
+        EntryTemplate.new(
+          account_id: id,
+          entry_type: type_name
+        )
+      end
+
+      allowed_existing_entry_templates + potential_entry_templates
+    end
+
```
tf, line 52: This looks a bit intimidating and should thus have a test.

lib/pageflow/global_config_api.rb
```diff
@@ -44,17 +44,11 @@ def configure(&block)
...
-
-      if target.respond_to?(:entry_type)
-        config = Configuration::ConfigView.new(config, target.entry_type)
+      if target.is_a?(EntryTemplate)
+        config_for(Account.find(target.account_id))
+      else
+        config_for_account_or_entry(target)
       end
-
-      config
```
tf, line 57: This does not use the entry template's entry type yet to scope, for example, widget types by entry type. If we want to save that for a later step, instead of doing this change we could also do `config_for(entry_template.account)` in the [one place where it is used](https://github.com/codevise/pageflow/pull/1420/files#diff-b378d3e9e492d24e8547d4d5fc204542R2). Otherwise it needs a spec.

spec/controllers/admin/entry_templates_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+                  theme_name: 'unregistered'
+                }
+              })
+
+        expect(response.body).to have_selector(
+          '.input.error#entry_template_theme_name_input'
+        )
+      end
+    end
+  end
```
tf, line 45: I'd look a the account admin spec for more things to test here. Code related to updating nested entry templates can in turn be removed there.

admins/pageflow/entry_templates.rb
```diff
@@ -0,0 +1,66 @@
...
+        "#{I18n.t('activerecord.values.pageflow/entry.type_names.' + entry_type_name)}"
+      end
+
+      def permitted_params
+        params.permit(entry_template: permitted_entry_template_attributes)
+      end
+
+      def permitted_entry_template_attributes
+        [
+          :account_id,
```
tf, line 49: Is this needed? If yes, we should have a spec that ensures a user cannot create/update entry templates of other accounts.

admins/pageflow/entry_templates.rb
aviav: > I'm not aware of? 

Not sure. See [reply to your comment here](https://redmine.codevise.de/issues/17306#note-59).

Basically, I agree with your problem description. See my reply for more on this

aviav: I'll add that idea to the todo list in any case, so no need to discuss it right now

tf: Ah, I missed your comment on the issue. Since discussion moved on over there, I'll reply here.

> Otherwise, opening the edit form twice and saving twice would still create two entry templates for the same account/entry type

I don't think so. All forms always have action urls of the form `/account/5/entry_tempaltes/scrolled` or `/account/5/entry_tempaltes/paged`. The `find_resource` finds or initializes entry templates. Update always just calls save. So the second form submission would be an update either way. The second update would still override attributes with potentially stale data. But that's a problem we have with every admin form if you open multiple instances.

lib/pageflow/global_config_api.rb
aviav: To me it seems better to drop this for now, and do `config_for(entry_template.account)`

spec/controllers/admin/entry_templates_controller_spec.rb
aviav: Indeed. After working on the entry template admin for a while today, I noticed that some account controller specs should be ported to entry template controller before doing more in that direction

admins/pageflow/entry_templates.rb
aviav: Thinking about this some more, I'm not sure just deleting the existing data is an improvement. There is a clear validation error message which can prompt the user to either dismiss stale data, or use the edit link instead. It's definitely annoying to have to copy over the data, but we could also argue that we have some unnecessary duplication of data between entry templates which significantly contributes to this being an annoyance in the first place.

I'd even favor leaving this as it is, and if users request it or if we have time, make entry template entry UX more pleasant by avoiding duplication and, if it then still seems like a nuisance, go with the approach to only have edit/update actions.

Would it be okay with you to leave this as it is in the first iteration?

aviav: PS: Maybe in this case it would make more sense to label the 'Edit' link for the create action with 'Create', and also let the forms' page titles differ somewhat

tf: Sounds good. That way there is also an easy way to see that an account is still missing an (explicitly configured) entry template.

aviav: Alright! Updated my todos accordingly

aviav:
> Looks good. One stylistic quibble: I would have vertically aligned the edit/new links with the entry type name in the entry template table, but that's definitely not a blocker. Would click merge now or is there anything still WIP about this?

Can be merged.

Yes, aligning those items seems like a good idea. I wasn't sure with IE and flex layout, so that's why I skipped it

PR #1436, opened 2020-05-27
Remove css custom properties

UsmanMuhammad:
[REDMINE#17372](https://redmine.codevise.de/issues/17372)

(4/6)

- [X] Remove CSS Custom Properties that manipulates `react-compare-image` to inline styles

> Since custom properties are not supported in IE11, removed them and replaced them with static values.


UsmanMuhammad:
@tf , can you please approve this percy issue and pass the test? 
Label max-width is not necessary so I deleted, same was done in [PR](https://github.com/codevise/pageflow/pull/1433) but that was passed.

PR #1446, opened 2020-06-02
Widgets for scrolled entries

aviav:
REDMINE-17226

### DIFF
--- a/entry_types/scrolled/package/src/frontend/Entry.js
+++ b/entry_types/scrolled/package/src/frontend/Entry.js
@@ -6,6 +6,8 @@ import ScrollToSectionContext from './ScrollToSectionContext';
 import {useEntryStructure} from '../entryState';
 import {withInlineEditingDecorator} from './inlineEditing';
 import {usePostMessageListener} from './usePostMessageListener';
+import {useSectionChangeEvents} from './useSectionChangeEvents';
+import {sectionChangeMessagePoster} from './sectionChangeMessagePoster';
 
 import styles from './Entry.module.css';
 
@@ -16,11 +18,10 @@ export default withInlineEditingDecorator('EntryDecorator', function Entry(props
 
   const entryStructure = useEntryStructure();
 
-  const setCurrentSectionIndex = useCallback(index => {
-    if (window.parent !== window) {
-      window.parent.postMessage({type: 'CHANGE_SECTION', payload: {index}}, window.location.origin);
-    }
+  useSectionChangeEvents(entryStructure, currentSectionIndex);
 
+  const setCurrentSectionIndex = useCallback(index => {
+    sectionChangeMessagePoster(index);
     setCurrentSectionIndexState(index);
   }, [setCurrentSectionIndexState]);
 
--- a/entry_types/scrolled/package/src/frontend/registerTemplateWidgetType.js
+++ b/entry_types/scrolled/package/src/frontend/registerTemplateWidgetType.js
@@ -1,0 +1,4 @@
+export default function(typeName, callback) {
+  let element = document.getElementById('template-widget-container');
+  callback(element);
+}
--- a/entry_types/scrolled/spec/controllers/pageflow_scrolled/entries_controller_spec.rb
+++ b/entry_types/scrolled/spec/controllers/pageflow_scrolled/entries_controller_spec.rb
@@ -1,6 +1,9 @@
 require 'spec_helper'
 
 require 'pageflow/entries_controller_test_helper'
+require 'pageflow/test_widget_type'
+
+require 'pageflow/matchers/have_meta_tag'
 
 module PageflowScrolled
   RSpec.describe EntriesController, type: :controller do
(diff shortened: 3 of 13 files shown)


entry_types/scrolled/package/src/frontend/Entry.js
```diff
@@ -17,12 +18,20 @@ export default withInlineEditingDecorator('EntryDecorator', function Entry(props
...
       window.parent.postMessage({type: 'CHANGE_SECTION', payload: {index}}, window.location.origin);
     }
 
+    events.trigger('page:change', {configuration: {
+      title: getSectionChapterTitle(index, ', ') + 'Section ' + index
+    }, index: index});
+
     setCurrentSectionIndexState(index);
-  }, [setCurrentSectionIndexState]);
+  }, [entryStructure, setCurrentSectionIndexState]);
```
tf, line 34: We can test that the desired events with the desired data are being triggered using a similar test like [the one for the `CHANGE_SECTION` message](https://github.com/codevise/pageflow/blob/5a650782dff8f58252ea084e0bd55d8eafca49b5/entry_types/scrolled/package/spec/frontend/features/changeSectionMessage-spec.js#L4).

Awais also brought to my attention that the callback might be invoked multiple times in certain scenarios. To prevent triggering multiple events, we could use the [technique described in the hooks FAQ](https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state) to track the previous section index and only trigger events when the index changes. We already have a [`usePrevious` hook in our code base](https://github.com/codevise/pageflow/blob/5a650782dff8f58252ea084e0bd55d8eafca49b5/entry_types/scrolled/package/src/frontend/usePrevious.js#L3) that we can use here. Let's also extract this logic into a separate file like `useSectionChangeEventEmitter` etc.

entry_types/scrolled/package/src/frontend/registerTemplateWidgetType.js
```diff
@@ -0,0 +1,4 @@
+export default function(typeName, callback) {
+  let element = document.getElementById('template-widget-container');
+  callback(element);
+}
```
tf, line 4: In Pageflow Paged, one of the main responsibilities of `widgetTypes.register` is to ensure that only the `enhance` method of those widget types are called that are actually present in the document. If we introduce a Scrolled counterpart, we should make sure the same is true.

Alternatively, we can remove the function completely and make sure our existing widgets check themselves whether there rendered output is present.

entry_types/scrolled/spec/controllers/pageflow_scrolled/entries_controller_spec.rb
```diff
@@ -10,13 +11,29 @@ module PageflowScrolled
...
+        pageflow_configure do |config|
+          config.widget_types.register(widget_type)
+        end
+
+        entry = create(:entry, :published, type_name: 'scrolled')
+        create(:widget, subject: entry.published_revision, type_name: 'test_widget')
+
+        get_with_entry_env(:show, entry: entry)
+
+        expect(response.body).to have_selector('div.test_widget')
```
tf, line 35: This could also cover head fragments.

entry_types/scrolled/package/src/frontend/registerTemplateWidgetType.js
aviav: As discussed, for now we go with the alternative you describe in the second paragraph. Up until now, I think there aren't any more widgets for which we'd like scrolled compatibility.

entry_types/scrolled/package/src/frontend/Entry.js
```diff
@@ -16,13 +19,22 @@ export default withInlineEditingDecorator('EntryDecorator', function Entry(props
...
     }
-
-    setCurrentSectionIndexState(index);
-  }, [setCurrentSectionIndexState]);
+  }, [
+    previousSectionIndex,
+    sectionChangeEventEmitter,
+    sectionChangeMessagePoster,
+    setCurrentSectionIndexState
+  ]);
```
tf, line 37: `sectionChangeMessagePoster` is an extra dependency since it is not from scope of the component's function. There should be a linter warnings about this in your editor.

This change makes the `useCallback` no-op since `sectionChangeEventEmitter` is a new function every time.

This still emits multiple events when the section changes. React does not apply state changes synchronously. So `setCurrentSectionIndex` can be called multiple times while `index` is still the old value.

You need to wrap the code in an effect to make it run after state changes have been applied:

```
function useSectionChangeEvents(entryStructure, currentSectionIndex) {
  const previousSectionIndex = usePrevious(currentSectionIndex);

  useEffect(() => {
    if (previousSectionIndex !== currentSectionIndex) {
      events.trigger('page:change', ...)
    }
  }, [entryStructure, previousSectionIndex, currentSectionIndex]);
}
```

For the `CHANGE` post message, we currently do not care about duplicate messages. So you can leave it unchanged for now if you want.

entry_types/scrolled/package/src/frontend/useSectionChangeEventEmitter.js
```diff
@@ -0,0 +1,11 @@
+import {events} from 'pageflow/frontend';
+
+export const useSectionChangeEventEmitter = (entryStructure) => (index) => {
+  const getSectionChapterTitle = (sectionId, delimiter = '') => entryStructure.filter(
+    chapter => chapter.sections.some(section => section.id === sectionId)
+  )[0]?.title + delimiter;
+
+  events.trigger('page:change', {configuration: {
+    title: getSectionChapterTitle(index, ', ') + 'Section ' + index
+  }, index: index});
```
tf, line 10: Section id and index are two independent concepts. Try to extend the test to make it check the title of the chapter is looked up correctly and passed in the event.

entry_types/scrolled/package/src/frontend/Entry.js
aviav: I don't understand the first paragraph of your comment. There is no linter warning on `sectionChangeMessagePoster`. When I change that reference into an invalid one, e.g. `sectionChangeMessagePoster**s**`, there *is* a linter warning about an unnecessary dependency, so my linter-editor integration works in principle

tf: This is what I see:
```
$ yarn run eslint src/frontend/Entry.js                                                                                                  
yarn run v1.22.0
$ /home/tfischbach/code/gh/pageflow/node_modules/.bin/eslint src/frontend/Entry.js

/home/tfischbach/code/gh/pageflow/entry_types/scrolled/package/src/frontend/Entry.js
  32:6  warning  React Hook useCallback has an unnecessary dependency: 'sectionChangeMessagePoster'. Either exclud
e it or remove the dependency array. Outer scope values like 'sectionChangeMessagePoster' aren't valid dependencie
s because mutating them doesn't re-render the component  react-hooks/exhaustive-deps                             

```

aviav: Interesting. My system eslint is version 7.3.1, as opposed to the 6.x versions that this branch currently uses. Perhaps my editor uses system eslint then. I'll look into this after responding to the remainder of your comment

aviav: So, as soon as I removed the event-related code from the callback, I got the warning you mentioned about one of the two remaining mentioned dependencies. Still, I never observed exactly what you observed. It shall remain a mystery

PR #1459, opened 2020-06-04
Add imageMobile configuration for section

DanielKuhn:
REDMINE-17644
REDMINE-17676


DanielKuhn:
I just realized that my merge-logic for the background images now does not allow deleting a background image for the configuration. I'll look into that after lunch

PR #1474, opened 2020-06-12
Bless Pool Players through play button

awaismbajwa:
REDMINE-17702

### DIFF
--- a/entry_types/scrolled/package/src/frontend/MediaPlayer/playerActions.js
+++ b/entry_types/scrolled/package/src/frontend/MediaPlayer/playerActions.js
@@ -1,3 +1,4 @@
+import { media } from 'pageflow/frontend';
 
 export const PLAY = 'MEDIA_PLAY';
 export const PLAYING = 'MEDIA_PLAYING';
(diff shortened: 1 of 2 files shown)


entry_types/scrolled/package/src/frontend/PlayerControls/PlayPauseButton.js
```diff
@@ -29,6 +32,10 @@ function clickHandler(props) {
       props.pause();
     } else {
       props.play();
+      if (!isBlessed) {
+        media.mute(false);
+        isBlessed = true;
+      }
     }
```
tf, line 40: `PlayerControls` is a presentational/dumb component that so far does not know about concepts like media. There is also `MediaPlayerControls`, which wires up the `PlayerControls` with `playerState`/`playerActions`. I think, it would be a better fit for media related functionality. Can you move this to a separate PR so it can already be merged?

awaismbajwa: I thought this player actions.play might get invoked from other places. I have rebased this PR can be merged.

PR #1487, opened 2020-06-16
Add invert-option support for GradientBox

DanielKuhn:
REDMINE-17731

### DIFF
--- a/entry_types/scrolled/package/src/contentElements/heading/Heading.js
+++ b/entry_types/scrolled/package/src/contentElements/heading/Heading.js
@@ -6,10 +6,11 @@ import {Text} from 'pageflow-scrolled/frontend';
 
 import styles from './Heading.module.css';
 
-export function Heading({configuration}) {
+export function Heading({sectionProps, configuration}) {
   return (
-    <h1 className={classNames(styles.root, {[styles.first]: configuration.first},
-                              withShadowClassName)}>
+    <h1 className={classNames(styles.root,
+                              {[styles.first]: configuration.first},
+                              {[withShadowClassName]: !sectionProps.invert})}>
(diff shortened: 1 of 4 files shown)


entry_types/scrolled/package/src/contentElements/heading/Heading.js
```diff
@@ -6,10 +6,11 @@ import {Text} from 'pageflow-scrolled/frontend';
...
 import styles from './Heading.module.css';
 
-export function Heading({configuration}) {
+export function Heading({sectionProps, configuration}) {
   return (
-    <h1 className={classNames(styles.root, {[styles.first]: configuration.first},
-                              withShadowClassName)}>
+    <h1 className={classNames(styles.root,
+                              {[styles.first]: configuration.first},
+                              {[withShadowClassName]: !sectionProps.invert})}>
```
tf, line 13: We could also make `GradientBox` not set the `gradient` CSS class if the section is inverted. At the moment it does not matter, but if we ever use `withShadowClassName` somewhere else we will have to duplicate this logic.

DanielKuhn: But in that case we would also lose the `shadowDark` / `shadowLight` CSS.
Didn't we say that content elements should be able to choose whether they want text-shadow or not? I think this justifies keeping the choice within the `Heading` component.

tf: > Didn't we say that content elements should be able to choose whether they want text-shadow or not?

Yes, but so far they did not need to know the logic when it is applied. Not showing a shadow when the section is inverted is not heading specific logic IMO. It's not that important, though.

PR #1493, opened 2020-06-18
Switch player controls color from currentColor to white.

DanielKuhn:
Since the black controls on white background are not used,
we can safely switch to always use white color for controls.

REDMINE-17733


tf:
see #1505

PR #1499, opened 2020-06-22
Tweak preview animations for section transitions

UsmanMuhammad:
REDMINE-17721

- [X] Use correct animationf for "fadeBg". The preview animation scrolls the background instead of fading it.

> Changed the **_`translate`_** transformation to **_`opacity`_** and decrease it on hover for animation.

- [X] Center A/B letters in section box.

> Give the parent div a display: flex and then align items to center A/B letters.

- [X] Increase animation speed. In particular the fade animation takes a long time to start. "Reveal" is the only animation that feels fast enough for me. But even there I'd remove the easing at the end of the animation.

> Gave all the animations consistent speed and animation. (**_`2s  and linear`_**)

- [X] Make "scroll" and "scrollOver" animations go all the way. All other animations start in a state where only A is visible and finish with only B. "scroll" and "scrollOver" use a different style that just wiggles the animation a bit. I think it would be more consistent to also show the full transition here.

> Changed it, so the A covers and the transition goes till the end to show B to be consistent. 

- [X] Fix relative movement of A in "scroll" animation. The letter A moves slower than the section boundary.

> Already fixed when fixed the above mentioned point.

- [X] Make all animation loop. Some of the animations only run a single time and require leaving and entering the element again to trigger the animation another time.

> Also, related to the third point where I changed the speed and animation and made it consistent throughout.

- [X] I'd suggest using white instead of dark borders between the items to better match the look of the editor. The color of the selected item is also new. We could use selelectionColor from views/colors.module.css.

> Changed the colors as suggested.

- [X] Make whole item (including animated thumbnail) clickable - not only label.

> Enclosed the whole transition div into a label, and the label points to the respective input element with the _**`for`**_ attribute

### DIFF
--- a/entry_types/scrolled/package/src/editor/views/EditSectionTransitionEffectView.module.css
+++ b/entry_types/scrolled/package/src/editor/views/EditSectionTransitionEffectView.module.css
@@ -1,13 +1,23 @@
+@value selectionColor from './colors.module.css';
+
 .upper_section {
     background: #50514f;
     height: 100%;
     width: 100%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    font-weight: normal;
 }
 
 .lower_section {
     background: #A2A2A2;
     height: 100%;
     width: 100%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    font-weight: normal;
 }
 
 .disabled {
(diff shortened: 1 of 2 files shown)


entry_types/scrolled/package/src/editor/views/EditSectionTransitionEffectView.module.css
```diff
@@ -1,13 +1,23 @@
+@value selectionColor from './colors.module.css';
+
 .upper_section {
     background: #50514f;
     height: 100%;
     width: 100%;
+    display: flex;
```
UsmanMuhammad, line 7: To center the A/B letters. Same goes for lower_section.

entry_types/scrolled/package/src/editor/views/EditSectionTransitionEffectView.module.css
```diff
@@ -16,68 +26,52 @@
...
-    animation: Scroll 2s ease infinite;
-}
-
-/*Reveal*/
-.transition:hover .reveal .upper_section {
-    animation: Reveal 2s ease infinite;
+.transition:hover .scrollOver .lower_section,
+.transition:hover .reveal .upper_section,
+.transition:hover .fadeBg .lower_section {
+    animation: Translate 2s linear infinite;
```
UsmanMuhammad, line 61: Changed the Scroll keyframe name to more generic as Translate since a lot of different transitions use the same.
I did same for the Fade.

entry_types/scrolled/package/src/editor/views/EditSectionTransitionEffectView.module.css
```diff
@@ -97,71 +91,42 @@
...
-    z-index: 1;
-}
-
-/*Reveal Upper Section*/
-.reveal .upper_section {
     position: absolute;
     z-index: 1;
 }
 
-/*Before After Upper Section*/
```
UsmanMuhammad, line 113: All of this was not needed, and were using the same properties. Moved transition to :hover above where they now use keyframe transitions.

PR #1500, opened 2020-06-23
Setup media multi player atmo

awaismbajwa:
REDMINE-17587
REDMINE-17389

### DIFF
--- a/entry_types/scrolled/package/spec/frontend/features/sectionAtmoAudio-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/features/sectionAtmoAudio-spec.js
@@ -37,7 +37,6 @@ describe('section AtmoAudio', () => {
     };
   }
 
-
   it('is played when section becomes active', async () => {
     const {getSectionByPermaId} = renderEntry({
       seed: {
@@ -55,9 +54,13 @@ describe('section AtmoAudio', () => {
 
     await act(async () => {
       media.mute(false);
-      getSectionByPermaId(11).simulateScrollingIntoView();  
+      getSectionByPermaId(11).simulateScrollingIntoView();
     });
 
-    expect(media.getPlayer).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({filePermaId: 100}))
+    expect(media.getPlayer).toHaveBeenCalledWith(expect.anything(),
+                                                 expect.objectContaining({
+                                                   filePermaId: 100,
+                                                   tagName: 'audio'
+                                                 }));
   });
 });
--- a/entry_types/scrolled/package/spec/frontend/features/sectionAtmoAudio-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/features/sectionAtmoAudio-spec.js
@@ -1,0 +1,63 @@
+
+
+import {act} from '@testing-library/react';
+
+import {media} from 'pageflow/frontend';
+
+import {renderEntry, usePageObjects} from 'support/pageObjects';
+import 'support/fakeBrowserFeatures';
+import {useFakeMedia} from 'support/fakeMedia';
+import {fakeParentWindow} from 'support';
+
+
+describe('section AtmoAudio', () => {
+  usePageObjects();
+  useFakeMedia();
+
+  beforeEach(() => {
+    fakeParentWindow()
+  });
+
+  function getAudioFileSeed({
+    id = 1,
+    permaId = 100,
+    basename = 'audio'
+  } = {}) {
+    return {
+      fileUrlTemplates: {
+        audioFiles: {
+          mp3: ':id_partition/:basename.mp3',
+          m4a: ':id_partition/:basename.m4a',
+          ogg: ':id_partition/:basename.ogg'
+        }
+      },
+      audioFiles: [
+        {id, permaId, isReady: true, basename}
+      ]
+    };
+  }
+
+
+  it('is played when section becomes active', async () => {
+    const {getSectionByPermaId} = renderEntry({
+      seed: {
+        sections: [
+          {permaId: 10},
+          {permaId: 11, configuration: {transition: 'scroll', atmoAudioFileId: 100}}
+        ],
+        ...getAudioFileSeed({
+          id: 1,
+          permaId: 100,
+          basename: 'audio'
+        })
+      }
+    });
+
+    await act(async () => {
+      media.mute(false);
+      getSectionByPermaId(11).simulateScrollingIntoView();  
+    });
+
+    expect(media.getPlayer).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({filePermaId: 100}))
+  });
+});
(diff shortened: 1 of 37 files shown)


package/src/frontend/media/media.js
```diff
@@ -32,7 +45,32 @@ export const media = {
...
+    playFromBeginning: false,
+    rewindOnChange: true,
+    pauseInBackground: true
+  }
+});
+
+browser.detectFeatures().then(()=>{
+  media.atmo.disabled =  browser.has('mobile platform');
+});
+
```
tf, line 75: The most immediate problem, which is causing the specs to fail, is that we do not want to call `detectFeatures` here. This is easy to solve since there is no need to disable atmo on mobile devices. This was only necessary since we had no way to reliably auto play atmo audios. With media pool we now do. So this can just be removed.

The second problem is that we do not want to create these `Atmo` objects in Paged entries since they (at least for now) have their own atmo implementation. So we will have to move away from the `atmo` singleton. As noted above, the `Atmo` spec can be decoupled from the singleton by working with a local `Atmo` instance.

Finally, I'd suggest to separate atmo and media concerns more strictly. Before, the media module mainly had two responsibilities:

* Provide players 
* Handle unmuting

Now we've added

* Maintain a map of sources by id
* Provide atmo functionality

The only overlap between these two pieces of functionality is that `getPlayer` needs to access `atmo` to create hooks.  If we can remove this dependency, atmo can be extracted from `media` and live next to it in `src/frontend/Atmo.js`. Here's one way to do it:

* Introduce `AtmoProvider` component in `pageflow-scrolled/frontend` that
  * creates a SourceIdMap object
  * creates a MultiPlayer passing the SourceIdMap as pool
  * creates an `Atmo` object passing MultiPlayer
  * places a function just like `media.updateAtmo` on the context which stores the sources in the SourceIdMap and updates the atmo with the given id
  * places a function that calls `atmo.createMediaPlayerHooks` on the React context. 

* Extract the `onActivate` code in `Section` into a `AtmoAudio` component that is rendered in `Section` inside the `SectionLifecycleProvider`. It can then look up the `updateAtmo` function from the Reat context and the audio file via `useFile` and use the `onActivate`/`onDeactivate` options of the `useSectionLifecycle` to update the atmo. This also fixes the problem with the current code that the atmo is never faded out when a section without atmo audio is reached. The component can simply render `null`.

* In `PlayerContainer`, get the `atmo.createMediaPlayerHooks` function from the context and call it to obtain media player hooks to pass to `getPlayer`. That way `getPlayer` does not need to know about `atmoDuringPlayback` - only about hooks.

* Everything related to SourceIdMap and atmo can then be removed from `media`.

* In Scrolled, this could be tested with an integration/feature spec ([example](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/frontend/features/changeSectionMessage-spec.js)) that uses [`fake media`](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/support/fakeMedia.js) to stub out the actual media player code and test that `media.getPlayer` is called with the expected sources when a section when the user scrolls to a section with an atmo audio file.

package/src/frontend/media/playerSourceIDMap.js
```diff
@@ -0,0 +1,15 @@
...
+  return {
+
+    mapSources: function (id, sources) {
+      this[id] = sources;
+    },
+    get: function (sourceID, options) {
+      return media.getPlayer(this[sourceID], {
+        filePermaId: sourceID,
+        ...options
+      });
```
tf, line 12: `MultiPlayer` and this class will somehow have to collaborate to release players.

entry_types/scrolled/package/spec/frontend/features/sectionAtmoAudio-spec.js
```diff
@@ -0,0 +1,55 @@
...
+      audioFiles: [
+        {id, permaId, isReady: true, basename}
+      ]
+    };
+  }
+
+
+  it('is posted when section becomes active', () => {
+    const {getSectionByPermaId} = renderEntry({
+      seed: {
```
awaismbajwa, line 38: @tf can you help me prepare seed data for section with atmo audio file ?

tf: `atmoAudioFileId` needs to go in a `configuration` property:

```
{permaId: 11, configuration: {atmoAudioFileId: 100}}
```

package/src/frontend/media/playerSourceIDMap.js
awaismbajwa: As discussed I have introduced a mechanism in MultiPlayer pool i.e. playerSourceIDMap.
Now this pool maintains current and previous player. If there are two sources than pool 
keeps returning current and previous player. When third source is requested the first player is released and new one is requested.

entry_types/scrolled/package/src/frontend/AtmoAudio.js
```diff
@@ -0,0 +1,37 @@
...
+      if (audioFile && audioFile.isReady) {
+        sources = processSources(audioFile);
+      }
+      
+      atmo.updateAtmo({
+        sources: sources,
+        atmoAudioFileId: props.atmoAudioFileId
+      });
+    }
+  });
```
tf, line 33: Atmo audio also needs to be updated when the file changes. Currently selecting an atmo audio file does not start the audio. Only when I deactivate and activate the section again the atmo starts.

You can use a combination of `usePrevious` and an effect that compares audio files to also trigger `updateAtmo` when the atmo changes.

package/src/frontend/media/createMediaPlayer.js
```diff
@@ -42,6 +43,8 @@ export const createMediaPlayer = function (options) {
     }
   };
 
+  AudioPlayer.rewindMethod(player);
+
```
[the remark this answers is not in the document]
awaismbajwa: Sorry somehow these conversations were being shown in hidden section and got ignored.
should I copy same method from AudioPlayer to VideoPlayer?

package/src/frontend/media/MultiPlayer.js
```diff
@@ -115,30 +130,20 @@ export const MultiPlayer = function(pool, options) {
...
-
-    that.listenTo(player, 'playfailed', function() {
-      that.trigger('playfailed', {audioFileId: id});
-    });
+  function startEventPropagation(player) {
+    player.on('play', playCallback);
+    player.on('pause', pauseCallback);
+    player.on('timeupdate', timeUpdateCallback);
+    player.on('ended', endedCallback);
+    player.on('playfailed', playFailedCallback);
```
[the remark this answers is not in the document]
awaismbajwa: does this issue still occur? since now currentId is updated with new player and then event propagations starts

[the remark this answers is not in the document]
tf: I've never really seen the issue, but from looking at the code, I think it is still a problem: If cross fade is enabled, the new player and the event propagation for the new player start while the old player is still playing. CurrentId changes soon as the new player starts. This now causes events triggered for the old player to already report the new id, where before the events of the old player still contained the correct, old id.

PR #1506, opened 2020-06-25
Display black poster in empty inline video

tf:
Prevent newly inserted video content element from collapsing.

REDMINE-17742


tf:
[Travid build](https://travis-ci.org/github/codevise/pageflow/builds/702080538) passed but failed to report back to PR.

PR #1510, opened 2020-06-30
Extend editor api to detect browser

UsmanMuhammad:
REDMINE-17565

- [X] Extend editor in pageflow/editor API to let entry types provide: added

> - [X] A function that is called in startEditor to check whether the current browser is supported added
> - [X] A view that is rendered instead of the editor if the function returns false.

- Created a new view, which will be rendered if the browser is not supported.
- Provide a function in `entry_type/scrolled/package/src/editor` to detect the type of browser and a view to display.
- Extended the `pageflow/editor` api to include a callback function which will get the result from scroll and either then start the app or display the view.

### DIFF
--- a/doc/creating_entry_types.md
+++ b/doc/creating_entry_types.md
@@ -190,6 +190,24 @@ const appearanceInputs = (tabView, _options) => {
 Pageflow is looking for translation keys of the form
 `pageflow.entry_types.paged.editor.entry_metadata_configuration_attributes.manual_start.{inline_help,label}`.
 
+The `browserNotSupportedView` is a Backbone view that will be displayed if the access to editor needs to be blocked due to browser compatibility issues.
+```javascript
+import {editor} from 'pageflow/editor;
+import {browser} from 'pageflow/frontend';
+
+import {BrowserNotSupportedView} from './views/BrowserNotSupportedView';
+
+editor.registerEntryType('rainbow', {
+  // ...
+
+  isBrowserSupported() {
+    return browser.has('some feature required to use the editor');
+  },
+
+  browserNotSupportedView: BrowserNotSupportedView
+});
+``` 
+
 ### REST Controllers
 
 Entry types can define new editor controllers, which can then be used
--- a/entry_types/scrolled/package/src/editor/config.js
+++ b/entry_types/scrolled/package/src/editor/config.js
@@ -9,6 +9,9 @@ import {EntryPreviewView} from './views/EntryPreviewView';
 import {SideBarRouter} from './routers/SideBarRouter';
 import {SideBarController} from './controllers/SideBarController';
 
+import {browser} from 'pageflow/frontend';
+import {BrowserNotSupportedView} from './views/BrowserNotSupportedView';
+
 editor.registerEntryType('scrolled', {
   entryModel: ScrolledEntry,
 
@@ -18,7 +21,12 @@ editor.registerEntryType('scrolled', {
       editor
     });
   },
-  outlineView: EntryOutlineView
+  outlineView: EntryOutlineView,
+
+  isBrowserSupported() {
+    return (!browser.agent.matchesMobilePlatform());
+  },
+  browserNotSupportedView: BrowserNotSupportedView
 });
 
 editor.registerSideBarRouting({
--- a/package/src/editor/api/index.js
+++ b/package/src/editor/api/index.js
@@ -92,6 +92,10 @@ export const EditorApi = Object.extend(
    *   Backbone view that will render the live preview of the entry.
    * @param {function} options.EntryOutlineView
    *   Backbone view that will be rendered in the side bar.
+   * @param {function} options.isBrowserSupported
+   *  Checks to see if the browser is supported.
+   * @param {function} options.browserNotSupportedView
+   *  Backbone view that will be rendered if the browser is not supported.
    */
   registerEntryType(name, options) {
     this.entryType = {name, ...options};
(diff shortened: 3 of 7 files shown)


entry_types/scrolled/package/src/editor/config.js
```diff
@@ -9,6 +9,9 @@ import {EntryPreviewView} from './views/EntryPreviewView';
 import {SideBarRouter} from './routers/SideBarRouter';
 import {SideBarController} from './controllers/SideBarController';
 
+import {agent} from '../../../../../package/src/frontend/browser/Agent';
```
tf, line 12: Relative imports are for package internal things. `pageflow-scrolled` should neither need to know where the `pageflow` package lives on the file system nor its internal directory structure. Both is hard coded here. Since `pageflow/frontend` exports `browser` ([1](https://github.com/codevise/pageflow/blob/49e310350cabd8e78e5f349034521b44fb79812a/package/src/frontend/index.js#L7), [2](https://github.com/codevise/pageflow/blob/49e310350cabd8e78e5f349034521b44fb79812a/package/src/frontend/browser/index.js#L18), [3](https://github.com/codevise/pageflow/blob/49e310350cabd8e78e5f349034521b44fb79812a/package/src/frontend/browser/browser.js#L7)) and [assigns `browser.agent`](https://github.com/codevise/pageflow/blob/49e310350cabd8e78e5f349034521b44fb79812a/package/src/frontend/browser/index.js#L20), you should be able to do:

```
import {browser} from 'pageflow/scrolled`;
browser.agent.matchesSafari();
```

entry_types/scrolled/package/src/editor/config.js
```diff
@@ -18,7 +21,12 @@ editor.registerEntryType('scrolled', {
       editor
     });
   },
-  outlineView: EntryOutlineView
+  outlineView: EntryOutlineView,
+
+  isBrowserSupported() {
+    return (agent.matchesSafari() || agent.matchesChrome());
```
tf, line 27: Maybe we can start with just excluding mobile browsers (e.g. `!matchesMobilePlatform`). Then we can fine tune the exact matchers in a later step while still being able to continue testing the editor in all desktop browsers once this is merged.

package/src/editor/api/index.js
```diff
@@ -301,5 +301,27 @@ export const EditorApi = Object.extend(
...
+   * Ensures that Editor only runs on modern browsers.
+   *
+   * @callback start
+   *  callback function to start the app.
+   * @function browserNotSupportedView
+   *  View to display in case of a not supporting browser.
+   * @function isBrowserSupported
+   *  Checks to see if the browser is supported.
+  */
+  ensureBrowserSupport: function(start) {
```
tf, line 316: Since this method is not part of the public API (it is only ever called by `startEditor`), it does not need inline doc. The public API we extended actually is [`registerEntryType`](https://github.com/codevise/pageflow/blob/49e310350cabd8e78e5f349034521b44fb79812a/package/src/editor/api/index.js#L96). Let's document these options there.

doc/creating_entry_types.md
```diff
@@ -187,6 +187,24 @@ const appearanceInputs = (tabView, _options) => {
...
+  // ...
+
+  isBrowserSupported() {
+    return browser.has('some feature required to use the editor');
+  },
+
+  browserNotSupportedView: BrowserNotSupportedView
+});
+``` 
+
```
tf, line 207: The text here is inserted in the middle of a sentence.

package/src/editor/api/index.js
tf: I should have written: Since this method is not part of the public API, it is not supposed to show up in the [API docs](https://codevise.github.io/pageflow-docs/js/master/index.html) and thus must not have an inline comment (since everything with an inline comment shows up in the docs.

package/src/editor/api/index.js
```diff
@@ -301,5 +305,29 @@ export const EditorApi = Object.extend(
...
+  /**
+   * Ensures that Editor only runs on modern browsers.
+   *
+   * @param {function} start
+   *  A callback function to start the app.
+  */
+  ensureBrowserSupport: function(start) {
+    if (this.entryType.isBrowserSupported) {
+      const isBrowserSupported = this.entryType.isBrowserSupported();
+      const browserNotSupportedView = new this.entryType.browserNotSupportedView;
```
tf, line 319: Missing "()" when invoking a constructor. This should show a linter warning in your editor. TIL this is not a syntax error.

Also, we could consider only creating the view when it is actually needed.

UsmanMuhammad:
Thank you for the suggestions once again, I will try to be more careful in future and plan my commits beforehand. I took your last paragraph by mistake and put everything in one commit. But, I realized afterwards that I should have put all the others in one and changes you last requested in another.
I will point out the changes in the code for now, so you can review them. In the future I will be more careful with reviewed changes, squashing non useful commits and planning what should be in one commit and separated.

doc/creating_entry_types.md
```diff
@@ -190,6 +190,24 @@ const appearanceInputs = (tabView, _options) => {
 Pageflow is looking for translation keys of the form
 `pageflow.entry_types.paged.editor.entry_metadata_configuration_attributes.manual_start.{inline_help,label}`.
 
+The `browserNotSupportedView` is a Backbone view that will be displayed if the access to editor needs to be blocked due to browser compatibility issues.
```
UsmanMuhammad, line 193: Text was between a sentence, so I moved it at the end of topic 'Backbone Integration.

PR #1512, opened 2020-07-03
Extend the existing user agent detection code for desktop browsers

UsmanMuhammad:
REDMINE-17565
We want to restrict entry to editor on certain devices and browsers.

- [X] Extend the existing user agent detection code to cover all cases we need to recognize.

> Extend the Agent API to cover four most used desktop browsers.

- [X] Add tests with example user agent codes.

> Extend the agent-specs to cover all the newly added cases.

### DIFF
--- a/package/spec/frontend/browser/agent_spec.js
+++ b/package/spec/frontend/browser/agent_spec.js
@@ -64,7 +64,7 @@ describe('pageflow.browser.Agent', function() {
         'AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15'
       );
 
-      expect(agent.matchesDesktopSafari({minVersion: 3})).toBe(true);
+      expect(agent.matchesDesktopSafari()).toBe(true);
     });
 
     it('returns false for Safari on iPhone', function() {
@@ -73,7 +73,16 @@ describe('pageflow.browser.Agent', function() {
         'AppleWebKit/602.1.50 (KHTML, like Gecko) Version/10.0 Mobile/14A403 Safari/602.1'
       );
 
-      expect(agent.matchesDesktopSafari({minVersion: 3})).toBe(false);
+      expect(agent.matchesDesktopSafari()).toBe(false);
+    });
+
+    it('returns false for Safari 3.2 if minVersion 4 is given', function() {
+      var agent = new Agent(
+        'Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10_5_6; en-us) '+
+        'AppleWebKit/525.27.1 (KHTML, like Gecko) Version/3.2.1 Safari/525.27.1'
+      );
+
+      expect(agent.matchesDesktopSafari({minVersion: 4})).toBe(false);
     });
   });
 
@@ -84,13 +93,22 @@ describe('pageflow.browser.Agent', function() {
         'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36 Edg/83.0.478.58'
       );
 
-      expect(agent.matchesDesktopEdge({minVersion: 20})).toBe(true);
+      expect(agent.matchesDesktopEdge()).toBe(true);
     });
 
     it('returns false for Edge on Windows 10 mobile', function() {
       var agent = new Agent(
         'Mozilla/5.0 (Windows Mobile 10; Android 10.0; Microsoft; Lumia 950XL) '+
         'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Mobile Safari/537.36 Edge/40.15254.603'
+      );
+
+      expect(agent.matchesDesktopEdge()).toBe(false);
+    });
+
+    it('returns false for Edge 12 if minVersion 20 is given', function() {
+      var agent = new Agent(
+        'Mozilla/5.0 (Windows NT 10.0) '+
+        'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.135 Safari/537.36 Edge/12.10136'
       );
 
       expect(agent.matchesDesktopEdge({minVersion: 20})).toBe(false);
@@ -104,7 +122,7 @@ describe('pageflow.browser.Agent', function() {
         'Gecko/20100101 Firefox/78.0'
       );
 
-      expect(agent.matchesDesktopFirefox({minVersion: 20})).toBe(true);
+      expect(agent.matchesDesktopFirefox()).toBe(true);
     });
 
     it('returns false for SeaMonkey', function() {
@@ -113,7 +131,7 @@ describe('pageflow.browser.Agent', function() {
         'Gecko/20100101 Firefox/29.0 SeaMonkey/2.26'
       );
 
-      expect(agent.matchesDesktopFirefox({minVersion: 20})).toBe(false);
+      expect(agent.matchesDesktopFirefox()).toBe(false);
     });
 
     it('returns false for Firefox on android', function() {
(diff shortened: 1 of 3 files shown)


package/src/frontend/browser/Agent.js
```diff
@@ -86,6 +86,44 @@ export const Agent = function(userAgent) {
...
+    },
+    
+    /**
+   * Returns true on Google Chrome.
+   * @return {boolean}
+   */
+    matchesChrome: function() {
+      return matches(/Chrome\//i) &&
+             !matches(/Chromium/i);
+    },
```
tf, line 110: This will also be true in Edge. I think it's fine to support Chromium since it's pretty much the same browser?

tf: This will also be true in Edge.

package/src/frontend/browser/Agent.js
```diff
@@ -86,6 +86,43 @@ export const Agent = function(userAgent) {
...
+      return this.matchesChrome() && !this.matchesMobilePlatform() && minVersion(/Chrome\/(\d+)/i, 25);
+    },
+
+    matchesDesktopFirefox: function() {
+      return this.matchesFirefox() && !this.matchesMobilePlatform() && minVersion(/Firefox\/(\d+)/i, 20);
+    },
+    
+    matchesDesktopEdge: function() {
+      return this.matchesEdge() && !this.matchesMobilePlatform() && minVersion(/Edg\/(\d+)/i, 20);
+    },
```
tf, line 101: Let's make `minVersion` a parameter of the function to be more useful and not require a min version by default. This will also make the tests more explicit since they will not depend on the hard coded values in these functions. We can instead move these min versions to the `isBrowserSupported` function of the scrolled editor.

UsmanMuhammad: You mean like this?
```
matchesDesktopChrome: function(minVersion) {
      return this.matchesChrome() && !this.matchesMobilePlatform() && matchesMinVersion(/Chrome\/(\d+)/i, minVersion);
    },
```

Or there could be one other option to be separate functions for separate browsers like:
```
matchesDesktopChrome: function(minVersion) {
      return this.matchesChrome() && !this.matchesMobilePlatform() && matchesMinChromeVersion(/Chrome\/(\d+)/i, minVersion);
    },
```

tf: Think about it from the perspective of the client - the code that uses `browser.agent`. If I call `agent.matchesDesktopChrome()`, I'd expect it to return true for any version of Chrome on Desktop. I'd pass the min version in an options object: `agent.matchesDesktopChrome({minVersion: 25})` instead of `agent.matchesDesktopChrome(25)`. The latter could be misread as requiring exactly version 25. Also this makes it easier to add options in the future if we have to.

package/src/frontend/browser/Agent.js
```diff
@@ -97,6 +142,26 @@ export const Agent = function(userAgent) {
...
+    }
+    //Safari
+    if (match[0].includes('Version')) {
+      return match && match[1] && parseInt(match[1], 10) >= version;
+    }
+    //Firefox
+    if (match[0].includes('Firefox')) {
+      return match && match[1] && parseInt(match[1], 10) >= version;
+    }    
+  }
```
tf, line 164: All these cases do the same.

package/src/frontend/browser/Agent.js
tf: Why is this resolved?

package/src/frontend/browser/Agent.js
UsmanMuhammad: It does, and basically it's the same as the existing `captureGroupGreaterOrEqual` function. 
So I will replace that.

package/src/frontend/browser/Agent.js
UsmanMuhammad: Forgot to add `!matches(/Edg/i)`. Just added that.

package/spec/frontend/browser/agent_spec.js
```diff
@@ -73,7 +73,123 @@ describe('pageflow.browser.Agent', function() {
         'AppleWebKit/602.1.50 (KHTML, like Gecko) Version/10.0 Mobile/14A403 Safari/602.1'
       );
 
-      expect(agent.matchesDesktopSafari()).toBe(false);
+      expect(agent.matchesDesktopSafari({minVersion: 3})).toBe(false);
+    });
```
tf, line 77: Can you add tests for the case when `minVersion` option is not present? I'm not sure, but I think `matchesDesktopSafari` now always returns false if the version option is not present.

tf:
Also can you update the `isBrowserSupported` function to use these new matchers?

package/spec/frontend/browser/agent_spec.js
UsmanMuhammad: It does return false. Added tests for that.

tf: But that's wrong, isn't it? It *is* a desktop Safari, we just did not give a min version.

tf: [Existing code like this](https://github.com/codevise/pageflow/blob/f382c40084e3bff3413ceef0e2864179af57b0b6/package/src/frontend/browser/volumeControlSupport.js#L9) depends on `agent.matchesDesktopSafari()` returning `true` on a Desktop Safari. With your change, this call results in an error now. And even when passing `agent.matchesDesktopSafari({})` it now returns `false`. The test suite no longer shows that since you also changed the tests that have nothing to do with min version to pass that option as well.

I'd remove the option from all other tests and move the tests from the `describe('#minVersion', ...)` block into the describe blocks for the individual methods. The `#` normally means that the following string is the name of a method on the tested class (e.g. `Agent#matchesDesktopEdge`) - but there is no `Agent#minVersion` method.  If you look at those spec descriptions as a sentence, you see that they do not capture why the return value is the way it is:

"pageflow.browser.Agent#minVersion returns true for Chrome 83"

Instead I'd go for something like

"pageflow.browser.Agent#matchesDesktopFirefox returns false for Firefox 17 when minVersion 20 is given"

UsmanMuhammad: You mean to change `matchesMinVersion` in `Agent` , like to add a condition in `matchesMinVersion` , that if `minVersion` option is not given then there should be a default min version value and then `machesDesktopSafari` will always return true?

package/src/frontend/browser/Agent.js
```diff
@@ -86,16 +88,68 @@ export const Agent = function(userAgent) {
...
   }
 
-  function captureGroupGreaterOrEqual(exp, version) {
+  function matchesMinVersion(exp, version) {
     var match = userAgent.match(exp);
-    return match && match[1] && parseInt(match[1], 10) >= version;
+    return match && match[1] && parseInt(match[1], 10) >= version;   
+  }
+
+  //to detect iPad with Ios 13 and above
```
tf, line 148: This comment is redundant. It might be nice to have an explanation here (like you did in the commit message), but this just repeats the method name.

package/src/frontend/browser/Agent.js
```diff
@@ -86,16 +88,68 @@ export const Agent = function(userAgent) {
...
+  function matchesiPadSafari13AndAbove() {
+    return navigator.maxTouchPoints > 1 &&
+           navigator.platform === 'MacIntel' &&
+           !window.MSStream; //IE exclusion from being detected as an iOS device
```
tf, line 152: Please move this to a separate PR. Also the commit that adds this method also changes `isBrowserSupported`. So, for example, if I wanted to cherry pick the commit to already deploy it for Hosted even though Hosted does not use master yet, it would not apply without conflict.

To prevent issues like the IE thing, let's make sure this check only takes effect when we have already detected that the user agent string looks like a Safari (either because it is a Desktop Safari or because it is a new iPad).

UsmanMuhammad: removed from this PR

package/src/frontend/browser/Agent.js
UsmanMuhammad: Will update the commit in the new PR

package/spec/frontend/browser/agent_spec.js
UsmanMuhammad: Yes, I should have considered the fact that it will affect the existing code where theses functions are being used.
Now, I took care of the possibility that if no `minVersion` is given that `agent.mathcesDesktopSafari` should return true.

Removed the option from existing test cases and have moved `minVersion` tests in their respective methods.

PR #1515, opened 2020-07-06
Prevent progress bar with more than 100%

tf:
There appear to be race conditions where `progress` events are
dispatched when the duration of media player has not been updated
yet.

REDMINE-17761


tf:
[Travis build](https://travis-ci.org/github/codevise/pageflow/builds/705440202) passed but failed to report back to PR.

PR #1519, opened 2020-07-08
Add logic to detect IOS 13 Safari on iPad

UsmanMuhammad:
REDMINE-17737

After IOS 13 update, iPad reports the same user string
as Safari on Dekstop MacOS

It adds further support of detecting multi touch devices

At the time of this writing there are no other devices
with multi-touch support other than IOS/iPadOS

Also, adds a check for IE devices to be detected as
IOS devices

### DIFF
--- a/package/src/frontend/browser/Agent.js
+++ b/package/src/frontend/browser/Agent.js
@@ -178,8 +178,9 @@ export const Agent = function(userAgent) {
   //with multi-touch support other than IOS/iPadOS
   //See: https://stackoverflow.com/a/58064481
   function matchesiPadSafari13AndAbove() {
-    return navigator.maxTouchPoints > 1 &&
-            navigator.platform === 'MacIntel';       
+    return agent.matchesSafari() &&
+           navigator.maxTouchPoints > 1 &&
+           navigator.platform === 'MacIntel';       
   }
 };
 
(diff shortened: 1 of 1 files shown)


package/src/frontend/browser/Agent.js
```diff
@@ -168,6 +171,16 @@ export const Agent = function(userAgent) {
...
+
+  //After ios13 update, iPad reports the same user string
+  //as Safari on Dekstop MacOS.
+  //At the time of this writing there are no other devices
+  //with multi-touch support other than IOS/iPadOS
+  //See: https://stackoverflow.com/a/58064481
+  function matchesiPadSafari13AndAbove() {
+    return navigator.maxTouchPoints > 1 &&
+            navigator.platform === 'MacIntel';       
+  }
```
tf, line 183: Let's also check the user agent string claims to be a Safari. Otherwise any browser on Mac will become an iPad soon as there is a Mac with touch input, right?

UsmanMuhammad: We want to check that above in matchesMobileSafari(). Right? Any other browser will return true for matchesMobileSafari() on IOS.

tf: > Any other browser will return true for matchesMobileSafari() on IOS.

I don't understand.

UsmanMuhammad: I think i misunderstood, i will add the check for Safari here.

PR #1520, opened 2020-07-13
Media player context data

awaismbajwa:
REDMINE-17754
REDMINE-17790

### DIFF
--- a/entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
@@ -5,6 +5,7 @@ import {useFakeMedia, fakeMediaRenderQueries} from 'support/fakeMedia';
 import {render as testingLibraryRender} from '@testing-library/react'
 import {media} from 'pageflow/frontend';
 import {MediaPlayer} from 'frontend/MediaPlayer';
+import {EventContext} from 'frontend/useEventContextData';
 
 describe('MediaPlayer', () => {
   useFakeMedia();
@@ -84,23 +85,22 @@ describe('MediaPlayer', () => {
       expect.anything()
     );
   });
-  
-  it('it passes correct context data to getPlayer options ', () => {
+
+  it('it passes data from EventContext to getPlayer as media events context data', () => {
+    const eventContextData = {some: 'data'};
+
     render(<MediaPlayer {...requiredProps()}
                         type={'video'}
-                        sources={getVideoSources()} />);
+                        sources={getVideoSources()} />,
+           {
+             wrapper: ({children}) =>
+               <EventContext.Provider value={eventContextData}>{children}</EventContext.Provider>
+           });
 
     expect(media.getPlayer).toHaveBeenCalledWith(
       expect.anything(),
       expect.objectContaining({
-        mediaEventsContextData: {
-          page: {
-            configuration: {
-              title: undefined
-            },
-            index: -1,
-          }
-        }
+        mediaEventsContextData: eventContextData
       }
     ));
   });
--- a/entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
@@ -84,6 +84,26 @@ describe('MediaPlayer', () => {
       expect.anything()
     );
   });
+  
+  it('it passes correct context data to getPlayer options ', () => {
+    render(<MediaPlayer {...requiredProps()}
+                        type={'video'}
+                        sources={getVideoSources()} />);
+
+    expect(media.getPlayer).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({
+        mediaEventsContextData: {
+          page: {
+            configuration: {
+              title: undefined
+            },
+            index: -1,
+          }
+        }
+      }
+    ));
+  });
 
   it('does not request new Player for same sources', () => {
     const {rerender} =
--- a/entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
@@ -6,7 +6,6 @@ import 'support/fakeBrowserFeatures';
 import {renderHookInEntry} from 'support';
 
 describe('useEventContextData', () => {
-
   it('returns an object with current section and chapter index', () => {
     const {result} = renderHookInEntry((props) => useEventContextData(), {
       seed: {
@@ -41,10 +40,12 @@ describe('useEventContextData', () => {
     });
 
     expect(result.current).toMatchObject({
-      configuration: {
-        title: 'Chapter 1, Section 0'
-      },
-      index: 0,
+      page: {
+        configuration: {
+          title: 'Chapter 1, Section 0'
+        },
+        index: 0
+      }
     });
   });
 });
--- a/entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
@@ -8,7 +8,6 @@ import {renderHookInEntry} from 'support';
 describe('useEventContextData', () => {
 
   it('returns an object with current section and chapter index', () => {
-    
     const {result} = renderHookInEntry((props) => useEventContextData(), {
       seed: {
         entry: {
@@ -37,16 +36,15 @@ describe('useEventContextData', () => {
           },
         ]
       },
-      wrapper: ({children}) => ( <EventContextDataProvider sectionIndex={1}>{children}</EventContextDataProvider> )
+      wrapper: ({children}) =>
+        <EventContextDataProvider sectionIndex={0}>{children}</EventContextDataProvider>
     });
-    
+
     expect(result.current).toMatchObject({
       configuration: {
-        title: null
+        title: 'Chapter 1, Section 0'
       },
-      index: 1,
+      index: 0,
     });
-    
   });
-
 });
--- a/entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
@@ -1,0 +1,52 @@
+import React from 'react';
+
+import {useEventContextData, EventContextDataProvider} from 'frontend/useEventContextData';
+
+import 'support/fakeBrowserFeatures';
+import {renderHookInEntry} from 'support';
+
+describe('useEventContextData', () => {
+
+  it('returns an object with current section and chapter index', () => {
+    
+    const {result} = renderHookInEntry((props) => useEventContextData(), {
+      seed: {
+        entry: {
+          locale: 'en'
+        },
+        sections: [
+          {
+            id: 1,
+            permaId: 101,
+            chapterId: 1,
+            position: 1,
+            configuration: {
+              transition: 'scroll'
+            }
+          },
+        ],
+        chapters: [
+          {
+            id: 1,
+            permaId: 10,
+            position: 1,
+            configuration: {
+              title: 'Chapter 1',
+              summary: 'An introductory chapter'
+            }
+          },
+        ]
+      },
+      wrapper: ({children}) => ( <EventContextDataProvider sectionIndex={1}>{children}</EventContextDataProvider> )
+    });
+    
+    expect(result.current).toMatchObject({
+      configuration: {
+        title: null
+      },
+      index: 1,
+    });
+    
+  });
+
+});
--- a/entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
+++ b/entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
@@ -5,10 +5,9 @@ import {useAtmo} from '../useAtmo';
 import './videojsBase.module.css';
 
 function PlayerContainer({
-  filePermaId, className, sources, textTrackSources, poster, type, playsInline, loop, controls, atmoDuringPlayback, onSetup, onDispose
+  filePermaId, className, sources, textTrackSources, poster, type, playsInline, loop, controls, mediaEventsContext, atmoDuringPlayback, onSetup, onDispose
 }){
   const playerWrapperRef = useRef(null);
-  
   let atmo = useAtmo();
 
   useEffect(() => {
@@ -23,7 +22,8 @@ function PlayerContainer({
         playsInline: playsInline,
         loop: loop,
         controls: controls,
-        hooks: atmoDuringPlayback ? atmo.createMediaPlayerHooks(atmoDuringPlayback) : {} //create hooks only for inline media players
+        hooks: atmoDuringPlayback ? atmo.createMediaPlayerHooks(atmoDuringPlayback) : {}, //create hooks only for inline media players
+        mediaEventsContextData: {page: mediaEventsContext},
       });
       
       let playerElement = player.el();
--- a/entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
+++ b/entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
@@ -5,7 +5,7 @@ import {useAtmo} from '../useAtmo';
 import './videojsBase.module.css';
 
 function PlayerContainer({
-  filePermaId, className, sources, textTrackSources, poster, type, playsInline, loop, controls, mediaEventsContext, atmoDuringPlayback, onSetup, onDispose
+  filePermaId, className, sources, textTrackSources, poster, type, playsInline, loop, controls, mediaEventsContextData, atmoDuringPlayback, onSetup, onDispose
 }){
   const playerWrapperRef = useRef(null);
   let atmo = useAtmo();
@@ -23,9 +23,9 @@ function PlayerContainer({
         loop: loop,
         controls: controls,
         hooks: atmoDuringPlayback ? atmo.createMediaPlayerHooks(atmoDuringPlayback) : {}, //create hooks only for inline media players
-        mediaEventsContextData: {page: mediaEventsContext},
+        mediaEventsContextData
       });
-      
+
       let playerElement = player.el();
       playerWrapper.appendChild(playerElement);
       if (onSetup) {
--- a/entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
+++ b/entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
@@ -74,7 +74,7 @@ function PreparedMediaPlayer(props){
                        loop={props.loop}
                        controls={props.controls}
                        playsInline={props.playsInline}
-                       mediaEventsContext={eventContextData}
+                       mediaEventsContextData={eventContextData}
                        atmoDuringPlayback={props.atmoDuringPlayback}
                        onSetup={onSetup}
                        onDispose={onDispose} />
--- a/entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
+++ b/entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
@@ -7,6 +7,8 @@ import watchPlayer, {unwatchPlayer} from './watchPlayer';
 import {applyPlayerState} from './applyPlayerState';
 import {updatePlayerState} from './updatePlayerState';
 
+import {useEventContextData} from '../useEventContextData';
+
 import {getTextTrackSources, updateTextTracksMode} from './textTracks';
 import textTrackStyles from './textTracks.module.css';
 import styles from '../MediaPlayer.module.css';
@@ -31,6 +33,7 @@ function PreparedMediaPlayer(props){
   let playerRef = useRef();
   let previousPlayerState = useRef(props.playerState);
   let scrollToSection = useContext(ScrollToSectionContext);
+  let eventContextData = useEventContextData();
 
   let onSetup = (newPlayer)=>{
     playerRef.current = newPlayer;
@@ -71,6 +74,7 @@ function PreparedMediaPlayer(props){
                        loop={props.loop}
                        controls={props.controls}
                        playsInline={props.playsInline}
+                       mediaEventsContext={eventContextData}
                        atmoDuringPlayback={props.atmoDuringPlayback}
                        onSetup={onSetup}
                        onDispose={onDispose} />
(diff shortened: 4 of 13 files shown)


entry_types/scrolled/package/src/frontend/MediaPlayer/index.js
```diff
@@ -30,11 +31,13 @@ function PreparedMediaPlayer(props){
   let playerRef = useRef();
   let previousPlayerState = useRef(props.playerState);
   let scrollToSection = useContext(ScrollToSectionContext);
+  let mediaContext = useMediaContext();
 
   let onSetup = (newPlayer)=>{
     playerRef.current = newPlayer;
     newPlayer.on('ended', () => props.nextSectionOnEnd && scrollToSection('next'));
-
+    newPlayer.context = mediaContext;
```
tf, line 39: To get around having to use such a property as a way to pass data to the context function, we could use the same approach we used with `updateHooks`:

`mediaEvents` could define an `updateMediaEventsContextData` method, `getPlayer` could take an `mediaEventsContextData` option and call that method. 

Then the above `MediaPlayer` test could test that the correct context data is passed to `getPlayer`. I think this would be a clearer interface and separation of concerns. We could then also add a test for `mediaEvents` that makes sure media events include the updated context when `updateMediaEventsContextData` is called.

awaismbajwa: I am not sure what should be the structure of event context object.
From our discussion in some previous tickets, we decided that this will be different in scrolled entries.
also there is no page in scrolled so this confuses me.
Can you please suggest the shape of required context object

tf: It's the one I linked in the summary of the review. Scrolled already dispatches `page:change` events to trigger analytics: The object passed there has the desired shape:

https://github.com/codevise/pageflow/blob/9450a8d481e62aa34535794c7b9f5b8065e79edf/entry_types/scrolled/package/src/frontend/useSectionChangeEvents.js#L13-L15

The same sort of object is expected as `context.page` in media events [here](https://gitlab.codevise.de/pageflow/pageflow-analytics/blob/1781c2549270074ddedf37c4e8d77b40e2478b22/app/assets/javascripts/pageflow/analytics/webtrekk/provider.js#L108).

awaismbajwa: So this is what it is I think

`{
  configuration: {
    title: <chapter-title>
  },
  index: <section-index>
  chapterIndex: <>
}`

currently title is going as undefined. let me know if it is required

tf: Yes, since we want to send the same data in `page:change` events as in media event context. And as I said above, it would be nice to extract this logic to construct this dummy page object to a shared function in Scrolled.

awaismbajwa: Currently it is being constructed in useMediaEventsContext. Or you mean something else?

package/src/frontend/VideoPlayer/mediaEvents.js
```diff
@@ -12,19 +12,17 @@ export const mediaEvents = function(player, context) {
...
-    triggerMediaEvent('pause');
-  });
+  let listener = function (event) {
+    triggerMediaEvent(event.type);
+  }
+  
+  player.off('play', listener);
+  player.off('timeupdate', listener);
+  player.off('pause', listener);
+  player.off('ended', listener);
```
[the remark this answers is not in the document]
awaismbajwa: forgot to assign listener to player, should be fine now

entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
```diff
@@ -1,14 +1,15 @@
...
+import {useMediaEventsContext} from '../useMediaEventsContext';
 
 import './videojsBase.module.css';
 
 function PlayerContainer({
   filePermaId, className, sources, textTrackSources, poster, type, playsInline, loop, controls, atmoDuringPlayback, onSetup, onDispose
 }){
   const playerWrapperRef = useRef(null);
-  
+  let mediaEventsContext = useMediaEventsContext();
```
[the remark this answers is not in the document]
awaismbajwa: moved it to MediaPlayer

package/src/frontend/VideoPlayer/mediaEvents.js
```diff
@@ -12,19 +12,21 @@ export const mediaEvents = function(player, context) {
...
+  if (context) {
+    player.mediaEventsListener = function (event) {
+      triggerMediaEvent(event.type);
+    }
+  
+    player.on('play', player.mediaEventsListener);
+    player.on('timeupdate', player.mediaEventsListener);
+    player.on('pause', player.mediaEventsListener);
+    player.on('ended', player.mediaEventsListener);
+  }
```
tf, line 31: This now only registers media events when context is passed. Before context was optional. I don't know if we depend on media events somewhere without passing a context. I think, if we introduce an `updateMediaEvents` method (similar to the already existing `updateHooks` method) these cases will be easier to separate.

Please add a spec for this now more complicated piece of functionality.

awaismbajwa: Changed it to register media events for all players on creation. 
On allocation then media events context is updated.
players without media events context do not trigger media event.

entry_types/scrolled/package/spec/frontend/useMediaEventsContext-spec.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import {useMediaEventsContext, MediaEventsContextDataProvider} from 'frontend/useMediaEventsContext';
+
+import 'support/fakeBrowserFeatures';
+import {renderHookInEntry} from 'support';
+
+describe('useMediaContext', () => {
+
+  it('returns an object with current section and chapter index', () => {
```
aviav, line 10: This isn't what the test tests for. Consider using `renderEntry` and maybe `simulateScrollingIntoView` to really test for *current* section and chapter index

awaismbajwa: [changeSectionEvent-spec](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/frontend/features/changeSectionEvent-spec.js) is also testing the same thing.

entry_types/scrolled/package/spec/frontend/MediaPlayer-spec.js
```diff
@@ -84,6 +84,26 @@ describe('MediaPlayer', () => {
...
+                        type={'video'}
+                        sources={getVideoSources()} />);
+
+    expect(media.getPlayer).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({
+        mediaEventsContextData: {
+          page: {
+            configuration: {
+              title: undefined
```
aviav, line 99: Consider passing a specific `mediaEventsContext` to have a more eloquent test that clarifies the format that `MediaPlayer` expects. This has the advantage that even in the absence of TypeScript etc. other developers can figure out the interface more quickly. Also, this way, you can test for more realistic context data

entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe('useEventContextData', () => {
+
+  it('returns an object with current section and chapter index', () => {
+
+    const {result} = renderHookInEntry(() => useEventContextData());
+    
+    expect(result.current).toMatchObject({
+      configuration: {
+        title: undefined
```
aviav, line 16: Consider setting up the spec in a way that tests a more realistic case. `renderHookInEntry` allows specifying e. g. a wrapper, which you can maybe use to provide realistic data.

Doing this will be more eloquent in a way of describing a situation in which this hook produces useful data

entry_types/scrolled/package/src/frontend/useAtmo.js
```diff
@@ -21,12 +21,10 @@ export function AtmoProvider({children}){
       playerOptions: {tagName: 'audio', loop: true}
     });
     currentAtmo.multiPlayer = new MultiPlayer(currentAtmo.pool, {
-      loop: true,
       fadeDuration: 500,
       crossFade: true,
       playFromBeginning: false,
-      rewindOnChange: true,
-      pauseInBackground: true
```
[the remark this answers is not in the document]
awaismbajwa: yes, it was a bug in master. I had discussed it with Tim. Did not wanted to create separate merge request and ticket for this change. Change is mentioned in the commit message.

package/src/frontend/media/createMediaPlayer.js
```diff
@@ -25,7 +24,7 @@ export const createMediaPlayer = function (options) {
     hooks: {},
 
     mediaEvents: true,
-    context: options.mediaContext
+    context: undefined
```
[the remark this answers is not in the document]
awaismbajwa: One thing I can do is don't specify context at all, in that case it would be difficult to understand.
but here it is clear that initially context is undefined.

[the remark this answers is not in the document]
awaismbajwa: also this issue is debatable and depends more on the development standards that we have according to me.
https://stackoverflow.com/a/33397680/6056825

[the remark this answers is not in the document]
aviav: As stated in the link in the first comment, for the purpose that you intend, JS has `null`

package/src/frontend/media/MediaPool.js
```diff
@@ -52,6 +53,7 @@ export class MediaPool {
         player.playsinline(true); 
       }
       player.updateHooks(hooks || {});
+      player.updateMediaEventsContext(mediaEventsContextData);
```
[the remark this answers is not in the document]
awaismbajwa: it is mentioned in the commit message already.
* player.updateMediaEventsContext is used to update media context for player on allocation.

entry_types/scrolled/package/src/frontend/useEventContextData.js
```diff
@@ -0,0 +1,39 @@
+import React, {createContext, useContext, useMemo} from 'react';
+import {useEntryStructure} from '../entryState';
+
+export function getEventObject(sectionIndex, chapterIndex, entryStructure){
```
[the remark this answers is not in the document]
awaismbajwa: I think first we should identify what is required. 
If chapterIndex is required then getEventObject interface 
is not more important than requirement may be still it can be adjusted.
if chapterIndex is not required than it can simply be removed from getEventObject .

package/src/frontend/media/createMediaPlayer.js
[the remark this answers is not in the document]
aviav: I take your point from the second comment. Probably, for an issue like this, we could find several Stack Overflow posts supporting either practice.

In support of what I wrote previously, I'll reference MDN:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null
https://developer.mozilla.org/en-US/docs/Glossary/Undefined

Using the `undefined` value in assignments against its intended usage in JS can be confusing to developers who assume we are using JS as intended, and don't assign `undefined` explicitly. Wherever we assign `undefined` explicitly, we lose this context, which can make debugging more difficult in practice.

I don't see a good reason to deviate from the standard here, but if you insist on using `undefined`, for the sake of short-term productivity, I'll merge this regardless

entry_types/scrolled/package/src/frontend/useAtmo.js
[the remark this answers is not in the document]
aviav: Sorry, I misinterpreted this change. You're right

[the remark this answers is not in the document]
aviav: Maybe, in case you're reworking the commit message, you could provide the reason for this change there. Otherwise, for somebody who reads the commit history of this file, it could look like a case of 'first we did that, then we did that', but the reason would remain unclear.

Still, in cases like this, we often don't provide the reasoning. It would be an improvement to do so, but I don't consider it necessary for merge

package/src/frontend/media/createMediaPlayer.js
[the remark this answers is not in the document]
awaismbajwa: I agree with your arguments. I will make change accordingly

package/src/frontend/media/MediaPool.js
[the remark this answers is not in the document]
aviav: For me, this line from the commit message re-iterates what is already obvious from the code. Still, I see that it is written in a way that it could be seen to answer some possible 'why' questions.

To be more clear, the specific question that is open for me would be: Why do we need to update a media context, when the context of media in an entry is constant?

entry_types/scrolled/package/src/frontend/useEventContextData.js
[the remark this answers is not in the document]
aviav: I'd say there is more to be gained by just always providing an event object with the same form, because there isn't a domain-based or pragmatic reason to omit chapter index from the page object in any case.

If I imagine a future developer who would like to do something with the page object, they might start cursing when they notice that it has the chapter index only in the case of media events.

I agree with you that this version of `getEventObject` does its job in terms of requirements, but I think we could easily do better both in terms of simplifying the function signature and in terms of providing consistent output.  

The latter is important especially because we don't have TypeScript or similar, and because usage of the `page` object happens in other modules. It could be really time-consuming to have to reconstruct the arcane reason why chapter index is only sometimes included in the page object

package/src/frontend/media/MediaPool.js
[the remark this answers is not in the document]
awaismbajwa: actually this approach was getting very common in our new setup. that is why I did not consider it that important to reiterate it here again.
but I can write some more words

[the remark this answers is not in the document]
aviav: If you think it isn't something that should be in the commit message, you can also write a few quick words here, and in case I still don't understand, we can discuss it tomorrow

entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js
awaismbajwa: I have tried encapsulating it but it is giving me error. There are some changes in renderHookInEntry like it is not taking wrapper as argument.
Even if I change it to take wrapper as argument it is giving me error.

```const {result} = renderHookInEntry((props) => useEventContextData(), {
      seed: {
        sections: [
          {
            id: 1,
            permaId: 101,
            chapterId: 1,
            position: 1,
            configuration: {
              transition: 'scroll'
            }
          },
        ],
        chapters: [
          {
            id: 1,
            permaId: 10,
            position: 1,
            configuration: {
              title: 'Chapter 1',
              summary: 'An introductory chapter'
            }
          },
        ]
      },
      wrapper: () => ( <EventContextDataProvider sectionIndex={1}/> )
    });```


do you have any idea how can I encapsulate it properly?

Comment (aviav) [reply entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js:16]: Indeed, `renderHookInEntry` doesn't accept `wrapper` yet. I think your approach of adding it is the way to go.

We have an example of something similar in scrollPositionLifecycle [1]. Have you tried using a call to a creator function instead of a function literal as the `wrapper` parameter, as we do it there?

[1] https://github.com/codevise/pageflow/blob/f4f4f9523df802c82990bb2ce96fcee5936462fb/entry_types/scrolled/package/spec/support/scrollPositionLifecycle.js#L38

Comment (awaismbajwa) [reply entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js:16]: Here in this example it is creating custom provider within Test environment.
If I do that that would mean providing hardcoded values like sectionIndex and title and then again testing those
which would become similar to what it is now.

Comment (aviav) [reply entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js:16]: It would be different in a way that is crucial for a good test: It would give an example for usage of the interface that will lead to non-default output, and thus explain as well as test this piece of functionality.

Phrased differently, the difference would be: Currently, the spec tests whether there is a hook by the given name that returns data in a specific format regardless of context. After the change we discuss, the test will include a possible valid context, and will test what the hook does in that context. In terms of test quality, I wouldn't call this 'similar to what it is now'

Comment (awaismbajwa) [reply entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js:16]: `      wrapper: () => ( <EventContextDataProvider sectionIndex={1}/> )`

it is doing same thing. only difference is that it uses actual Provider. 
The error I am getting is related to entryState. that entryState is not defined

Comment (awaismbajwa) [reply entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js:16]: I think it is more important to test Provider rather than context

Comment (aviav) [reply entry_types/scrolled/package/spec/frontend/useEventContextData-spec.js:16]: > ` wrapper: () => ( <EventContextDataProvider sectionIndex={1}/> )`
> 
> it is doing same thing. only difference is that it uses actual Provider.
> The error I am getting is related to entryState. that entryState is not defined

In the example from scrollPositionLifecycle, we don't pass a function literal that returns a wrapper and still needs to be called. Rather, the difference there is that we call a function named `createScrollPositionProvider`, and that function call returns a component `ScrollPositionProvider`. So maybe we could try something along the lines of

```javascript
wrapper: createEventContextDataProvider({sectionIndex: 1, chapterIndex: 2})
```
and
```javascript
function createEventContextDataProvider({sectionIndex, chapterIndex}) {
  return (
    <EventContextDataProvider sectionIndex={sectionIndex} chapterIndex={chapterIndex}/>
  );
}
```
which would mean we also pass a component as `wrapper` and not a function that returns a component when called

awaismbajwa: I am afraid this is not helpful.
wrapper parameter requires a function (which returns react component).
In the example above that you shared it is assigning react component directly to wrapper.

Wrapping it in the function is syntactic sugar. What we actually want is a function which returns react component.
I have tried it by wrapping it inside a function or without it. In both cases I get entryState undefined error.

I suspect I am getting this error because somehow hook is not getting wrapped inside provided wrapper.
That is why it is giving undefined entryState.

awaismbajwa: do you know if there is any place where I can get reference to the seed data structure?

aviav: > I am afraid this is not helpful.
> wrapper parameter requires a function (which returns react component).
> In the example above that you shared it is assigning react component directly to wrapper.
> 
> Wrapping it in the function is syntactic sugar. What we actually want is a function which returns react component.
> I have tried it by wrapping it inside a function or without it. In both cases I get entryState undefined error.
> 
> I suspect I am getting this error because somehow hook is not getting wrapped inside provided wrapper.
> That is why it is giving undefined entryState.

It shouldn't make a difference, but it is something I still would've tried. Also I agree that if it doesn't work, it isn't helpful. Probably it doesn't make much sense if I try to provide you input for debugging via GitHub comments

awaismbajwa: I am not looking for input on debugging or programming solutions. I need to get clear understanding like e.g. I don't know what should the seed data or inputs to renderHookInEntry look like? 
I tried finding answer to this question in source code but I could not find it.
If I know how renderHookInEntry can be used and its seed data structure then I will be able to solve this issue easily.
Otherwise I am not sure how much I need to look into source code. I am trying to figure out, I am not even sure if what I am looking for is present in source code or not, do you see my problem?

aviav: > do you know if there is any place where I can get reference to the seed data structure?

You can find it in the documentation of `normalizeSeed` [1], which we use indirectly in `renderHookInEntry`

[1] https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/support/normalizeSeed.js

aviav: > If I know how renderHookInEntry can be used

The way I'd do this would be to look at existing usages of `renderHookInEntry`

awaismbajwa: So the question to which I am trying to find answer for is:

What parameters I can pass to renderHookInEntry, so that my hook useEventContextData get wrapped inside 
the entry?

As it is mentioned in normalizeSeed.js I have tried passing 
```
{
  seed: {
    entry: {}
    sections: [...],
    chapters: [...]
  }
}
```

but still it is giving me error that entryState is undefined

awaismbajwa: > > If I know how renderHookInEntry can be used
> 
> The way I'd do this would be to look at existing usages of `renderHookInEntry`

That was the first thing I did but as I told you renderHookInEntry does not even take wrapper so this is the first use case

aviav: If you push a branch with your failing spec to your GitHub account, I could have a look

awaismbajwa: sure, please have a look now

aviav: > > The way I'd do this would be to look at existing usages of `renderHookInEntry`
> 
> That was the first thing I did but as I told you renderHookInEntry does not even take wrapper so this is the first use case

In this case, you could look for usages of `renderInEntry` that receive a wrapper. `renderInEntry` and `renderHookInEntry` are very similar.

Comparing the documentations of `render` [1] and `renderHook` [2], it seems to me that both use a `wrapper` parameter in the same way

[1] https://testing-library.com/docs/react-testing-library/api#wrapper
[2] https://react-hooks-testing-library.com/reference/api#wrapper

awaismbajwa: > > > The way I'd do this would be to look at existing usages of `renderHookInEntry`
> > 
> > 
> > That was the first thing I did but as I told you renderHookInEntry does not even take wrapper so this is the first use case
> 
> In this case, you could look for usages of `renderInEntry` that receive a wrapper. `renderInEntry` and `renderHookInEntry` are very similar.
> 
> Comparing the documentations of `render` [1] and `renderHook` [2], it seems to me that both use a `wrapper` parameter in the same way
> 
> [1] https://testing-library.com/docs/react-testing-library/api#wrapper
> [2] https://react-hooks-testing-library.com/reference/api#wrapper

yes I did that already and that is how I came up with the idea of using wrapper in renderHookInEntry.

aviav: As you requested in our discussion just now, we'll leave this PR until Tim returns. In that case, no need for me to look into that failing spec, I think

PR #1521, opened 2020-07-13
Media in background

awaismbajwa:
REDMINE-17771

### DIFF
--- a/entry_types/scrolled/package/src/frontend/Backdrop.js
+++ b/entry_types/scrolled/package/src/frontend/Backdrop.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import classNames from 'classnames';
 
 import Fullscreen from './Fullscreen';
@@ -10,6 +10,7 @@ import useDimension from './useDimension';
 import {usePlayerState} from './MediaPlayer/usePlayerState';
 import {usePortraitOrientation} from './usePortraitOrientation';
 import {useSectionLifecycle} from './useSectionLifecycle';
+import {documentHiddenState} from 'pageflow/frontend';
 
 import styles from './Backdrop.module.css';
 
@@ -129,6 +130,20 @@ function BackgroundVideo(props) {
       playerActions.pause()
     }
   });
+  
+  useEffect(() => {
+    let documentState = documentHiddenState((visibilityState) => {
+      if (visibilityState === 'hidden') {
+        playerActions.pause();
+      }
+      else{
+        playerActions.play();
+      }
+    });
+    return () => {
+      documentState.removeCallback();
+    }
+  }, [playerActions])
 
   return (
     <VideoPlayer isPrepared={isPrepared}
--- a/entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
+++ b/entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
@@ -1,6 +1,6 @@
 import React, {useEffect, useRef} from 'react';
 import {media} from 'pageflow/frontend';
-import { useAtmo } from '../useAtmo';
+import {useAtmo} from '../useAtmo';
 
 import './videojsBase.module.css';
 
@@ -23,7 +23,7 @@ function PlayerContainer({
         playsInline: playsInline,
         loop: loop,
         controls: controls,
-        hooks: atmo.createMediaPlayerHooks(atmoDuringPlayback)
+        hooks: atmoDuringPlayback ? atmo.createMediaPlayerHooks(atmoDuringPlayback) : {} //create hooks only for inline media players
       });
       
       let playerElement = player.el();
--- a/package/src/frontend/documentHiddenState.js
+++ b/package/src/frontend/documentHiddenState.js
@@ -1,0 +1,24 @@
+
+let isEventAdded = false;
+let callbacks = [];
+
+let muteInBackground = () => {  
+  callbacks.forEach((cb)=>{    
+    cb(document.visibilityState);
+  })
+};
+
+export function documentHiddenState(callback){
+  callbacks.push(callback);
+
+  if (!isEventAdded) {
+    isEventAdded = true;
+    document.addEventListener('visibilitychange', muteInBackground, false);
+  }
+
+  return {
+    removeCallback: () => {
+      callbacks = callbacks.filter(c => c !== callback);
+    }
+  }
+}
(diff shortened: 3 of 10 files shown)


package/src/frontend/media/media.js
```diff
@@ -33,3 +44,4 @@ export const media = {
 };
 
 Object.assign(media, BackboneEvents);
+document.addEventListener('visibilitychange', media.muteInBackground, false);
```
tf, line 47: This will break server side rendering and also run in Paged where it is not needed. Maybe we can set this up when media is first unmuted? Since `muteInBackground` is never supposed to be called by outside code, I'd test it by dispatching `visibilitychange` events on document.

awaismbajwa: Do you know how can I dispatch event to hide page?
i have tried this 
`  window.dispatchEvent(new Event('pagehide'));`
but it does not work

tf: You listen on `document`, so I'd guess you need to use `document.dispatchEvent`. Also the event name is different.

awaismbajwa: I had tried it with document also. what would be the event name?
tried this also `document.dispatchEvent(new Event('visibilitychange'));`

awaismbajwa: Initially I tried testing through document visibility change but I could not find any solution for that.
after that I decided to test it manually

tf: This passes on my machine:
```js
const listener = jest.fn();

document.addEventListener('visibilitychange', listener);
document.dispatchEvent(new Event('visibilitychange'));
      
expect(listener).toHaveBeenCalled();
```

awaismbajwa: Listener gets called for me as well. I am testing muteInBackground function and visibilityState property of document does not change through this event.

tf: see https://github.com/jsdom/jsdom/issues/2190

awaismbajwa: defineProperty is already being used. I will use both approaches to makes the tests complete

awaismbajwa:
A new approach is introduced in the last commit. If it is fine then I will remove older commits

entry_types/scrolled/package/src/frontend/Backdrop.js
```diff
@@ -129,6 +130,21 @@ function BackgroundVideo(props) {
...
+        playerActions.pause();
+      }
+      else{
+        playerActions.play();
+      }
+    }
+    documentHiddenState(callback); //register callback
+    return () => {
+      documentHiddenState(callback, true); //remove callback
+    }
```
tf, line 146: If `documentHiddenState` returned a function that removes the event listeners, it would be a better fit for `useEffect`. We would also not need the comments, that currently work around the fact that the `true` parameter does not convey meaning.

When a background video is playing and atmo played at least once before, switching away from the tab resumes the atmo. This  happens since all media players attach atmo hooks - even those for background videos. Switching the tab now pauses the video and calls the after hook which resumes the atmo.

awaismbajwa: so background videos should not have atmo hooks right?

awaismbajwa: In the recent changes I have adjusted the documentHiddenState as you suggested.
Now only assigning hooks to InlineVideo and InlineAudio elements.
BackgroundVideos player should not be getting hooks assigned.
Since Inline players are not getting paused on visibilitychange, this problem should be resolved now?

package/src/frontend/documentHiddenState.js
```diff
@@ -0,0 +1,27 @@
+
+let isEventAdded = false;
+let callbacks = [];
+
+let muteInBackground = () => {  
+  callbacks.forEach((cb)=>{    
+    if (cb instanceof Function) {
```
aviav, line 7: If we take this check out, developers could get more direct feedback if they register a non-function callback by mistake. Since we control the type of cb, I'd suggest taking out the `instanceOf` check here

package/src/frontend/documentHiddenState.js
```diff
@@ -0,0 +1,27 @@
...
+export function documentHiddenState(callback){
+  callbacks.push(callback);
+
+  if (!isEventAdded) {
+    isEventAdded = true;
+    document.addEventListener('visibilitychange', muteInBackground, false);
+  }
+
+  return {
+    visibilityState: document.visibilityState,
```
aviav, line 22: So far, we don't use the returned `visibilityState` anywhere. Since `document.visibilityState` is globally available, users can always use that one. Since values between returned `visibilityState` and `document.visibilityState` can diverge, let's not return `document.visibilityState`'s current value here

entry_types/scrolled/package/src/frontend/Backdrop.js
awaismbajwa: Should we stop Atmo when BackgroundVideo starts playing?
I realised it today that BackgroundVideo has sound as well

entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
```diff
@@ -23,7 +23,7 @@ function PlayerContainer({
         playsInline: playsInline,
         loop: loop,
         controls: controls,
-        hooks: atmo.createMediaPlayerHooks(atmoDuringPlayback)
+        hooks: atmoDuringPlayback ? atmo.createMediaPlayerHooks(atmoDuringPlayback) : {}
```
aviav, line 26: As discussed, depending on user expectation, this could be right or wrong. Let's find out what happens in case of simultaneous usage of background video and atmo audio in the case where the video has sound. It might also be interesting to see whether behaviors differ between paged and scrolled, because in this case it might be an oversight in scrolled

package/src/frontend/media/MultiPlayer.js
```diff
@@ -47,6 +47,15 @@ export const MultiPlayer = function(pool, options) {
     return current.fadeOutAndPause(options.fadeDuration);
   };
 
+  this.fadeOutIfPlaying = function () {
+    if (current.paused()) {
+      return Promise.resolve();
+    }
+    else{
+      current.fadeOutAndPause(options.fadeDuration);
```
aviav, line 55: `return` is missing

entry_types/scrolled/package/src/frontend/Backdrop.js
aviav: I would do it the same way we currently do it in paged entries, because there's no reason to do anything different here

entry_types/scrolled/package/src/frontend/MediaPlayer/PlayerContainer.js
awaismbajwa: In pageflow-paged background video and atmo play simultaneously.
And here in scrolled as well.

PR #1528, opened 2020-07-16
Make entry templates editable

aviav:
Re-opening #1420 to see whether we can un-confuse Hound that way

REDMINE-17306


aviav:
Didn't help. Hound still finds 159 violations

PR #1533, opened 2020-07-17
Responsive Datawrapper chart height

UsmanMuhammad:
REDMINE-17670

DataWrapperChart height was fixed previously to a
static value

It receives a post message from DataWrapper with the
ideal height according to device

It creates a hook, which register event and listen to
the messages from iframe and ajust the height accordingly

### DIFF
--- a/entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
+++ b/entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
@@ -1,0 +1,44 @@
+import {useIframeHeight} from 'contentElements/dataWrapperChart/useIframeHeight';
+
+import {renderHook} from '@testing-library/react-hooks';
+import {fakeParentWindow, tick} from 'support';
+
+describe('useIframeHeight', () => {
+  it('Confirms the default height', async () => {
+    const testURL = 'testUrl/id/1/';
+    fakeParentWindow();
+
+    const {result} = renderHook(() => useIframeHeight(testURL));
+
+    window.postMessage('SOME_MESSAGE', '*');
+    await tick();
+
+    expect(result.current).toEqual('400px');
+  });
+
+  it('removes listener on cleanup', async () => {
+    const testURL = 'testUrl/id/1/';
+    fakeParentWindow();
+    const {unmount} = renderHook(() => useIframeHeight(testURL));
+    const spy = jest.spyOn(window.document, 'removeEventListener');
+
+    unmount();
+    window.postMessage('SOME_MESSAGE', '*');
+    await tick();
+
+    expect(spy).toBeCalledWith('message', expect.any(Function));
+  });
+
+  it('sets the given height', async () => {
+    const testURL = 'testUrl/id/1/';
+    fakeParentWindow();
+    const {result, rerender} = renderHook(() => useIframeHeight(testURL));
+    expect(result.current).toEqual('400px');
+
+    window.postMessage({'datawrapper-height': {id : 350}}, '*');
+    await tick();
+    rerender();
+
+    expect(result.current).toEqual('350px');
+  });
+});
--- a/entry_types/scrolled/package/src/contentElements/dataWrapperChart/DataWrapperChart.module.css
+++ b/entry_types/scrolled/package/src/contentElements/dataWrapperChart/DataWrapperChart.module.css
@@ -1,6 +1,5 @@
 .container {
   min-height: 200px;
-  /* height: 400px; */
   background-color: rgba(0, 0, 0, 0.5);
 }
 
--- a/entry_types/scrolled/package/src/contentElements/dataWrapperChart/DataWrapperChart.module.css
+++ b/entry_types/scrolled/package/src/contentElements/dataWrapperChart/DataWrapperChart.module.css
@@ -1,6 +1,6 @@
 .container {
   min-height: 200px;
-  height: 400px;
+  /* height: 400px; */
   background-color: rgba(0, 0, 0, 0.5);
 }
 
--- a/entry_types/scrolled/package/src/frontend/useIframeHeight.js
+++ b/entry_types/scrolled/package/src/frontend/useIframeHeight.js
@@ -1,0 +1,22 @@
+import {useState, useEffect} from 'react'
+
+export function useIframeHeight(url) {
+  const [height, setHeight] = useState('400px');
+
+  useEffect(() => {
+    window.addEventListener('message', receive)
+
+    function receive(event) {
+      if (typeof event.data['datawrapper-height'] !== 'undefined') {
+        for (var chartId in event.data['datawrapper-height']) {
+          if (url.indexOf(chartId) > -1) {
+            setHeight(event.data['datawrapper-height'][chartId] + 'px');
+          }  
+        }
+      }
+    }
+    return () => document.removeEventListener('message', receive);
+  }, [url]);
+
+  return height
+}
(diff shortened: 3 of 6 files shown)


entry_types/scrolled/package/src/contentElements/dataWrapperChart/DataWrapperChart.module.css
```diff
@@ -1,6 +1,6 @@
 .container {
   min-height: 200px;
-  height: 400px;
+  /* height: 400px; */
```
tf, line 3: Can be removed.

entry_types/scrolled/package/src/frontend/useIframeHeight.js
```diff
@@ -0,0 +1,22 @@
...
+            setHeight(event.data['datawrapper-height'][chartId] + 'px');
+          }  
+        }
+      }
+    }
+    return () => document.removeEventListener('message', receive);
+  }, [url]);
+
+  return height
+}
```
tf, line 22: Let's move this hook to `src/contentElements/dataWrapperChart` since it's so specific to this use case. 

Let's also add a test for the hook. It can live in `spec/contentElements/dataWrapperChart` (the `spec/contentElements` directory does not yet exist since the other content elements do not have tests. They mostly assemble functionality from `src/frontend` that is already tested there. But in this case we have some content element specific logic in this hook that can be tested individually). You can follow a similar approach like the [spec of `usePostMessageListener`](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/frontend/usePostMessageListener-spec.js#L7). (We could have also considered reusing that hook here, but it currently only registers listeners when the page is rendered in an iframe itself to receive messages from the parent. So I'd say we keep our new hook exactly the way it is) Note that [jsdom dispatches messages asynchronously](https://github.com/jsdom/jsdom/blob/574eb51a29c17c01e6529dda4cdbdb96222d01c2/lib/jsdom/living/post-message.js#L35). This is why the test uses the `tick` helper to wait until the message has been processed.

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
```diff
@@ -0,0 +1,19 @@
+import {useIframeHeight} from '../../../src/contentElements/dataWrapperChart/useIframeHeight'
```
aviav, line 1: Consider using an absolute path here, as we also do in [1]. This would decouple the spec file's content from its path

[1] https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/frontend/usePostMessageListener-spec.js#L1

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+import {renderHook} from '@testing-library/react-hooks';
+import {fakeParentWindow, tick} from 'support';
+import {act} from '@testing-library/react';
+
+describe('useIframeHeight', () => {
```
aviav, line 7: Consider adding a spec that posts a message with the expected format, but leads to a width that is different from 400px, using `postMessage`. I see two advantages:

- spec tests more of the hook's functionality than it currently does
- spec contains an example of how we'd expect message payload to look like, which makes it easier for us to adapt in case Datawrapper's payload changes

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
```diff
@@ -0,0 +1,19 @@
...
+    const testURL = 'https://datawrapper.dwcdn.net/CXXQo/1/';
+    fakeParentWindow();
+
+    const {result} = renderHook(() => useIframeHeight(testURL));
+
+    window.postMessage('SOME_MESSAGE', '*');
+    await tick();
+       
+    expect(result.current).toEqual('400px');
+  });
```
aviav, line 18: Consider adding a spec that tests whether we can successfully remove the event listener. This has the advantage that in case somebody breaks this functionality, there will be a failing spec and we'd notice before the merge. For an example, see [1]

[1] https://github.com/codevise/pageflow/blob/f4f4f9523df802c82990bb2ce96fcee5936462fb/entry_types/scrolled/package/spec/frontend/usePostMessageListener-spec.js#L18

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
```diff
@@ -0,0 +1,19 @@
+import {useIframeHeight} from '../../../src/contentElements/dataWrapperChart/useIframeHeight'
+
+import {renderHook} from '@testing-library/react-hooks';
+import {fakeParentWindow, tick} from 'support';
+import {act} from '@testing-library/react';
```
aviav, line 5: This is unused

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+describe('useIframeHeight', () => {
+  it('sets the default height', async () => {
+    const testURL = 'https://datawrapper.dwcdn.net/CXXQo/1/';
```
aviav, line 9: While it is nice to have a sample URL here to remind us of the expected format, it might be safer to use a mock instead of a real graph. Maybe something like `https://datawrapper.dwcdn.net/aBcDe/1/`?

This would also protect us against hypothetical side effects that occur just because we provide the URL of a specific graph, i. e. a spec that passes only by accident

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
UsmanMuhammad: I changed the functionality of the hook to accommodate the tests. Now we can check that if the value changes then the listener is called with the new value and changes the state.

aviav: I wouldn't change the hook for that. The previous interface was leaner and better. Consider spying on `window.document`'s `removeEventListener` in the spec and reverting the changes to the hook

aviav: Now, the spec that posted the `({'datawrapper-height': {id : 350}}` message seems to have been reverted as well. That spec was fine. Could you re-add that?

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
```diff
@@ -0,0 +1,32 @@
...
+  it('removes listener on cleanup', async () => {
+    const testURL = 'testUrl/id/1/';
+    fakeParentWindow();
+    const {unmount} = renderHook(() => useIframeHeight(testURL));
+
+    unmount();
+    window.postMessage('SOME_MESSAGE', '*');
+    await tick();
+
+    const spy = jest.spyOn(window, 'removeEventListener');
```
aviav, line 28: We'd expect `removeEventListener` on `window.document` to be called with parameters 'message' and any `Function`. I think this spec now tests something else

entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js
aviav: Very nice now! Also thanks to @awaismbajwa for taking a look!

PR #1539, opened 2020-07-24
Add meta tags in scrolled

UsmanMuhammad:
REDMINE-17778

Social sharing meta tags and meta tags
are rendered by paged in entry head. Same
can be used for scrolled

It adds the same helpers used for paged in scrolled
entry head

It also changes the lang attribute and reflect the entry
locale as in paged

### DIFF
--- a/entry_types/scrolled/app/controllers/pageflow_scrolled/entries_controller.rb
+++ b/entry_types/scrolled/app/controllers/pageflow_scrolled/entries_controller.rb
@@ -4,10 +4,15 @@ module PageflowScrolled
     include Pageflow::EntriesControllerEnvHelper
 
     helper Pageflow::WidgetsHelper
+    helper Pageflow::PublicI18nHelper
+    helper Pageflow::SocialShareHelper
+    helper Pageflow::MetaTagsHelper
 
     def show
       @entry = get_published_entry_from_env
       @widget_scope = get_entry_mode_from_env
+
+      I18n.locale = @entry.locale
     end
   end
 end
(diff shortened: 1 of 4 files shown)


entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<%= content_tag(:html, lang: I18n.locale) do %>
```
tf, line 2: **suggestion:** We can also port the [corresponding test from paged](https://github.com/codevise/pageflow/blob/f51a288a5abe2df495ee644e80a64ff8b345a0b6/entry_types/paged/spec/controllers/pageflow_paged/entries_controller_spec.rb#L169).

entry_types/scrolled/app/views/pageflow_scrolled/entries/show.html.erb
```diff
@@ -39,4 +42,4 @@
 
     <%= scrolled_entry_json_seed_script_tag(@entry, @seed_options || {}) %>
   </body>
-</html>
+  <% end %>
```
tf, line 45: **nitpick:** Indentation is off.

entry_types/scrolled/app/controllers/pageflow_scrolled/entries_controller.rb
```diff
@@ -4,10 +4,15 @@ class EntriesController < ActionController::Base
     include Pageflow::EntriesControllerEnvHelper
 
     helper Pageflow::WidgetsHelper
+    helper Pageflow::PublicI18nHelper
```
tf, line 7: **question**: Why is this needed?

PR #1540, opened 2020-07-28
Fix storybook documentation

aviav:
REDMINE-17670

### DIFF
--- a/entry_types/scrolled/doc/creating_content_elements.md
+++ b/entry_types/scrolled/doc/creating_content_elements.md
@@ -158,6 +158,12 @@ Run the following command in the root directory of your host application
 $ bundle exec rake pageflow_scrolled:storybook:seed:setup[./seed.json]
 ```
 
+Note that brackets need to be escaped in `zsh`, so that the latter becomes
+
+```bash
+$ bundle exec rake pageflow_scrolled:storybook:seed:setup\[./seed.json\]
+```
+
 Then move the generated `seed.json` file into the
 `entry_types/scrolled/package/.storybook/` directory of your
 development checkout of the `pageflow` project.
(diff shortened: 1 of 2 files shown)


entry_types/scrolled/doc/creating_content_elements.md
```diff
@@ -155,7 +155,14 @@ Run the following command in the root directory of your host application
 [documentation below](#using-transcoded-files-in-storybook-or-percy)):
 
 ```bash
-$ bundle exec rake pageflow_scrolled:storybook:seed:setup[./seed.json]
+$ bundle exec rake pageflow_scrolled:storybook:seed:create_entry
+$ bundle exec rake pageflow_scrolled:storybook:seed:generate_json[./seed.json]
```
tf, line 159: Why can't we call `setup` here any longer? It should invoke the other two:

https://github.com/codevise/pageflow/blob/f51a288a5abe2df495ee644e80a64ff8b345a0b6/entry_types/scrolled/lib/tasks/pageflow_scrolled_tasks.rake#L9-L10

aviav: I concluded that, since `bundle exec rake -T` from Dedicated only lists `create_entry`, `destroy_entry`, and `generate_json[output]`, `setup` wouldn't work anymore

aviav: I just found out about `rake -T -A`, which shows `setup`, as well.

What do you think about the note on `zsh`? `zsh` has become more and more common, so maybe it makes sense to mention this kind of thing

tf: The setup task does not show up since it is missing a description. So maybe we should add that instead. I do not use zsh myself, but I'm totally fine with including the hint.

PR #1541, opened 2020-07-28
Waveform Player Controls for AudioPlayer Component

awaismbajwa:
REDMINE-17677


entry_types/scrolled/package/spec/frontend/AudioPlayerControls-spec.js
```diff
@@ -34,6 +34,23 @@ describe('AudioPlayerControls', () => {
...
+      seed: {
+        audioFiles: [{
+          id: 100,
+          permaId: 10
+        }]
+      }
+    });
+    expect(getByTestId('waveform-controls')).not.toBeNull();
+  });
+
```
tf, line 53: **praise**: Good call testing the feature on `AudioPlayerControls` since that's the place where it is actually used.

entry_types/scrolled/package/src/frontend/PlayerControls/WaveformPlayerControls/Waveform.js
```diff
@@ -0,0 +1,57 @@
...
+  let measureRect = useRef();
+
+  if (props.mediaElementId) {
+    mediaElementId = `#${props.mediaElementId}`;
+    let mediaElement = global.document.querySelector(mediaElementId);
+    if (mediaElement) {
+      if(mediaElement.src !== blankSources['audio'])
+        isReady = true;
+    }
+  }
```
tf, line 26: **issue**: Exception when adding or removing caption

When adding a caption, the component gets rerendered since it is now wrapped in a `Figure` component. During this rerender I observe a case where the media element is still in the document while this render function, but by the time the new `Waversurfer` component is mounted the media element has been removed. This causes a "Media element not found" exception in `_loadMediaElt` below.

The root of the problem, is using DOM APIs in a render function here. React render function should be pure. Since they potentially run a lot of time, doing more expensive stuff like looking up DOM nodes also risks causing performance issues.

I think, I have found a different solution. Which I will push to this branch along with the SSR fix. Maybe you can test again if it also covers all cases that you found during development

tf: I pushed two commits to fix SSR issue and issue mentioned here.

entry_types/scrolled/package/src/frontend/MediaPlayerControls.js
```diff
@@ -16,36 +16,61 @@ export function MediaPlayerControls(props) {
...
+                      textTracksMenuItems={getTextTracksMenuItems(textTracks, t)}
+                      onTextTracksMenuItemClick={textTracks.select}
 
-                    qualityMenuItems={props.qualityMenuItems}
-                    onQualityMenuItemClick={props.onQualityMenuItemClick} />
-  )
+                      qualityMenuItems={props.qualityMenuItems}
+                      onQualityMenuItemClick={props.onQualityMenuItemClick} />
+    )
+  }
```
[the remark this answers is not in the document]
tf: There was a misunderstanding. I did not mean to suggest moving the `playerState`/`playerActions` props from `MediaPlayerControls` to `PlayerControls` since that:

* Does not fix the mentioned duplication - it has just been moved

* Does not allow simplifying the duplicated specs

* Does not fix [the storybook](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/src/frontend/__stories__/playerControls-stories.js#L28) which still expects to be able to use low level props like `duration` on `PlayerControls`.

* Moreover, the divide of responsibilities between `PlayerControls` and `MediaPlayerControls` is now gone: Before `PlayerControls` was about the low level mechanics and look of the player controls, `MediaPlayerControls` was about binding the player controls to player state.

What I meant was:

* Basically keep the props of `PlayerControls` the way they are on `master` and only add a `variant` prop.

* Align the prop names of `WaveformPlayerControls` with `ClassicPlayerControls` (e.g. `onPlayerButtonClick` vs `play`).

* Based on `variant` render either `WaveformPlayerControls` or `ClassicPlayerControls` in `PlayerControls` passing exactly the same props. This removes the need to have two long, subtly different prop lists. Not all props are used by both components, but that does not matter.

* `MediaPlayerControls` can then stay basically the way it is on `master` and set the `variant` based on `configuration.playerControlsVariant`.d

* The specs which are now duplicated for `WaveformPlayerControls` and `ClassicPlayerControls` can be put into a loop that iterates over the two possible variants and render the `PlayerControls` component passing the `variant`.

* This fixes the storybook since the interface of `PlayerControls` was only extended.

entry_types/scrolled/package/src/frontend/PlayerControls/WaveformPlayerControls/Waveform.js
```diff
@@ -0,0 +1,57 @@
...
+               }}>
+        { ({measureRef}) =>
+          <div ref={measureRef} className={styles.waveWrapper}>
+            <Wavesurfer mediaElt={mediaElementId}
+                        playing={props.isPlaying}
+                        options={{
+                          normalize: true,
+                          removeMediaElementOnDestroy: false,
+                          hideScrollbar: true,
+                          progressColor: props.waveformColor || props.mainColor,
```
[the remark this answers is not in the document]
tf: The input field still shows transparent color as default when creating a new waveform audio.

entry_types/scrolled/package/src/frontend/PlayerControls/WaveformPlayerControls/Waveform.js
```diff
@@ -22,6 +22,8 @@ export function Waveform(props) {
             <div ref={measureRef} className={styles.waveWrapper}>
               <Wavesurfer mediaElt={`#${props.mediaElementId}`}
                           playing={props.isPlaying}
+                          play={props.play}
+                          pause={props.pause}
```
tf, line 26: I think this does not do anything. Originally setting the `playing` prop to true called play on Wavesurfer. Now it calls `playerActions.play`. But that's what caused `playerState.isPlaying` to become true in the first place. If I remove the `playing`/`isPlaying`, `play` and `pause` props here and in `WavefromPlayerControls/index.js` everything appears to work the same way. This makes sense since Wavesurfer already operates directly on the underlying media element.

This is also the way [it used to work in Paged](https://github.com/codevise/pageflow/blob/1f1a694151db27a0d5a535c279a0ffea78de72bb/entry_types/paged/packages/pageflow-paged-react/src/media/components/WaveformPlayerControls/Waveform.jsx#L25).

awaismbajwa: You are right, I included this isPlaying property while trying to make waveform peaks visible previously, and let it remain there after the problem was fixed.

PR #1542, opened 2020-07-31
Fix use iframe height hook

awaismbajwa:
REDMINE-17670


aviav:
Thanks for pushing the sample. Closing now, consolidated PR is merged

PR #1553, opened 2020-08-07
Config for entry template

aviav:
REDMINE-17793

### DIFF
--- a/app/models/pageflow/entry_template.rb
+++ b/app/models/pageflow/entry_template.rb
@@ -4,6 +4,7 @@ module Pageflow
     include SerializedConfiguration
     serialize :default_share_providers, JSON
     belongs_to :account
+    delegate :enabled_feature_names, to: :account
     has_many :widgets, as: :subject, dependent: :destroy
 
     validates :account, presence: true
@@ -22,7 +23,7 @@ module Pageflow
     end
 
     def resolve_widgets(options = {})
-      widgets.resolve(Pageflow.config_for(account), options)
+      widgets.resolve(Pageflow.config_for(self), options)
     end
 
     def copy_defaults_to(revision)
@@ -58,7 +59,7 @@ module Pageflow
     end
 
     def available_themes
-      Pageflow.config_for(account).themes
+      Pageflow.config_for(self).themes
     end
 
     def hashify_provider_array(arr)
--- a/app/models/pageflow/entry_template.rb
+++ b/app/models/pageflow/entry_template.rb
@@ -7,14 +7,18 @@ module Pageflow
     has_many :widgets, as: :subject, dependent: :destroy
 
     validates :account, presence: true
-    validates :entry_type, presence: true
-    validates :entry_type,
+    validates :entry_type_name, presence: true
+    validates :entry_type_name,
               uniqueness: {
                 scope: :account
               }
 
-    def translated_entry_type
-      I18n.t("activerecord.values.pageflow/entry.type_names.#{entry_type}")
+    def entry_type
+      Pageflow.config.entry_types.find_by_name!(entry_type_name)
+    end
+
+    def translated_entry_type_name
+      I18n.t("activerecord.values.pageflow/entry.type_names.#{entry_type_name}")
     end
 
     def resolve_widgets(options = {})
--- a/spec/controllers/admin/entry_templates_controller_spec.rb
+++ b/spec/controllers/admin/entry_templates_controller_spec.rb
@@ -391,5 +391,22 @@ module Admin
           .to eq('signal' => true)
       end
     end
+
+    describe '#index' do
+      it 'redirects to account page' do
+        admin = create(:user, :admin)
+        sign_in(admin, scope: :user)
+        account_id = admin.accounts.first.id
+
+        get(:index,
+            params: {
+              account_id: account_id
+            })
+
+        expect(response).to redirect_to admin_account_path(
+          id: account_id, tab: 'entry_templates'
+        )
+      end
+    end
   end
 end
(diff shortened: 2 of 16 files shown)


aviav:
I like the idea. Yesterday, I thought more along the lines of "don't we already test all the parts somehow?", but indeed, it sounds like a good idea to me to test this in a more integrated way. Also, I didn't really check what part of this we already cover by existing integration tests. If you also think adding integration tests for this makes sense, that's enough reason for me to do it

tf:
Yeah, we somehow already had all the required parts and only needed to assemble them in a slightly different way. The admin controller spec looks like a good place to observe the new behavior since it allows us to verify that filtering config for entry template type works and that the admin actually does filter the config using the entry template. Approaching the problem from a TDD perspective, this probably would have been the spec I would have liked to see fail before starting to implement the feature.

aviav:
Indeed, as long as everything is synchronous, controller specs also work

spec/controllers/admin/entry_templates_controller_spec.rb
```diff
@@ -4,6 +4,93 @@ module Admin
...
+        sign_in(admin, scope: :user)
+
+        get(:new,
+            params: {
+              account_id: admin.accounts.first.id,
+              entry_type_name: 'ironed'
+            })
+
+        expect(response.body).to have_text('iron_ore')
+        expect(response.body).not_to have_text('carbon')
```
tf, line 30: **thought**: Should one use a more specific selector here?

Since the names are such original, I see the advantage of just asserting "this shall not be mentioned". On the other hand I had to stop and think whether `have_text` actually matches text inside `<option>` tags. Using a more specific matcher like `have_selector('[name="entry_template[theme_name]"] option', text: 'iron_ore')` etc. would maybe make it clearer what we are actually looking for here.

app/models/pageflow/entry_template.rb
```diff
@@ -4,17 +4,22 @@ class EntryTemplate < ApplicationRecord
...
-      I18n.t("activerecord.values.pageflow/entry.type_names.#{entry_type}")
+    def entry_type
+      Pageflow.config.entry_types.find_by_name!(entry_type_name)
+    end
+
+    def translated_entry_type_name
+      I18n.t("activerecord.values.pageflow/entry.type_names.#{entry_type_name}")
     end
 
     def resolve_widgets(options = {})
```
tf, line 25: **issue**: Theme validation does not consider entry type

I just saw this now and cannot comment on the correct line since it is not part of the change, but `available_themes` a few lines below still uses `config_for(account)`, which allows using all themes, paged and scrolled, that are available for the account, even though the select box only offers a subset. In Dedicated we now already have examples of themes that are registered for specific entry types.

aviav: For `resolve_widgets`, this should also be fixed, right?

aviav: > For `resolve_widgets`, this should also be fixed, right?

PS: Only usage of `resolve_widgets` on entry templates is for the editor

tf: Hm. Not even sure this is used. But yes, would make sense to fix it there as well.

aviav: There are definitely widgets with widget type `Pageflow::EntryTemplate` in the database. My guess is that we list widgets that are available for batch update using `index` action in `WidgetsController`

aviav: One more thing: We actually got a failing spec after switching `available_themes` to entry template config, because entry templates factory uses `'default'` theme. I would fix this by adding `config.themes.register(:default)` to `spec/support/pageflow/dummy/config/pageflow.rb`. What do you think?

tf: That's a bit too global for my taste. I faced a similar problem when introducing entry type specific themes and therefore extended the `TestEntryType` helper to [register a `default` theme](https://github.com/codevise/pageflow/blob/76d95420e2608c9488ced3854adf70fa58e0e2de/spec/support/helpers/test_entry_type.rb#L7). Can that maybe also be used in this situation?

spec/controllers/admin/entry_templates_controller_spec.rb
aviav: Indeed, the original names are meant to work around incidental matching. Matching this specific selector would be a coupling of the spec to Active Admin's current implementation, if there isn't some longer-lasting convention at work here that I'm not aware of.

Even if we just match for a select option with that text or whatever kind of input, we'd depend somewhat on implementation. We might at some point want to use some strange searchable select that doesn't use a select input under the hood, or maybe Active Admin starts using a seed json to populate form inputs, or something like that. I can still see how matching for a select option would make this spec more readable, and it might be unlikely that we switch this part of the implementation in a way that it doesn't use a select option. Also, it might happen that somebody manages to do something very strange like putting all available theme names into some other visible element.

Still, given Pageflow's long lifecycle, I remain in favor of just leaving this as it is. I have the feeling that this way, these specs might be less likely to break because of unrelated changes, and I think the risk of other strange changes that make these specs pass even though they shouldn't is negligible

app/models/pageflow/entry_template.rb
aviav: Ah yes, that sounds like a better idea. I'll try that then

spec/controllers/admin/entry_templates_controller_spec.rb
tf: Yeah, you are probably right. Let's leave it like this.

PR #1559, opened 2020-08-13
Hide mobile sharing icons on desktop

UsmanMuhammad:
REDMINE-17797

Mobile sharing icons don't need to be
displayed in a desktop version.

Since editor has an emulation mode which
won't work with a normal mobile device identifying
hook. It creates separate views for editor and published view

In view it checks for the device and
simply hides the sharing icons.

In editor it listens for a post message
on emulation mode event change and hides
the icons accordingly


entry_types/scrolled/package/src/frontend/inlineEditing/SharingTooltip.js
```diff
@@ -0,0 +1,64 @@
...
+                  clickable={true}
+                  offset={{right: -64}}
+                  className={classNames(headerStyles.navigationTooltip,
+                                        styles.sharingTooltip)}>
+      <div onMouseLeave={() => { ReactTooltip.hide() }}>
+        {renderShareLinks(shareProviders)}
+      </div>
+    </ReactTooltip>
+  )
+}
```
tf, line 64: Instead of replacing the `SharingTooltip` with an inline editing specific component, I'd suggest

* introducing a `PhonePlatformProvider` component in `src/frontend` which calls `browser.has` and renders the `PhonePlatformContext.Provider` 
* adding an inline editing replacement for `PhonePlatformProvider` which contains the above post message listening code and state and renders a `PhonePlatformContext.Provider` based on that state.
* defining a `usePhonePlatform` hook that is a wrapper for `useContext(PhonePlatformContext)`

This has a couple of advantages:

* No need to duplicate the code of `SharingTooltip`
* `usePhonePlatform` can then be used freely in any component and will work correctly in inline editing mode. With the current approach, we would have to create an inline editing alternative of every component that is supposed to depend on phone platform.

entry_types/scrolled/package/src/frontend/navigation/SharingMenu.js
```diff
@@ -3,12 +3,15 @@ import classNames from 'classnames';
...
-import {SharingTooltip} from "./SharingTooltip";
+import {SharingTooltip} from "../navigation/SharingTooltip";
 import ReactTooltip from "react-tooltip";
 import {useShareProviders} from "../../entryState";
 
+import PhonePlatformContext from './PhonePlatformContext';
+
 export function SharingMenu() {
-  const shareProviders = useShareProviders();
+  const useIsPhonePlatform = React.useContext(PhonePlatformContext);
```
tf, line 13: Reserve names of form `useXyz` for hook functions. The variable could be called `isPhonePlatform` instead.

UsmanMuhammad: Done

entry_types/scrolled/package/src/editor/controllers/PreviewMessageController.js
```diff
@@ -85,6 +85,13 @@ export const PreviewMessageController = Object.extend({
           })
         );
 
+        this.listenTo(this.entry, 'change:emulation_mode', entry =>
+          postMessage({
+            type: 'CHANGE',
```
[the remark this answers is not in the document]
UsmanMuhammad: Done

entry_types/scrolled/package/src/frontend/inlineEditing/SharingTooltip.js
UsmanMuhammad: Changed the logic according to the suggestion. Please, see the commit.

entry_types/scrolled/package/src/frontend/navigation/SharingMenu.js
```diff
@@ -7,8 +7,11 @@ import {SharingTooltip} from "./SharingTooltip";
 import ReactTooltip from "react-tooltip";
 import {useShareProviders} from "../../entryState";
 
+import {usePhonePlatform} from '../usePhonePlatform';
+
 export function SharingMenu() {
-  const shareProviders = useShareProviders();
+  const IsPhonePlatform = usePhonePlatform();
+  const shareProviders = useShareProviders(IsPhonePlatform);
```
tf, line 14: **nitpick**: local variables should start with a lower case letter

entry_types/scrolled/package/src/frontend/inlineEditing/PhonePlatformProvider.js
```diff
@@ -0,0 +1,30 @@
...
+      if (event.data.type === 'CHANGE_EMULATION_MODE') {
+        if(event.data['payload'] === 'phone') {
+          setPhoneEmulationMode(true);
+        }
+        else {
+          setPhoneEmulationMode(false);
+        }
+      }
+    }
+    return () => document.removeEventListener('message', receive);
```
tf, line 22: **issue**: inconsistent event registration

Event is added to `window` and removed from `document`. If we had a spec that unmounts the component and post a message afterwards, this would have lead to a warning about manipulating state on an unmounted component.

entry_types/scrolled/package/src/frontend/PhonePlatformProvider.js
```diff
@@ -0,0 +1,20 @@
...
+import {browser} from 'pageflow/frontend';
+
+import {withInlineEditingAlternative} from './inlineEditing';
+
+export const PhonePlatformProvider = withInlineEditingAlternative('PhonePlatformProvider', function PhonePlatformProvider({children}) {
+
+  var isPhonePlatform = false;
+  if (browser.has('phone platform')) {
+    isPhonePlatform = true;
+  }
```
tf, line 13: **nitpick**: This could just be an assignment.

entry_types/scrolled/package/src/entryState/sharing.js
```diff
@@ -72,23 +72,34 @@ export function useShareProviders() {
...
+function filterShareProviders(shareProvidersConfig, isPhone) {
+  if (!isPhone) {
+    return Object.keys(shareProvidersConfig).filter(function (provider) {
+      return (provider !== 'telegram' && provider !== 'whats_app');
+    });
+  }
+
+  return Object.keys(shareProvidersConfig);
+}
+
```
tf, line 102: **suggestion**: Let's also cover this new behavior with  a spec.

entry_types/scrolled/package/spec/frontend/usePhonePlatform/inFrontend-spec.js
```diff
@@ -0,0 +1,22 @@
...
+  it('returns true on phone platform', () => {
+
+    jest.spyOn(browser, 'has').mockReturnValue(true);
+
+    const {result} = renderHook(() => usePhonePlatform(), {wrapper: PhonePlatformProvider});
+
+    expect(result.current).toEqual(true);
+  });
+
+});
```
tf, line 22: **suggestion**: Always test both cases for a boolean function.

Both tests for `usePhonePlatform` would also pass if the implementation of the hook was just `return true`.

entry_types/scrolled/package/spec/frontend/usePhonePlatform/inEditorPreview-spec.js
```diff
@@ -0,0 +1,36 @@
...
+  
+  async function asyncHandlingOfMessage(callback) {
+    await act(async () => {
+      callback();
+      // Ensure React update triggered by async handling of message is
+      // wrapped in act
+      await tick();
+    });
+  }
+});
```
tf, line 36: **suggestion**: Extract test helper

As already suggested [here](https://redmine.codevise.de/issues/17670#note-23) by @aviav, now that we have the third place in the code base that uses this "await-act-tick-trick" (here are the [other](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/frontend/features/scrollToSectionMessage-spec.js#L26) [two](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/contentElements/dataWrapperChart/useIframeHeight-spec.js#L48)), it might be time to extract the helper.

When we move the function to `spec/support/`, I'd suggest to rename it to `asyncHandlingOf` since the helper itself does not really care about what is handled asynchronously. Together with the `await` and the actual callback, the whole line will read as a sentence that says asynchronous handling of what is being awaited.

entry_types/scrolled/package/spec/editor/controllers/PreviewMessageController-spec.js
```diff
@@ -341,6 +341,27 @@ describe('PreviewMessageController', () => {
...
+
+    await postReadyMessageAndWaitForAcknowledgement(iframeWindow);
+
+    return expect(new Promise(resolve => {
+      iframeWindow.addEventListener('message', event => {
+        if (event.data.type === 'CHANGE_EMULATION_MODE') {
+          resolve(event.data);
+        }
+      });
+      entry.trigger('change:emulation_mode', entry.sections.first());
```
tf, line 362: **suggestion(if-minor)**: Trigger effect by changing attribute

In cases like this, it can be safer to perform the action that causes the event to be dispatched (i.e. changing the attribute) instead of triggering the event itself. That way we can be sure the actual data send with the event (if the handler depends on it) is used correctly and not just that from our fixture event.

entry_types/scrolled/package/spec/frontend/usePhonePlatform/inEditorPreview-spec.js
```diff
@@ -0,0 +1,36 @@
...
+import {loadInlineEditingComponents} from 'frontend/inlineEditing';
+
+import {renderHook, act} from '@testing-library/react-hooks';
+
+import '@testing-library/jest-dom/extend-expect'
+import {tick} from "support";
+
+
+describe('usePhonePlatform', () => {
+  beforeAll(loadInlineEditingComponents);
```
tf, line 13: **praise**: Great that you found out how to cover both the editor and frontend parts in the specs. :+1:

UsmanMuhammad: Thanks :+1:

entry_types/scrolled/package/spec/frontend/usePhonePlatform/inEditorPreview-spec.js
UsmanMuhammad: While replacing the helper [here](https://github.com/codevise/pageflow/blob/master/entry_types/scrolled/package/spec/frontend/features/scrollToSectionMessage-spec.js#L26), it doesn't work for some reason. I have tried a lot.
Any idea?

entry_types/scrolled/package/spec/editor/controllers/PreviewMessageController-spec.js
UsmanMuhammad: done

entry_types/scrolled/package/spec/frontend/usePhonePlatform/inEditorPreview-spec.js
tf: The problem is that `render` from `@testing-library/react` and  `renderHook` from `@testing-library/react` each use a different implementation of `act`. We therefore cannot have a shared `asyncHandlingOf` helper that works for both hook and component test.

For hooks, I found that there is also [`waitForNextUpdate`](https://react-hooks-testing-library.com/usage/advanced-hooks#async) which solves a similar problem. But this does not help in the case of `useIframeHeight` when we want to assert that no change happens. There is no way to "wait for nothing to happen".

My best idea so far therefore is to have two copies of the helper: one in `spec/support/asyncHandlingOf/index.js` that uses `act` form `@testing-library/react`, the other in `spec/support/asyncHandlingOf/forHooks.js` that uses `act` from `@testing-library/react-hooks`.

entry_types/scrolled/package/src/entryState/sharing.js
```diff
@@ -31,7 +31,7 @@ import WhatsAppIcon from "../frontend/assets/images/navigation/icons/social/what
  *     }
  *   ]
  */
-export function useShareProviders() {
+export function useShareProviders(isPhone) {
```
tf, line 34: One last thing that I missed earlier. Let's use an options object here, so that the call looks like `useShareProvider({isPhonePlatform: true})` instead of `useShareProvider(true)`. In the latter case you have to look at the definition of the hook to find out what the `true` means.

entry_types/scrolled/package/src/entryState/sharing.js
```diff
@@ -72,26 +72,26 @@ export function useShareProviders(isPhone) {
...
+    return activeShareProviders(shareProviders, options.isPhonePlatform).map((provider) => {
       const config = sharing[provider];
       return ({
         name: config.name,
         icon: config.icon,
         url: config.url
       })
     })
-  }, [shareProviders, isPhone]);
+  }, [shareProviders, options]);
```
tf, line 83: Since the `options` object will be a new object on each render, this defeats the purpose of `useMemo`. You can instead destructure the options right in the parameter list (i.e. `useShareProviders({isPhonePlatform})`) and pass the boolean here as before

UsmanMuhammad: Done

PR #1565, opened 2020-08-19
Improve scrolled editor blank slate

tf:
* Blank slate styles in `pageflow/editor` use Source Sans Pro. So far
  that was only defined in `pageflow/application.scss`. Also import it
  `pageflow/editor.scss` to make it available for Pageflow
  Scrolled. For Paged the duplicate definition should not hurt.

* Since the blank slate page disappears as soon as the first section
  has been added, it does not make sense to include steps with
  instructions about adding and editing content elements. Keep a third
  item since there maybe is a chance that the user reads ahead one
  item and since a two item list looks weird.

REDMINE-17841

### DIFF
--- a/app/assets/stylesheets/pageflow/editor/base.scss
+++ b/app/assets/stylesheets/pageflow/editor/base.scss
@@ -1,5 +1,7 @@
 @import "bourbon";
 
+@import "pageflow/fonts";
+
 @import "pageflow/jquery_ui";
 @import "pageflow/ui";
 @import "pageflow/animations";
--- a/entry_types/scrolled/config/locales/new/de.yml
+++ b/entry_types/scrolled/config/locales/new/de.yml
@@ -552,8 +552,6 @@ de:
         create_chapter: Klicke jetzt auf <em>Neues Kapitel</em>, um ein erstes Kapitel zu erstellen.
         create_section: Klicke dann auf <em>Neuer Abschnitt</em>, um dem Kapitel einen ersten Abschnitt hinzuzufügen.
         create_content_element: Über die <em>Plus-Buttons</em> in der Vorschau kannst Du Elemente hinzufügen.
-        edit_content_element: Klicke auf Elemente in der Vorschau, um ihre Einstellungen in der Seitenleiste zu bearbeiten.
-        edit_section: Klicke auf eine freie Fläche, um den Abschnitt zu markieren.
         header: Dies ist ein leerer Pageflow.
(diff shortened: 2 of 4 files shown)


aviav:
**nitpick**: I would really let point 2 include the text from point 3, and drop point 3. Of course there's a chance that the customer reads ahead one point, but the expectation would be that point 3's instruction doesn't just vanish if they don't read ahead, which was the main reason we introduce this change in the first place.

As to looking weird: This is probably a case where it looks weird if and only if you have known a version of the list with more than two items.

If you say due to time pressure this should be merged regardless, just let me know

tf:
Just for reference, here's the way the PR currently makes the view look:

Bildschirmfoto von 2020-08-20 11-24-17

By that logic, wouldn't we need to remove the footer sentence about the preview as well? I see your point, but the way I look at it there is also value in having the three item list with the three aligned bold parts. Those are the ones that stand out to me when first scanning the page: New chapter, New section, Plus buttons.

This (for me) works less well, when the items are comined:

Bildschirmfoto von 2020-08-20 11-28-29

aviav:
> By that logic, wouldn't we need to remove the footer sentence about the preview as well?

No, we wouldn't. The footer text isn't needed as instruction, because the change in the area's content is instructive enough. In terms of accessibility, the sentence provides a marginal benefit, but accessibility and the editor is a whole other issue. We could consider removing the sentence, but I don't think it's important enough to discuss in the context of this change

> three aligned bold parts

We could have three aligned bold parts that are spread over two points, by breaking the line. I agree that this is a valuable feature

tf:
I have a bit of hard time imagining the best way to make this happen: [Just removing the number prefix and keeping the same margings](https://user-images.githubusercontent.com/26554/90757758-53840700-e2de-11ea-90d2-723ae29e21cb.png) to me looks like someone forgot the number. [Using different margins to group things logically](https://user-images.githubusercontent.com/26554/90757759-541c9d80-e2de-11ea-9822-172e831998c1.png) destroy the visual rhythm of the section.

I feel this might be turning into a bit of bike shed. Since the customer initiated the change, there is a good chance they will come up with the same criticism. Maybe they also come up with additional ideas, which can then be incorporated in a future step.

aviav:
This *is* a bikeshed, but one that has been noticed by the customer. I completely agree with your aesthetic judgement and ideas on how to proceed. For completeness, I'll add your drafts for alternative bikesheds to the Redmine issue, but merge this now

PR #1566, opened 2020-08-19
Improve editor inline help and labels

tf:
* Inline help for entry locale

* Better labels text for invert section option

* Better inline help for full viewport height section option

REDMINE-17841

### DIFF
--- a/entry_types/scrolled/config/locales/new/en.yml
+++ b/entry_types/scrolled/config/locales/new/en.yml
@@ -398,16 +398,18 @@ en:
           fullHeight:
             label: Use Full Viewport-Height
             inline_help: >
-              Activate this option if the section should span at least one viewport height.
-              Deactivate this option to create a short section which height is determined by its
-              foreground content. This option needs to be activated for two adjacent sections
-              to be able to use a fade transition between them.
+              Activate this option if the background of this section (e.g. image, color or video) is supposed
+              to be enlarged such that the whole browser window (the so called viewport) is covered.
+              Deactivate this option to create a short section. The height of the section and its background
+              is then determined only by the content of the section.  Multiple sections with different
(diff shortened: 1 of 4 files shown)


entry_types/scrolled/config/locales/new/en.yml
```diff
@@ -398,16 +398,18 @@ en:
           fullHeight:
             label: Use Full Viewport-Height
             inline_help: >
-              Activate this option if the section should span at least one viewport height.
-              Deactivate this option to create a short section which height is determined by its
-              foreground content. This option needs to be activated for two adjacent sections
-              to be able to use a fade transition between them.
+              Activate this option if the background of this section (e.g. image, color or video) is supposed
+              to be enlarged such that the whole browser window (the so called viewport) can be covered.
```
aviav, line 402: **nitpick**: 'is covered' instead of 'can be covered' would be clearer. 'Can be' sounds like there could be more to be done until it actually happens

aviav (changes requested):

entry_types/scrolled/config/locales/new/en.yml
tf: Well, at least when not using a fade transition, it depends on the scroll position whether the viewport is actually covered. It will be covered once you scroll far enough that all parts of the previous section have vanished.

aviav: While this is a plausible justification of the wording, still, we have inconsistency between German and English text here. The eventuality behind 'can be' is self-explaining, and therefore I'd still say 'is' is better

tf: Fixed.

aviav (approved):

PR #1568, opened 2020-08-20
Let first chapter begin at the very top

tf:
Prevent scrolling when clicking the first chapter link after loading
the document. Before, the first chapter began right below the
navigation bar. Now the navigation bar no longer has a height. We have
added a padding to the first chapter to prevent content or the section
selection rect from being hidden behind the navigation bar.

REDMINE-17874

### DIFF
--- a/entry_types/scrolled/package/src/frontend/Entry.module.css
+++ b/entry_types/scrolled/package/src/frontend/Entry.module.css
@@ -5,3 +5,9 @@
   background-color: #000;
   color: #fff;
 }
+
+.Entry > div:first-child {
+  /* Let content begin below navigation bar. Navigation bar has zero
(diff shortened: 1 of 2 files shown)


entry_types/scrolled/package/src/frontend/Entry.module.css
```diff
@@ -5,3 +5,9 @@
   background-color: #000;
   color: #fff;
 }
+
+.Entry > div:first-child {
+  /* Let content begin below navigation bar. Navigation bar has zero
+     height to let first chapter start at the very top. */
```
aviav, line 11: **praise:** This is a well-chosen use for an inline comment. The line that it describes might change if we just need to adjust the value slightly, which could make the original commit message harder to find

PR #1569, opened 2020-08-20
Use theme specific favicons

UsmanMuhammad:
REDMINE-17779

It add a helper to render theme
specific favicons paced in hosted
application for scrolled entries


entry_types/scrolled/doc/creating_themes.md
```diff
@@ -19,3 +19,7 @@ Register the theme in the Pageflow initializer:
     end
 
 Copy all options from the default theme and make the desired changes.
+
+Place the favicons next to logos in the theme. Please create favicons
+according to the devices for performance reasons and replace them with
+the default favicons.
```
tf, line 25: > Place the favicons next to logos in the theme.

**suggestion**: Place favicon files in sub directory of theme directory

Among all these favicon files, other theme images are hard to spot. Since all these files are about the same thing and are probably generated at once, I suggest placing them in a `favicons` sub directory of the theme directory.

> Please create favicons ￼according to the devices for performance reasons

**question**: What is the recommended way to generate these files?

Did you use one of the favicon generators? I think it would be handy if the file names used here matched up with what one of those services generates. We could then name that service here.

app/views/pageflow/favicons/_entry.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if icon_path.present? %><link rel='icon' href='<%= icon_path %>' type='image/ico'><% end %>
+<% if icon_32_path.present? %><link rel='icon' type='image/png' sizes='32x32' href='<%= icon_32_path %>'><% end %>
+<% if icon_16_path.present? %><link rel='icon' type='image/png' sizes='16x16' href='<%= icon_16_path %>'><% end %>
+
+<% if apple_touch_icon_path.present? %><link rel='apple-touch-icon' sizes='180x180' href='<%= apple_touch_icon_path %>'><% end %>
+<% if webmanifest_path.present? %><link rel='manifest' href='<%= webmanifest_path %>'><% end %>
+<% if safari_pinned_tab_path.present? %><link rel='mask-icon' href='<%= safari_pinned_tab_path %>' color='#5bbad5'><% end %>
```
tf, line 7: **question**: In which case will these conditions be `false`?

app/views/pageflow/favicons/_entry.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<meta name='msapplication-TileColor' content='#da532c'>
+<meta name='theme-color' content='#ffffff'>
```
tf, line 9: **thought**: These colors could be theme options

Like this, there is no way for a theme to change these colors. They could be specified the same way `logo_alt_text` is specified in the host application.

entry_types/scrolled/spec/helpers/pageflow_scrolled/favicon_helper_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        )
+
+        expect(html).to have_css(
+          %{meta[name='theme-color']},
+          visible: false,
+          count: 1
+        )
+      end
+    end
+  end
```
tf, line 56: **praise**: Nice tests

The selector based approach makes them really concise. :+1:

tf (changes requested):

entry_types/scrolled/doc/creating_themes.md
UsmanMuhammad: Placed the favicons in a sub-directory. 

Yes, the source that you mentioned in the ticket is the one. It is really helpful and creates all these files, you can further edit those files like background, and border.
Furthermore, the names are just as they are mentioned when generated and the source also gives all the html tags to include as well using the same names.

I will name the source and give a link in the docs.

app/views/pageflow/favicons/_entry.html.erb
UsmanMuhammad: At first thought, I thought if there is no file then it should return false and the tag won't be rendered. But, if there isn't a file then it returns an error. Removed these.

entry_types/scrolled/spec/helpers/pageflow_scrolled/favicon_helper_spec.rb
UsmanMuhammad: Thanks :+1:

tf (approved):

PR #1574, opened 2020-09-02
Renders alt text for media

UsmanMuhammad:
REDMINE-17864

It renders alt texts if provided or empty
text for all background and inline images,
videos, and audios

beforeAfter images already had a mechanism
to render alt texts

We already get the files in video, and audio player.
It sends the value to MediaPool where it adds the attribute
to the player

### DIFF
--- a/entry_types/scrolled/package/spec/frontend/AudioPlayer-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/AudioPlayer-spec.js
@@ -175,6 +175,6 @@ describe('AudioPlayer', () => {
         })
       });
 
-    expect(result.container.querySelector('audio')).toHaveAttribute('alt', ' ');
+    expect(result.container.querySelector('audio').hasAttribute('alt')).toBe(true);
   });
 });
--- a/entry_types/scrolled/package/spec/frontend/AudioPlayer-spec.js
+++ b/entry_types/scrolled/package/spec/frontend/AudioPlayer-spec.js
@@ -136,4 +136,45 @@ describe('AudioPlayer', () => {
       })
     );
   });
+
+  it('renders alt text', () => {
+    function getAudioFileSeedWithConfiguration({
+      id = 1,
+      permaId = 100,
+      basename = 'audio'
+    } = {}) {
+      return {
+        fileUrlTemplates: {
+          audioFiles: {
+            mp3: ':id_partition/:basename.mp3',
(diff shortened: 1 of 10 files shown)


package/src/frontend/media/MediaPool.js
```diff
@@ -44,7 +44,8 @@ export class MediaPool {
 
     player = this.unAllocatedPlayers[playerType].pop();
     if (player) {
-      
+
+      player.el().children[0].setAttribute('alt', altText); //add alt texts;
```
tf, line 48: **suggestion**: Use `getMediaElement`

Two lines down we also change attributes on the media element. By using `getMediaElement`, we get around duplicating the knowledge that VideoJS wraps the media element in a div. The comment does not add any value.

entry_types/scrolled/package/spec/frontend/AudioPlayer-spec.js
```diff
@@ -136,4 +136,45 @@ describe('AudioPlayer', () => {
...
+
+  it('renders empty alt attr', () => {
+    const result =
+      renderInEntry(<AudioPlayer {...requiredProps()} id={100} />, {
+        seed: getAudioFileSeed({
+          permaId: 100
+        })
+      });
+
+    expect(result.container.querySelector('audio')).toHaveAttribute('alt', ' ');
```
tf, line 178: **question**: Why do we default to a space character?

package/src/frontend/media/MediaPool.js
tf: **suggestion**: Add test

This is covered indirectly through the React component tests. Those components belong to a different package (`pageflow-scrolled`), though. We could also add a test in the `pageflow` package. Similar to [the one for loop](https://github.com/codevise/pageflow/blob/93180fde9abe03ae24b7aff37969083a44bd3b96/package/spec/frontend/media/media_spec.js#L44)

tf (changes requested):

entry_types/scrolled/package/spec/frontend/AudioPlayer-spec.js
UsmanMuhammad: Removed that.

package/src/frontend/media/MediaPool.js
UsmanMuhammad: Done

tf (approved):

PR #1576, opened 2020-09-16
String literals in SQL should only be enclosed by single quotes

DTrierweiler:
Double quotes will work with mysql, but will break postgresql.
This fixes an error message after logging in with a non-admin user, while using a postgresql db.


tf:
Thanks! Are you able to run Pageflow with Postgres with these changes?

PR #1580, opened 2020-10-07
Add motif area support for background videos

tf:
REDMINE-17879

### DIFF
--- a/entry_types/scrolled/package/spec/support/rendering.js
+++ b/entry_types/scrolled/package/spec/support/rendering.js
@@ -18,7 +18,23 @@ import {normalizeSeed} from './normalizeSeed';
  * The `setup` option can be used to simulate rendering the component
  * in the editor where data is synchronized from Backbone models.
  *
- * @param {React.Component} ui - React component
+ * To be able to render components that expect the result of certain
+ * hooks as part of their props, instead of a React component, you can
+ * pass a function returning a React component as first parameter. The
+ * function will be evaluated in a context of a React component and
+ * can thus make use of hooks
+ *
+ *     // DOES NOT WORK
+ *     renderInEntry(<Image file={useFile({collectionName: 'imageFiles', permaId: 4})} />, {seed});
+ *
+ *     // WORKS
+ *     renderInEntry(() => <Image file={useFile({collectionName: 'imageFiles', permaId: 4})} />, {seed});
+ *
+ * When using the `rerender` function from the result, you again need
+ * to use the same type of parameter you passed to the original
+ * `renderInEntry` call.
+ *
+ * @param {React.Component|Function} ui - React component or function returning a React component
  * @param {Object} [options]
  * @param {Object} [options.seed] - Seed data for entry state. Passed through {@link normalizeSeed}.
  * @param {Function} [options.setup] -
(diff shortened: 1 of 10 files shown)


entry_types/scrolled/package/spec/support/rendering.js
```diff
@@ -27,11 +43,41 @@ import {normalizeSeed} from './normalizeSeed';
...
+    // `rerender` with a different function than the one passed to the
+    // original `renderInEntry` call, e.g.
+    //
+    //   const {rerender} = renderInEntry(() => <MyComponent someProp={1} />);
+    //   rerender(() => <MyComponent someProp={2} />);
+    //
+    // React would unmount the `MyComponent` component from the first
+    // render call and mount a new one. We therefore define a single
+    // component that is reused across rerenders to ensure
+    // `MyComponent` stays mounted and just receives new props.
```
aviav, line 64: **praise:** This comment answers very well precisely the questions about 'why?' that came up when I read the code below

entry_types/scrolled/package/spec/support/rendering.js
```diff
@@ -18,7 +18,23 @@ import {normalizeSeed} from './normalizeSeed';
...
  *
- * @param {React.Component} ui - React component
+ * To be able to render components that expect the result of certain
+ * hooks as part of their props, instead of a React component, you can
+ * pass a function returning a React component as first parameter. The
+ * function will be evaluated in a context of a React component and
+ * can thus make use of hooks
+ *
+ *     // DOES NOT WORK
+ *     renderInEntry(<Image file={useFile({collectionName: 'imageFiles', permaId: 4})} />, {seed});
```
aviav, line 28: **praise:** I think this is an error that is likely to happen for those who don't use `renderInEntry` frequently. Having 'does not work' and 'works' directly adjacent makes it easy to move on in this kind of situation

entry_types/scrolled/package/src/frontend/Backdrop.js
```diff
@@ -146,13 +157,20 @@ function BackgroundVideo(props) {
...
+                   textTracksDisabled={true}
+                   fit="cover"
+                   loop={true}
+                   playsInline={true} />
+      <MotifArea key={video.permaId}
+                 onUpdate={onMotifAreaUpdate}
+                 file={video}
+                 containerWidth={containerDimension.width}
+                 containerHeight={containerDimension.height}/>
+    </>
```
aviav, line 174: **nitpick:** This file is becoming long. We could move components to individual files. It is okay if this happens in a different PR than this one

entry_types/scrolled/package/src/frontend/Backdrop.js
```diff
@@ -146,13 +157,20 @@ function BackgroundVideo(props) {
...
+    <>
+      <VideoPlayer isPrepared={isPrepared}
+                   playerState={playerState}
+                   playerActions={playerActions}
+                   id={video.permaId}
+                   textTracksDisabled={true}
+                   fit="cover"
+                   loop={true}
+                   playsInline={true} />
+      <MotifArea key={video.permaId}
```
aviav, line 169: **question:** Is it realistic to have some spec that tests some effect/s of this addition?

aviav (changes requested):
Nice PR! Apart from requesting a spec if it seems feasible, only minor stuff remains, as far as I see

entry_types/scrolled/package/src/frontend/Backdrop.js
tf: Good idea. I added tests that ensure that the `onMotifAreaUpdate` callback is invoked. The specifics of when/how the callback is called are already covered in the `MotifArea` tests. Here, the fact that the callback is called merely acts as proof that a motif area has been rendered and that this now also happens for video files.

entry_types/scrolled/package/src/frontend/Backdrop.js
tf: I agree. Since I already have two more PRs lined up that build upon the changes here, I'll try to keep this in mind for later.

entry_types/scrolled/package/src/editor/views/EditSectionView.js
```diff
@@ -8,6 +8,19 @@ export const EditSectionView = EditConfigurationView.extend({
   translationKeyPrefix: 'pageflow_scrolled.editor.edit_section',
 
   configure: function(configurationEditor) {
+    const editMotifAreaMenuItem = {
+      name: 'editMotifArea',
+      label: I18n.t('pageflow_scrolled.editor.edit_section.edit_motif_area'),
```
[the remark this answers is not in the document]
tf: I changed the menu item and header text to talk about "motif areas" instead of "image motifs". I left the prompt to select the most important part of the image unchanged since we only display a poster image. So technically it is not completely wrong after all.

I also added a checklist item to REDMINE-17296 as a reminder to also make the help entry talk about background videos instead of only images.

entry_types/scrolled/package/src/frontend/Backdrop.js
aviav: Better than trying to keep it in mind would be to add a checklist item to any Redmine issue that is suitable according to you

PR #1585, opened 2020-10-19
Implements skip link for navigation

UsmanMuhammad:
REDMINE-17403

It adds a skip link for better navigation
It adds an invisible link to the content and
the same functionality can be extended for further
components as well

While navigation using keyboard a link appear which refers
to the content. It increases usability for users who wants
to skip navigation bar and go directly to content

It renders a button which will only be visible on tab press
and will disappear once clicked or skipped

Works with both mouse click and enter


entry_types/scrolled/package/src/frontend/navigation/AppHeader.module.css
```diff
@@ -151,3 +151,11 @@
...
   }
 }
+
+@media (min-width: 1025px) {
+    .chapterList {
+      width: 55%;
+      position: absolute;
+      left: 23%;
+    }
+}
```
tf, line 161: Why is this needed?

UsmanMuhammad: The chapter list covers the whole navigation bar. So, if a user skips links one time then he has to use keyboard to use the skip link again. 
So if a user clicks on navigation bar and hit tab then without this the first chapter list item is focused.
This also doesn't provide a considerable solution as it only makes the chapter list smaller so the user has some space between the chapter list and the logo to click and use the skip link.

UsmanMuhammad: I removed this for now. Let me know if this makes sense and I will add it again or any other idea you have for that.

tf: Yeah, I think that's fine. The user can always return to the top of the tab order by focusing the location bar (Crtl+L etc.)

tf (approved):

PR #1633, opened 2020-11-30
WIP: Fix BeforeAfter wiggle effect

awaismbajwa:
REDMINE-18076

### DIFF
--- a/entry_types/scrolled/package/src/contentElements/inlineBeforeAfter/BeforeAfter.module.css
+++ b/entry_types/scrolled/package/src/contentElements/inlineBeforeAfter/BeforeAfter.module.css
@@ -26,13 +26,13 @@
 /* With react-compare-image 2.0.4 (commit 7410d14), this selects the
    slider */
 .container.wiggle div div:nth-child(3) {
-  animation: SliderLeftRightShake 1.5s cubic-bezier(.36,.07,.19,.97) both;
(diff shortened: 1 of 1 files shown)


tf:
I have the impression that this change rather works around the issue introduced in #1623 instead of addressing the root cause. The way I understood the change, before #1623, the 100% keyframe was undefined which lead to an undesirable effect at the end of the animation in Safari. #1623 then defined 100%, but set it to an off center position. Since fill mode was set to `both`, the slider remained in this off center position. This PR now changes fill mode to use the first keyframe, which is again not defined explicitly. Should not the 100% keyframe be set to the correct value instead?

awaismbajwa:
> I have the impression that this change rather works around the issue introduced in #1623 instead of addressing the root cause. The way I understood the change, before #1623, the 100% keyframe was undefined which lead to an undesirable effect at the end of the animation in Safari. #1623 then defined 100%, but set it to an off center position. Since fill mode was set to `both`, the slider remained in this off center position. This PR now changes fill mode to use the first keyframe, which is again not defined explicitly. Should not the 100% keyframe be set to the correct value instead?

Yes, I agree with you but this was just my first solution as I was not sure what should be the correct value for 100% key frame ? I thought using the first keyframe can solve the problem, and this resolves the problem of last slider position. But there are more issues in the animation which can be observed through storybook entries of BeforeAfter.

tf:
see #1659.

PR #1637, opened 2020-12-02
Chapter deep linking for scrolled entries

awaismbajwa:
REDMINE-17393

### DIFF
--- a/entry_types/scrolled/package/spec/entryState/structure-spec.js
+++ b/entry_types/scrolled/package/spec/entryState/structure-spec.js
@@ -95,6 +95,7 @@ describe('useEntryStructure', () => {
   const expectedEntryStructure = [
     {
       title: 'Chapter 1',
+      chapterSlug: 'chapter-1',
       summary: 'An introductory chapter',
       sections: [
         {
@@ -109,6 +110,7 @@ describe('useEntryStructure', () => {
     },
     {
       title: 'Chapter 2',
+      chapterSlug: 'chapter-2',
       summary: 'A great chapter',
       sections: [
         {
@@ -199,26 +201,47 @@ describe('useChapters', () => {
     expect(chapters).toMatchObject([
       {
         permaId: 10,
-        chapterRef: 'chapter-1',
+        chapterSlug: 'chapter-1',
         title: 'Chapter 1',
         summary: 'An introductory chapter'
       },
       {
         permaId: 11,
-        chapterRef: 'chapter-2',
+        chapterSlug: 'chapter-2',
         title: 'Chapter 2',
         summary: 'A great chapter'
       }
     ]);
   });
 
-  it('escapes chapter titles', () => {
-    chaptersSeed[0].configuration.title = "<script>alert('test')</script>Titleä"
+  it('sanitizes chapter titles', () => {
     const {result} = renderHookInEntry(
       () => useChapters(),
       {
         seed: {
-          chapters: chaptersSeed
+          chapters: [
+            {
+              configuration: {
+                title: 'SmallCase',
+              }
+            },
+            {
+              configuration: {
+                title: 'RemöveUmläütß',
+              }
+            },
+            {
+              configuration: {
+                title: 'remove space',
+              }
+            },
+            {
+              configuration: {
+                title: 'remove#special$character',
+              }
+            },
+            
+          ]
         }
       }
     );
@@ -227,16 +250,16 @@ describe('useChapters', () => {
 
     expect(chapters).toMatchObject([
       {
-        permaId: 10,
-        chapterRef: "lessscriptgreateralert('test')lessscriptgreatertitlea",
-        title: chaptersSeed[0].configuration.title,
-        summary: 'An introductory chapter'
+        chapterSlug: "smallcase",
       },
       {
-        permaId: 11,
-        chapterRef: 'chapter-2',
-        title: 'Chapter 2',
-        summary: 'A great chapter'
+        chapterSlug: 'remoeveumlaeuetss',
+      },
+      {
+        chapterSlug: 'remove-space',
+      },
+      {
+        chapterSlug: 'removespecialdollarcharacter',
       }
     ]);
   });
--- a/entry_types/scrolled/package/spec/entryState/structure-spec.js
+++ b/entry_types/scrolled/package/spec/entryState/structure-spec.js
@@ -199,11 +199,42 @@ describe('useChapters', () => {
     expect(chapters).toMatchObject([
       {
         permaId: 10,
+        chapterRef: 'chapter-1',
         title: 'Chapter 1',
         summary: 'An introductory chapter'
       },
       {
         permaId: 11,
+        chapterRef: 'chapter-2',
+        title: 'Chapter 2',
+        summary: 'A great chapter'
+      }
+    ]);
+  });
+
+  it('escapes chapter titles', () => {
+    chaptersSeed[0].configuration.title = "<script>alert('test')</script>Title"
+    const {result} = renderHookInEntry(
+      () => useChapters(),
+      {
+        seed: {
+          chapters: chaptersSeed
+        }
+      }
+    );
+
+    const chapters = result.current;
+
+    expect(chapters).toMatchObject([
+      {
+        permaId: 10,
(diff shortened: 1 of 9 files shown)


awaismbajwa:
Tried speakingurl for handling of umlauts and special character in url fragment.
One issue was that it didnot work on server side and another issue was that somehow hash was not getting encoded.
For now just used javascript escape function. Will change it after finding appropriate solution.

entry_types/scrolled/package/src/frontend/Entry.js
```diff
@@ -18,13 +18,20 @@ export default withInlineEditingDecorator('EntryDecorator', function Entry(props
...
-
   useSectionChangeEvents(entryStructure, currentSectionIndex);
 
   const setCurrentSectionIndex = useCallback(index => {
     sectionChangeMessagePoster(index);
     setCurrentSectionIndexState(index);
-  }, [setCurrentSectionIndexState]);
+    let sectionCounter = 0;
+    let chapter = entryStructure.find(chapter => {
+      sectionCounter += chapter.sections.length-1;
```
tf, line 28: **question**: Why `length - 1`?

Doesn't this mean that `sectionCount` will never be greater than 0 if all chapters only have one section? If we change the test above to simulate scrolling to the second section, we could make sure that this part of the logic is covered.

entry_types/scrolled/package/src/entryState/structure.js
```diff
@@ -105,12 +103,29 @@ export function useSectionContentElements({sectionId}) {
 
 export function useChapters() {
   const chapters = useEntryStateCollectionItems('chapters');
-
+  let chapterTitles = {};
   return chapters.map(chapter => {
+    let chapterRef = chapter.configuration.title;
+    if (chapterRef) {
+      if (slugify) {
```
tf, line 110: **question**: When would this be false?

entry_types/scrolled/package/spec/entryState/structure-spec.js
```diff
@@ -199,11 +199,42 @@ describe('useChapters', () => {
...
+    expect(chapters).toMatchObject([
+      {
+        permaId: 10,
+        chapterRef: "lessscriptgreateralert('test')lessscriptgreatertitlea",
+        title: chaptersSeed[0].configuration.title,
+        summary: 'An introductory chapter'
+      },
+      {
+        permaId: 11,
+        chapterRef: 'chapter-2',
```
tf, line 237: **suggestion**: Use separate seed only for this test

If we used an array of chapter seed data directly inside of this test, we could include more different examples. Since the passed [seed data is normalized](https://github.com/codevise/pageflow/blob/cf0fe63d1804eeb643530a906a4aef9020229b7b/entry_types/scrolled/package/spec/support/normalizeSeed.js#L131), we would only need to include the chapter titles and could skip all other unrelated data like `permaId` or `id`. This would make it easy to add additional examples in the future and also allow covering currently untested logic like making chapter slugs unique.

It also [looks like Slugify supports replacing umlauts](https://github.com/simov/slugify/blob/d2646e97cf1bb3784b38dbf7ca4a45b1a925b110/slugify.js#L16) if we pass the `locale` option. I'd also consider enabling the `strict` option to remove all other potentially problematic characters.

We could then replace the above "<script>" example, with more realistic examples for things like whitespace or special characters. As we discussed, script injection is not the vector we are addressing here. Therefore this test example could be confusing someone trying to understand this part of the code base.

tf: **suggestion**: Add slug property to expected result in `useEntryStructure` test

Since we depend on the slug to also be included in the data returned by `useEntryStructure` and actually changed the implementation of `useEntryStructure` to ensure this, we should also change the test. Since we know that all the details of escaping chapter slugs are already covered by the `useChapters` test, I'd suggest to use the existing test and only add the slug property to `expectedEntryStructure`.

tf (changes requested):

entry_types/scrolled/package/src/frontend/Entry.js
awaismbajwa: Yes you are right, thanks for pointing out. Fixed it now.

entry_types/scrolled/package/src/entryState/structure.js
awaismbajwa: While testing out different libraries, initial ones were undefined during server side rendering.
But slugify is not undefined so it was unnecessary.

entry_types/scrolled/package/spec/entryState/structure-spec.js
awaismbajwa: tried both locale en and de but umlauts are not being translated like ä->ae.

awaismbajwa: strict is also not working as expected I think because $ is replaced by dollar. I expected strict to remove it
also <script> translates to lessscriptgreater in strict mode.

entry_types/scrolled/package/src/frontend/Chapter.js
```diff
@@ -5,7 +5,7 @@ import {EventContextDataProvider} from './useEventContextData';
 
 export default function Chapter(props) {
   return (
-    <div id={`chapter-${props.permaId}`}>
+    <div id={props.chapterRef}>
```
tf, line 8: **issue**: Forgot to rename property

entry_types/scrolled/package/src/frontend/Entry.js
```diff
@@ -65,6 +76,7 @@ function renderChapters(entryStructure,
   return entryStructure.map((chapter, index) => {
     return(
       <Chapter key={index}
+               chapterRef={chapter.chapterRef}
```
tf, line 79: **issue**: Forgot to rename property

entry_types/scrolled/package/src/frontend/Entry.js
```diff
@@ -18,13 +18,24 @@ export default withInlineEditingDecorator('EntryDecorator', function Entry(props
   const [scrollTargetSectionIndex, setScrollTargetSectionIndex] = useState(null);
 
   const entryStructure = useEntryStructure();
-
   useSectionChangeEvents(entryStructure, currentSectionIndex);
 
+  let updateChapterSlug = (slug) => {
+    history.replaceState(null, null, '#'+slug);
```
tf, line 24: **issue**: History is still used without feature detection

Also linter still complains that this should be `window.history`.

entry_types/scrolled/package/src/entryState/structure.js
```diff
@@ -105,12 +104,29 @@ export function useSectionContentElements({sectionId}) {
...
   const chapters = useEntryStateCollectionItems('chapters');
-
+  let chapterSlugs = {};
   return chapters.map(chapter => {
+    let chapterSlug = chapter.configuration.title;
+    if (chapterSlug) {
+      chapterSlug = slugify(chapterSlug, {
+        lower: true,
+        locale: I18n.locale,
+        strict: true
```
tf, line 114: Passing `'de'` here makes it replace umlauts for me. `I18n.locale` is not save to use in React components, since it's value depends on the environment (editor, published entry etc.). That's why we have hooks like `useLocale`/`useI18n`. But since Slugify only supports one other locale besides `de` which in turn is not supported by Pageflow, I suggest to hard code `'de'` here. The other replacements (like $ -> dollar) are a bit weird, but inline with what I'd expect after looking at the Slugify source.

tf (changes requested):

entry_types/scrolled/package/spec/entryState/structure-spec.js
tf: see [comment below](https://github.com/codevise/pageflow/pull/1637#discussion_r538460783).

entry_types/scrolled/package/src/entryState/structure.js
awaismbajwa: weird for me it did not work by hard coding 'de' as well.

tf: Check your locally installed slugify package. I had to run `yarn install` to update it since, before your commit, there already was another package depending on an older version which did not yet include de locale.

tf (approved):

PR #1645, opened 2020-12-14
Fix IntersectionObserver issue on Safari

awaismbajwa:
REDMINE-18070

### DIFF
--- a/package/src/frontend/browser/safariPlatform.js
+++ b/package/src/frontend/browser/safariPlatform.js
@@ -1,0 +1,6 @@
+import {browser} from './browser';
+import {agent} from './Agent';
+
+browser.feature('safari platform', function() {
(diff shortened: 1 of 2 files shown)


package/src/frontend/browser/safariPlatform.js
```diff
@@ -0,0 +1,6 @@
+import {browser} from './browser';
+import {agent} from './Agent';
+
+browser.feature('safari platform', function() {
+    return agent.matchesMobileSafari() || agent.matchesSafari();
```
awaismbajwa, line 5: @tf we probably do not need mobile safari ? since I think editor is not supported for mobile version of safari.
and in preview mode it works fine.
But I could not test it on safari

tf:
see #1660

PR #1663, opened 2021-02-24
[WIP] Analytics consent

aviav:
REDMINE-18004


tf:
See #1705, #1710, #1710

PR #1820, opened 2022-05-30
Allow to disable page-level caching

aviav:
Depending on app usage and the caching back-end, it might be faster to
not cache pages and only cache the entries that contain them.

REDMINE-19719

### DIFF
--- a/entry_types/paged/app/views/pageflow_paged/pages/_page.html.erb
+++ b/entry_types/paged/app/views/pageflow_paged/pages/_page.html.erb
@@ -1,4 +1,4 @@
-<%= cache [entry.locale, page] do %>
+<%= decide_caching(page, entry) do %>
   <%= content_tag(:section,
                   id: page.perma_id,
                   class: page_css_class(page),
--- a/app/helpers/pageflow/pages_helper.rb
+++ b/app/helpers/pageflow/pages_helper.rb
@@ -100,5 +100,15 @@ module Pageflow
       ThumbnailFileResolver.new(@entry, page.page_type.thumbnail_candidates, page.configuration)
                            .find_thumbnail
     end
+
+    def decide_caching(page, entry)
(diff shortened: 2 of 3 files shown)


entry_types/paged/app/views/pageflow_paged/pages/_page.html.erb
```diff
@@ -10,3 +10,11 @@
...
   <a class="to_top" href="#content"><%= t('pageflow.public.goto_top') %></a>
 <% end %>
+
+<% if entry.feature_state('page_level_caching') %>
+  <%= cache [entry.locale, page] do %>
+    <%= content_for :page_content %>
+  <% end %>
+<% else %>
+  <%= content_for :page_content %>
+<% end %>
```
tf, line 20: I'd suggest to move this code into a helper method that wraps the `cache` call. This would help avoid the rather unidiomatic usage of `content_for`. Feature still needs translations (i.e. `pageflow.page_level_caching.feature_name`), but otherwise looks fine.

tf: We could also consider deploying the change from a temporary edge branch and only merge it to master once we know whether we actually need it. This would make these concerns less relevant. In this case, I'd still like to check the implementation of `content_for` to make sure it does not impose its own performance costs.

aviav: Thanks for the reminder to use a helper! Yes, I think I'll push the change to an edge branch.

Do we need a spec for this, or is that also in case we actually need it? I tested the change in development, and am waiting for CI jobs to finish before preparing the edge deploy

tf: As long as it's not supposed to be merged, manual testing is fine with me.

PR #1878, opened 2022-10-18
Generate and display feeds for published entries

UsmanMuhammad:
REDMINE-20013

It adds a route, controller and atom builder to display some static text for a published entry


tf:
see #1941

PR #2164, opened 2024-10-21
Ignore console error caused by wysihtml5

tf:
Prevent hint about removal of DOMNodeRemoved event from breaking Capybara specs.


tf:
see #2165

PR #2166, opened 2024-10-21
Introduce helper method for translating public locale names

DanielKuhn:


DanielKuhn:
closed in favor of #2167

